PR #14, opened 2022-04-28
fix: set shebang directly instead of via sh_toolchain

mattem:
Follow up from https://github.com/aspect-build/rules_py/pull/12#discussion_r860151443

### DIFF
--- a/py/private/py_binary.bzl
+++ b/py/private/py_binary.bzl
@@ -7,7 +7,6 @@ load("//py/private:utils.bzl", "PY_TOOLCHAIN", "SH_TOOLCHAIN", "dict_to_exports"
 load("//py/private/venv:venv.bzl", _py_venv = "py_venv_utils")
 
 def _py_binary_rule_imp(ctx):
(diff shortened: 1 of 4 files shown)


py/private/py_binary.bzl
```diff
@@ -7,7 +7,6 @@ load("//py/private:utils.bzl", "PY_TOOLCHAIN", "SH_TOOLCHAIN", "dict_to_exports"
 load("//py/private/venv:venv.bzl", _py_venv = "py_venv_utils")
 
 def _py_binary_rule_imp(ctx):
-    bash_bin = ctx.toolchains[SH_TOOLCHAIN].path
```
alexeagle, line 10: feels like we should leave a comment - I'm not that confident we've got it correct yet. this might just be a workaround for RBE not having the bash toolchain defined.

alexeagle (approved):

mattem:
I'm not convinced this is the right thing, so I'm going to close this. If someone convinces me that this is right fix, rather than registering the right toolchain on RBE, then we can reopen.

PR #28, opened 2022-05-25
chore: tidy and add django example

mattem:
Adds an django example and moves the requirements to the root of the repo


examples/ipython/WORKSPACE
```diff
@@ -0,0 +1,48 @@
+workspace(name = "rules_py_ipython_example")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+local_repository(
```
mattem, line 5: This kinda bugs me - I'd like these examples to be standalone, allowing someone to take the directory as a starting point for a project, but I also don't want to have to maintain the `http_archive` alternative here.

py/tests/external-deps/__main__.py
```diff
@@ -22,7 +22,3 @@
 from lib import greet
 print(f'\nFrom lib with wheel dependency: {greet("Matt")}')
 print(f'lib filepath: {inspect.getsourcefile(greet)}')
-
-from foo import get_branding
```
mattem, line 26: This test is about external deps, will add other test coverage for internal deps.

JesseTatasciore (approved):

alexeagle (approved):

PR #36, opened 2022-06-02
fix: don't add source files to requirements files from transitive native py_library rules

mattem:
Only take the `DefaultInfo` when there is no `PyInfo`. This stops us adding the source files from a rules_python / native py_library to the requirements file when installing wheels.

This is kinda a workaround that the upstream rules_python doesn't have a provider that can give a reference to the wheel, and we must get it from the `DefaultInfo` of a filegroup.

### DIFF
--- a/py/private/py_wheel.bzl
+++ b/py/private/py_wheel.bzl
@@ -24,7 +24,11 @@ def _make_py_wheel_info(ctx, wheel_filegroups):
         if PyWheelInfo in filegroup:
             files_depsets.append(filegroup[PyWheelInfo].files)
             runfiles.append(filegroup[PyWheelInfo].default_runfiles)
-        elif DefaultInfo in filegroup:
+        elif DefaultInfo in filegroup and not PyInfo in filegroup:
+            # This is slightly incorrect, but we don't yet have a better way of knowing if the dependency is a filegroup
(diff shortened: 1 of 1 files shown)


py/private/py_wheel.bzl
```diff
@@ -24,7 +24,11 @@ def _make_py_wheel_info(ctx, wheel_filegroups):
         if PyWheelInfo in filegroup:
             files_depsets.append(filegroup[PyWheelInfo].files)
             runfiles.append(filegroup[PyWheelInfo].default_runfiles)
-        elif DefaultInfo in filegroup:
+        elif DefaultInfo in filegroup and not PyInfo in filegroup:
+            # This is slightly incorrect, but we don't yet have a better way of knowing if the dependency is a filegroup
+            # that we should consume a wheel from.
```
alexeagle, line 29: file an issue on rules_python, if we think the right fix is to put more Info on these targets?

alexeagle (approved):

py/private/py_wheel.bzl
mattem: Yeah ideally the `PyWheelInfo` is removed from rules_py and added to rules_python, and the filegroup in the generated BUILD files for PyPi dependencies become the `py_wheel` rule.

PR #38, opened 2022-06-02
chore: rename BUILD to BUILD.bazel

mattem:
Make the BUILD file naming consistent


alexeagle (approved):
I wonder if we should consider using WORKSPACE.bazel again, now that managed_directories is going away (that was the main bug I know of)

PR #44, opened 2022-06-04
chore: update aspect_bazel_lib to 0.12.1

mattem:
ronseal


alexeagle (approved):
Let's set up renovate everywhere

PR #46, opened 2022-06-07
fix: fail gracefully py_binary or py_test if no main or srcs provided

mattem:
Fail if no `srcs` or `main` is provided to a `py_test` or `py_binary`, this currently fails with an index out of range error.
Also extract all the logic of both these rules into the `py_base` macro.

### DIFF
--- a/py/defs.bzl
+++ b/py/defs.bzl
@@ -28,31 +28,42 @@ def py_binary(name, srcs = [], main = None, imports = ["."], **kwargs):
         name: name of the rule
         srcs: python source files
         main: the entry point. If absent, then the first entry in srcs is used.
+        imports: list of paths that this rule should contribute to the Python path.
         **kwargs: see [py_binary attributes](./py_binary)
     """
-    _py_binary(
+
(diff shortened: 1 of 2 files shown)


py/defs.bzl
```diff
@@ -28,31 +28,42 @@ def py_binary(name, srcs = [], main = None, imports = ["."], **kwargs):
...
     )
 
-    _py_venv(
-        name = "%s.venv" % name,
-        deps = kwargs.pop("deps", []),
-        imports = imports,
+def py_test(name, main = None, srcs = [], imports = ["."], **kwargs):
+    "Identical to py_binary, but produces a target that can be used with `bazel test`."
+
+    _py_base(
```
alexeagle, line 47: nit: this will leak (e.g. with bazel query) so might be worth nitpicking the name a bit.

in rules_js we just named the underlying rule `js_binary` so it matches up for `js_binary` wrapper macro, and `js_test` looks slightly funny

alexeagle (approved):

PR #54, opened 2022-06-10
tests: adds first party dependencies to the container test, add note about usage

mattem:


alexeagle (approved):
nice, is there any related README content for users?

mattem:
Good point. The readme and examples could use some additions in general so will include it then.

PR #60, opened 2022-06-10
refactor: remove use of maybe() helper

alexeagle:
It's no fewer lines than inlining that function, and as a result renovate will understand our dependencies

Alternative to getting support for this pattern in renovate itself

### DIFF
--- a/internal_deps.bzl
+++ b/internal_deps.bzl
@@ -4,13 +4,15 @@ Users should *not* need to install these. If users see a load()
 statement from these, that's a bug in our distribution.
 """
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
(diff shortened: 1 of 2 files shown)


mattem:
This is very repetitive, _maybe_ define `http_archive` at the top of the file with the `maybe` in.

```
load("@bazel_tools//tools/build_defs/repo:http.bzl", _http_archive = "http_archive")
load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")

def http_archive(name, **kwargs):
   maybe(
       _http_archive,
      name = name,
      **kwargs
   )
```

alexeagle:
ah, that's really nice

mattem (approved):
🚀

PR #87, opened 2022-09-22
Demonstrate virtual dependencies

hashbrowncipher:
Virtual dependencies are metadata on a py_library that indicate the need for a given dependency, but do not resolve that dependency to a set of files. The virtual dependency is "resolved" at the terminal rule (e.g. `py_binary`/`py_test`), by use of the `resolutions` parameter.

This way, any number of terminal rules can depend on a single py_library, and each of them can specify its own preferred resolutions. This permits three useful patterns:
 1) monorepos in which many Python projects coexist, but do not want to
    link their dependency versions together
 2) single Python projects that want to test against a matrix of
    different Python interpreters and library versions.
 3) mocking of external Python dependencies with simpler internal
    versions.

In the example given, I added a dependency on `six` in the django-virtual project. If the dependency is removed from the `virtuals` list in `django-virtual/mysite/BUILD.bazel`, the generated virtualenv does not contain the `six` library, and an ImportError is produced.

This implementation could potentially be extended to include Python requirements specifiers [1] for each virtual dependency (e.g. "six>=1.10"), to allow another tool (e.g. pip-compile) to gather all of the dependencies and resolve them into a requirements.txt file.

[1] https://pip.pypa.io/en/stable/reference/requirement-specifiers/


alexeagle:
Hi, I'm not sure if @mattem has reached out to you already? Thanks for the PR, and I home we can incorporate your work into #201

hashbrowncipher:
That's great news! Should we close this one then?

mattem:
Done in #201

PR #112, opened 2022-12-12
Clean up all symlinks in venv

mvukov:
Cleans up all symlinks in generated venv. This needs to happen such that remote caching works properly when `--remote_download_minimal` or `--remote_download_toplevel` are used.


mvukov:
@mattem PTAL.

mattem (approved):

PR #129, opened 2023-01-19
fix black and flake8 errors in generated pytest runner.

tellett:
Fix flake8 and black formatting issues in generated pytest driver by adding a no-op statement instead of an empty string. The previous version resulted in a file with 2 blank lines if chdir was undefined, resulting in the following issues being raised:

```
bazel-out/k8-fastbuild/bin/REDACTED/__test__.py:21:1: BLK100 Black would make changes.
bazel-out/k8-fastbuild/bin/REDACTED/__test__.py:21:1: W293 blank line contains whitespace
bazel-out/k8-fastbuild/bin/REDACTED/__test__.py:23:5: E303 too many blank lines (2)
```

### DIFF
--- a/py/private/py_pytest_main.bzl
+++ b/py/private/py_pytest_main.bzl
@@ -20,7 +20,9 @@ load("@rules_python//python:defs.bzl", default_py_library = "py_library")
 def _py_pytest_main_impl(ctx):
     substitutions = {
         "$$FLAGS$$": ", ".join(['"{}"'.format(f) for f in ctx.attr.args]).strip(),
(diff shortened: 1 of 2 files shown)


mattem:
Thanks for the PR!

Can we add a comment around the line in the template that gets replaced? Even if it points to the comment here, it makes the noop assignment a little less mystical to someone looking at the generated code.

tellett:
You're welcome, thanks for the library :) .

I added a comment in the generated file, and verified that it works as intended in my local branch and in my workspace.

mattem (approved):

PR #130, opened 2023-01-20
s/external/../ if the python interpreter cannot be found in the resolved venv creation script.

tellett:
s/external/../ if the python interpreter cannot be  found in the resolved venv creation script.

In my environment the python interpreter is up a level and doesn't have 'external/' as a prefix. Not really sure why. This works for my environment, but not certain that it's the correct change to make. Let me know what you think.


mattem:
Are you able to provide a minimal repro of the issue? Make sure to include how the interpreter is fetched. I think we should understand the problem before merging a fix.

alexeagle:
This has changed a fair amount now that we stopped using the virtualenv python package to lay this out. I think it's likely not reproducible anymore, and pretty stale.

PR #142, opened 2023-02-09
release: publish release artifact

alexeagle:
GitHub's stability guarantee for the archive is iffy, and we want metrics on downloads. See https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1411222578


gregmagolan (approved):
🧄

Issue #31, opened 2022-05-27
non-wheel external dependencies don't make it into the IDE venv


mattem:
If there is a dependency on an external `py_library` that isn't consumed via the `py_wheel` rule, eg it's a `py_library` generated by rules_python for a pip install, and referenced via the `:pkg` label, then when a venv is created for the test or binary target for IDE consumption, the dependency is missing from the venv, as the `.pth` file points to the path in the runfiles tree.

psalvaggio-dl:
I have been able to verify this bug. For my use case, this is important because we rely on virtual environments for people's development environments and running some command line tools that have proven difficult to integrate into Bazel due to other bugs or missing features. In order to migrate a monorepo onto these rules, I would need to ensure that all of my dependencies were wheel dependencies, which requires explicitly annotating them in the `pip_parse` call. Since Starlark can't read files and you can't subclass `dict()`, there's not a clean way to do this with a traditional `requirements.txt` file.

alexeagle:
Maybe already fixed...

mattem:
Fixed by https://github.com/aspect-build/rules_py/pull/233

Issue #2, opened 2022-04-14
Investigate PyCharm issue with .pth files


alexeagle:

mattem:
Opened upstream issue https://youtrack.jetbrains.com/issue/IDEA-294464/pth-files-inside-a-venv-site-packages-is-ignored with minimal non-bazel repro https://github.com/mattem/ij-venv-pth

alexeagle:
Fixed upstream, users need a new enough PyCharm.

PR #153, opened 2023-03-22
Fix PTH generation for standalone virtualenvs

pbecotte:
Virtualenvs created in the runfiles tree can safely assume where they live- as such, they can happily use relative paths to find the correct location of whatever py_library targets they need to link in. However, using 'bazel run :sometarget.venv' is expected to create a virtualenv in the root source directory, presumably for user IDE integration. The existing code creates a broken one however- basically every path is changed to ../../../.. . Since this virtualenv doesn't live in a runfiles tree, it cannot find packages that way- but since it is running in 'bazel run' instead of during execution, we can use runfiles to find the correct absolute paths. I could not figure out a use case for strip_pth_workspace_root, or any uses of it that didn't result in a broken virtualenv, so I removed it.

### DIFF
--- a/py/private/BUILD.bazel
+++ b/py/private/BUILD.bazel
@@ -51,8 +51,8 @@ bzl_library(
 bzl_library(
     name = "py_pytest_main",
     srcs = ["py_pytest_main.bzl"],
-    deps = ["@rules_python//python:defs"],
     visibility = ["//py:__subpackages__"],
+    deps = ["@rules_python//python:defs"],
 )
 
 bzl_library(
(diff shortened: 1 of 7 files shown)


pbecotte:
Should address https://github.com/aspect-build/rules_py/issues/31 and probably https://github.com/aspect-build/rules_py/issues/2

alexeagle:
Thanks @pbecotte  - FYI @mattem saw this and plans to get to it.

mattem:
Took a quick look over this. I don't quite understand the change. When running the venv creation targets now (for example, ` //py/tests/internal-deps:main.venv`) the resulting venv can't resolve first party imports, however the version at HEAD can.

Have I missed something here? The manual testing from #181 seemed to result in the venv pathing as expected.

mattem:
Resolved as part of #233

PR #163, opened 2023-06-22
Allow passing deps in py_pytest_main

linzhp:
### Type of change

`py_pytest_main` generates a `py_library` that depends on pytest without declaring it as the dependency. This PR adds a `deps` attribute to `py_pytest_main`, and pass it down to the `py_library`, because:

1. it's more "correct" to have a `py_library` declaring its own dependency instead of having its consumer to do that job
2. Gazelle would remove `@pypi_pytest//:pkg` from the `py_test` target that depends on `py_pytest_main`, because the `py_test` target doesn't import pytest.

### Test plan


mattem (approved):
Thanks!

PR #181, opened 2023-09-08
feat: expose py_venv rule, allowing creation of virtual envs from multiple targets

mattem:
Exposes the `py_venv` rule, which can create a Python virtual environment from a set of targets.
This allows for easier creation of environments for IDE use, as they can contain dependencies from multiple targets, such as tests.

The virtual environment that is created uses paths for first party libraries that resolve into the source tree. While this allows for a nicer user experience within the IDE, such as editing files being "live", it is less "correct" as the Bazel target may reflect a different state.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -2,6 +2,29 @@
 
 Public API re-exports
 
+<a id="py_venv"></a>
+
+## py_venv
+
+<pre>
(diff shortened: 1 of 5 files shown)


docs/rules.md
```diff
@@ -2,6 +2,29 @@
 
 Public API re-exports
 
+<a id="py_venv"></a>
+
+## py_venv
```
alexeagle, line 7: in a later PR I'll fix all the missing docs here

alexeagle (approved):

Issue #189, opened 2023-09-18
[FR]: Ability to pass-in `testonly` to `py_pytest_main`


kpark-hrp:
### What is the current behavior?

`kwargs` in `py_pytest_main` macro is only applied to the underlying `_py_pytest_main` rule. Thus, `testonly` does not get applied to the `py_library` rule which makes up the pytest entry point.

### Describe the feature

It would be useful to be able to pass in `testonly` flag to the underlying `py_library` rule defined by the `py_pytest_main` macro.

Specifically [here](https://github.com/aspect-build/rules_py/blob/901405bf353084568b602c80b5c5995b6109ae9b/py/private/py_pytest_main.bzl#L78C1-L84C1)

kpark-hrp:
Also, `data` parameter would also be nice, so I can pass in like `pytest.ini`

PR #190, opened 2023-09-18
Add `testonly` and `data` parameters to `py_pytest_main` rule

kpark-hrp:
### Type of change

Adds 2 new **optional** parameters to `py_pytest_main(...)` rule: `data` and `testonly`. These are standard arguments to the `py_library`. This change simply exposes the arguments to the `py_pytest_main` rule

closes: #189 

### Test plan

### DIFF
--- a/py/private/py_pytest_main.bzl
+++ b/py/private/py_pytest_main.bzl
@@ -53,13 +53,15 @@ _py_pytest_main = rule(
     },
 )
 
-def py_pytest_main(name, py_library = default_py_library, deps = [], **kwargs):
+def py_pytest_main(name, py_library = default_py_library, deps = [], data = [], testonly = False, **kwargs):
     """py_pytest_main wraps the template rendering target and the final py_library.
 
     Args:
         name: The name of the runable target that updates the test entry file.
         py_library: Use this attribute to override the default py_library rule.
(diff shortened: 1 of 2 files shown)


py/private/py_pytest_main.bzl
```diff
@@ -53,13 +53,15 @@ _py_pytest_main = rule(
     },
 )
 
-def py_pytest_main(name, py_library = default_py_library, deps = [], **kwargs):
+def py_pytest_main(name, py_library = default_py_library, deps = [], data = [], testonly = False, **kwargs):
```
alexeagle, line 56: Wouldn't `testonly=True` be a better default? Is there a use case for using the result from production code?

dzbarsky, line 56: this is a test rule, when would you use testonly = False? i.e. should the default be True, or even force to True?

kpark-hrp: I was debating about that. My rationale was that `testonly = False` is the `py_library` default, but didn't have strong reason.

kpark-hrp: Same reasoning as above. Changed to `True` 😄

alexeagle (approved):

alexeagle:
Thanks!

PR #204, opened 2023-10-03
chore: dogfood our rules rather than rules_python

alexeagle:
### Type of change

### Test plan


thesayyn (approved):

alexeagle:
@mattem maybe it's like this on purpose for py_pytest_main to demonstrate that it works with rules_python#py_test ?

mattem (approved):

mattem:
> @mattem maybe it's like this on purpose for py_pytest_main to demonstrate that it works with rules_python#py_test ?

I don't think so, probably just not placing rules_py in the critical path.

CI failure here looks legit.

PR #206, opened 2023-10-05
Use coreutils toolchain for venv creation

dzbarsky:
Doesn't work yet, `grep` is still not found, `find` fails and then we abort. Chmod probably needs similar treatment

### Type of change

**- Bug fix (change which fixes an issue)**
- Style (white-space, formatting, etc...)
- Performance (a code change that improves performance)
- Documentation (updates to documentation or READMEs)

### Test plan

- New test cases added


mattem:
Will close in favour of #233

PR #226, opened 2023-11-20
fix: use original interpreter path as 'home' value in pyvenv.cfg

mattem:
Use the original (source) interpreters bin path as the 'home' value in the generated `pyvenv.cfg` file.

Closes #159 

---

### Type of change

- Bug fix (change which fixes an issue)

**For changes visible to end-users**

- Suggested release notes are provided below:

Ensure correct `home` pathing is used when generating the venv.

### Test plan

- Covered by existing test cases
- Manual testing; please provide instructions so we can reproduce:

Updated e2e smoke to use py3.11

PR #232, opened 2023-12-18
fix: use original interpreter path as 'home' value in pyvenv.cfg

mattem:
This is a follow-up of #226 to fix another place.


alexeagle (approved):

liningpan:
Hey guys! Was this fix ever released? I suspect this might be why the virtual environment I tried to created with `py_venv` is broken (for IDE).

alexeagle:
I'm working on a release right now but it's a long process because we introduced Rust for writing tools and don't want to expose that dependency to users.
https://github.com/aspect-build/rules_py/releases - last release was november.

That said, I think we should just cut a release from before the rust toolchain introduction.

alexeagle:
published https://github.com/aspect-build/rules_py/releases/tag/v0.6.0

PR #241, opened 2024-01-11
fix: handle py_binary#main like rules_python does

alexeagle:
I don't love the semantics but it's better for us to follow our 'drop-in replacement' philosophy than explain to users how they're meant to change their code.

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/py/defs.bzl
+++ b/py/defs.bzl
@@ -18,7 +18,7 @@ _a_struct_type = type(struct())
 _a_string_type = type("")
 
 def _py_binary_or_test(name, rule, srcs, main, imports, deps = [], resolutions = {}, **kwargs):
-    # Compatibility with rules_python, see docs on find_main
+    # Compatibility with rules_python, see docs in py_executable.bzl
     main_target = "_{}.find_main".format(name)
     determine_main(
         name = main_target,
@@ -56,8 +56,10 @@ def py_binary(name, srcs = [], main = None, imports = ["."], resolutions = {}, *
     Args:
         name: name of the rule
         srcs: python source files
-        main: the entry point. If absent, then the first entry in srcs is used. If srcs is non-empty,
-            then this is treated as a suffix of a file that should appear among the srcs.
+        main: the entry point.
+            Like rules_python, this is treated as a suffix of a file that should appear among the srcs.
+            If absent, then "[name].py" is tried. As a final fallback, if the srcs has a single file,
+            that is used as the main.
         imports: List of import paths to add for this binary.
         resolutions: FIXME
         **kwargs: additional named parameters to the py_binary_rule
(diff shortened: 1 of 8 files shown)


py/defs.bzl
```diff
@@ -37,6 +38,32 @@ def _py_binary_or_test(name, rule, srcs, main, imports, deps = [], resolutions =
         tags = ["manual"],
     )
 
+def _find_main_impl(ctx):
```
mattem, line 41: Would prefer if this went in its own bzl file to keep this simply wrappers and re-exports.

py/defs.bzl
```diff
@@ -37,6 +38,32 @@ def _py_binary_or_test(name, rule, srcs, main, imports, deps = [], resolutions =
...
+    found = None
+    for s in ctx.files.srcs:
+        if s.path.endswith(ctx.attr.main):
```
mattem, line 44: This will keep searching over the sources after it has been found, either wastfully or potentially finding another match. What is the behaviour of upstream, does it match on the last or first? Perhaps we need a `break` in here.

mattem (changes requested):
Please add tests too.

py/defs.bzl
alexeagle: done - used the same filename as the rules_python code to make it a bit easier to reason about

py/defs.bzl
alexeagle: I went to locate the upstream behavior, it's already in starlark so it was easy to just vendor their `determine_main` logic.
(to answer your question, they also loop over all of them, then error if more than one match found)

py/tests/external-deps/custom-macro/macro.bzl
```diff
@@ -1,17 +1,12 @@
...
 def click_cli_binary(name, deps = [], **kwargs):
-    py_library(
-        name = name + "_lib",
-        srcs = ["//py/tests/external-deps/custom-macro:__main__.py"],
-    )
-
     py_binary(
         name = name,
         main = "//py/tests/external-deps/custom-macro:__main__.py",
+        srcs = ["//py/tests/external-deps/custom-macro:__main__.py"],
```
alexeagle, line 8: note: this is required because this file was previously relying on locating a main among deps rather than srcs. This would not have worked with rules_python, so there shouldn't be user code like this.

py/private/py_executable.bzl
```diff
@@ -0,0 +1,93 @@
...
+        ctx: The rule ctx.
+
+    Returns:
+        Artifact; the main file. If one can't be found, an error is raised.
+    """
+    if ctx.attr.main:
+        # Deviation from rules_python: allow a leading colon, e.g. `main = ":my_target"`
+        proposed_main = ctx.attr.main.removeprefix(":")
+        if not proposed_main.endswith(".py"):
+            fail("main must end in '.py'")
```
mattem, line 31: nit: it would be nice if the target was included in these error messages, makes it much easier to find in a large repo.

mattem (approved):

honnix:
If I understand it correctly, the following code would stop working:
```starlark
py_pytest_main(
    name = "__test__",
)

py_test(
    name = "test_all",
    srcs = glob(["**/*.py"]),
    main = ":__test__.py",
    deps = [":__test__", ...],
)
```
and it should now be:
```starlark
py_pytest_main(
    name = "__test__",
)

py_test(
    name = "test_all",
    srcs = glob(["**/*.py"]) + [":__test__"],
    main = ":__test__.py",
    deps = [...],
)
```

Is that correct?

PR #234, opened 2023-12-21
chore: minimal changes required to upgrade to Bazel 7 and bzlmod

alexeagle:
In a follow-up change I'll try to empty out the WORKSPACE file.
Also we need another change to add CI testing for other Bazel versions, assuming we decide to continue supporting Bazel 6 and have tests for it.

### Type of change

- Chore (any other change that doesn't affect source or test files, such as configuration)

**For changes visible to end-users**

This shouldn't be user-visible, it's only applicable when building within rules_py.

### Test plan

- Covered by existing test cases

PR #246, opened 2024-01-20
chore: enable bzlmod

alexeagle:
Next step from #234 

After this lands, the final step(s) are to clear out the WORKSPACE file.

### Type of change

This should not affect users, only our local dev/testing.

### Test plan


mattem:
I think this is blocked by https://github.com/bazelbuild/rules_rust/issues/2420

PR #249, opened 2024-02-04
chore: upgrade to Bazel 7.0.2

gregmagolan:


py/dependencies.bzl
```diff
@@ -0,0 +1,35 @@
+"""Declare runtime dependencies
```
alexeagle, line 1: bad merge maybe?

gregmagolan: whoops. some other wip snuck in.

jbedard (approved):

PR #258, opened 2024-02-15
feat: detect prerelease, use source bins

alexeagle:
Also switch to rusttls to avoid a dependency on compiling libssl


tools/version.bzl
```diff
@@ -0,0 +1,15 @@
+"version information. replaced with stamped info with each release"
+
+# This is automagically replace by git during git archive using `git export-subst`
+# See https://git-scm.com/docs/git-archive#Documentation/git-archive.txt-export-subst
+_VERSION_PRIVATE = "$Format:%(describe:tags=true)$"
+
+VERSION = "0.0.0" if _VERSION_PRIVATE.startswith("$Format") else _VERSION_PRIVATE.replace("v", "", 1)
+
+# Whether bazel-lib is a pre-release, and therefore has no release artifacts to download.
```
mattem, line 9:
```suggestion
# Whether rules_py is a pre-release, and therefore has no release artifacts to download.
```

mattem (approved):

PR #264, opened 2024-02-20
Update ci.yaml

alexeagle:
Update to latest from the "upstream" rules-template.

This adds the missing test execution of the e2e/ folders.
Expecting it to be red to start with, as we have a rules_rust dangling load.

---

### Type of change


mattem:
Merged as part of #265

PR #266, opened 2024-02-20
feat: support Bazel 6

alexeagle:
### Type of change

### Test plan

- New test cases added


mattem:
The `py_unpacked_wheel` rule also needs this info, so seems like it needs to live in a more common place.

mattem (approved):

PR #233, opened 2023-12-20
feat: switch to creation of venv without Python or pip

mattem:
Switches creating the venv to a dedicated tool, removing the requirement for pip.

Venv is now created just before the binary or test runs, rather than as an action output. The overhead is minimal, venv creation and activation takes at most 7ms as observed during testing.

Closes #233
Closes #238
Closes #176
Closes #218
Closes #248
Closes #152
Closes #110

Issue #218, opened 2023-11-07
[Bug]: py_binary breaks cross_platform build when building image with rules_oci


lyao-77:
### What happened?

When we use `py_binary` to build docker images with rules_oci, we couldn't build the tarball for ubuntu images. The error log:

> ERROR: /Users/lyao/bazel/bazel-examples/oci_python_image/hello_world/BUILD.bazel:18:10: Creating virtual environment for //hello_world:hello_world failed: (Exit 126): sandbox-exec failed: error executing command 
>   (cd /private/var/tmp/_bazel_lyao/1e813cdbb954646ce252ec3692bb82d4/sandbox/darwin-sandbox/10/execroot/_main && \
>   exec env - \
>     TMPDIR=/var/folders/gm/r7n8598x7zq6rwkd1_r75_3h0000gp/T/ \
>   /usr/bin/sandbox-exec -f /private/var/tmp/_bazel_lyao/1e813cdbb954646ce252ec3692bb82d4/sandbox/darwin-sandbox/10/sandbox.sb /var/tmp/_bazel_lyao/install/7d4a2659eff664c24e2f7d717eef5139/process-wrapper '--timeout=0' '--kill_delay=15' '--stats=/private/var/tmp/_bazel_lyao/1e813cdbb954646ce252ec3692bb82d4/sandbox/darwin-sandbox/10/stats.out' /bin/bash -c bazel-out/darwin_arm64-fastbuild-ST-baeacbe55d95/bin/hello_world/hello_world.venv_venv.sh)
> bazel-out/darwin_arm64-fastbuild-ST-baeacbe55d95/bin/hello_world/hello_world.venv_venv.sh: line 38: external/rules_python~0.21.0~python~python3_9_aarch64-unknown-linux-gnu/bin/python3: cannot execute binary file
> INFO: Elapsed time: 0.763s, Critical Path: 0.37s
> INFO: 3 processes: 3 internal.
> FAILED: Build did NOT complete successfully

I'm using https://github.com/aspect-build/bazel-examples/tree/main/oci_python_image to reproduce the error.

The issue seems relate to the wrong python interpreter path and is not platform agnostic when calling venv.

I tried the following command with Mac as a workaround:
`❯ bazel build //... --sandbox_debug --platform=//hello_world:x86_64_linux`
it ran into the same issue (different error) and seems the current implementation of py_binary breaks the cross-platform build.

### Version

Development (host) and target OS/architectures:
host: Mac with M1
[...]

PR #269, opened 2024-02-21
Remove openssl

btobolaski:
### Type of change

### Test plan

#233 was mentioned as a fix for #218 but #233 includes a dependency on openssl. This dependency on openssl makes building containers on MacOS difficult as we then need to cross build openssl. I see that there was an attempt to remove the openssl dependency `rattler_installs_packages` however, `native-tls` is in the default features and is the only default feature. Disabling `default-features` removes the dependency on openssl.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -12,6 +12,7 @@ bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_python", version = "0.29.0")
 bazel_dep(name = "platforms", version = "0.0.7")
+bazel_dep(name = "rules_rust", version = "0.35.0")
 
 archive_override(
     module_name = "rules_python",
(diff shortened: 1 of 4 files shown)


mattem:
Thanks! There was a recent PR that should remove the requirement to build the binaries from source, and not push the rust dependencies onto the consumer too. We should still land this as this is the right shape.

Are you able to sign the CLA?

btobolaski:
I needed to run it through my Company but, it is now signed.

MODULE.bazel
```diff
@@ -12,6 +12,7 @@ bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_python", version = "0.29.0")
 bazel_dep(name = "platforms", version = "0.0.7")
+bazel_dep(name = "rules_rust", version = "0.35.0")
```
alexeagle, line 15: We shouldn't introduce a new dependency for users in this change. I'll push a commit to restore this file for now.
(We'll fix this up soon...)

PR #275, opened 2024-02-27
site-packages symlink forest in venv

### LINKED ISSUE #274: [Bug]: Some packages rely on site-packages layout
### What happened?

rules_py doesn't really create a standard `pip install`-style `site-packages` folder currently, rather it creates a `.pth` file and expects packages to "do the right thing".

This plays badly with loading shared objects in particular, such as 

```
Error: /main.runfiles/pip_deps_nvidia_cudnn_cu11/site-packages/nvidia/cublas/lib/libcublas.so.11: cannot open shared object file: No such file or directory
# But that file libcublass.so.11 exists in a different external package:
./main.runfiles/pip_deps_nvidia_cublas_cu11/site-packages/nvidia/cublas/lib/libcublas.so.11
```

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

siddharthab:
This emulates the behavior where a single site-packages directory
contains all/most PyPi packages. Packages distributed by NVIDIA
currently assume this through the use of rpath as `$ORIGIN/../../`
to reach the `nvidia` package location. Downstream libraries like torch
and jax do not set up the dynamic library search path based on sys.path
either.

This is a casual attempt, but can be refined by others.

Fixes #274.


alexeagle:
Note, at HEAD we no longer have checked-in binaries so you can remove that when you rebase. I'm very +1 to having compatibility with ~all existing packages and satisfying the assumptions they make. I can't read Rust yet so handing to Matt.

mattem:
Thanks for this! Will try to take a look soon.

py/tools/py/src/pth.rs
```diff
@@ -48,13 +32,90 @@ impl PthFile {
...
+    ))?;
+    for entry in read_dir {
+        let entry = entry.into_diagnostic().wrap_err(format!(
+            "unable to read directory entry {}",
+            dir.to_string_lossy()
+        ))?;
+        let path = entry.path();
+        if path.is_dir() {
+            create_symlinks(&path, root_dir, dst_dir)?;
+        } else if dir != root_dir || entry.file_name() != "__init__.py" {
```
mattem, line 79: following this through, what's the case for special casing `__init__.py` here and below?

py/tools/py/src/pth.rs
```diff
@@ -48,13 +32,90 @@ impl PthFile {
...
+            create_symlink(&entry, root_dir, dst_dir)?;
+        }
+    }
+    Ok(())
+}
+
+fn create_symlink(e: &DirEntry, root_dir: &Path, dst_dir: &Path) -> miette::Result<()> {
+    let tgt = e.path();
+    let link = dst_dir.join(tgt.strip_prefix(root_dir).unwrap());
+    if link.exists() && link.file_name().is_some_and(|x| x == "__init__.py") && is_same_file(link.as_path(), tgt.as_path())? {
```
mattem, line 89: Can we add some comments here around what the casing is? ie, why check that the `__init__.py` files are the same? I'm not fully sure I follow the logic here.

py/tools/py/src/pth.rs
```diff
@@ -48,13 +32,90 @@ impl PthFile {
...
+            if self.prefix.is_some() {
+                entry = Path::new(self.prefix.as_deref().unwrap()).join(Path::new(line.trim()));
+            } else {
+                entry = PathBuf::from(line.trim());
+            }
             line.clear();
-            writeln!(writer, "{}", entry.to_string_lossy())
-                .into_diagnostic()
-                .wrap_err("Unable to write new .pth file entry with prefix")?;
+            if entry.file_name().is_some_and(|x| x == "site-packages") {
```
mattem, line 42: Perhaps I shouldn't push more on https://github.com/bazelbuild/rules_python/pull/1766 as that would remove this as a path suffix.

py/tools/py/src/pth.rs
siddharthab: Now explained more in comments.

py/tools/py/src/pth.rs
mattem: Ah, I see. I think most folks set `--incompatible_default_to_explicit_init_py` and these files don't get created. Good to be defensive here though imo.

mattem (approved):

PR #280, opened 2024-02-29
chore: add release rules for binaries

mattem:
Adds release infra for the rust binaries.

---

### Type of change

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,10 +1,14 @@
 # Declare the local Bazel workspace.
 workspace(name = "aspect_rules_py")
 
+load("//tools/release:fetch.bzl", _release_deps = "fetch_deps")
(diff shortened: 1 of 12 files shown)


WORKSPACE
```diff
@@ -30,13 +34,39 @@ load("@aspect_bazel_lib//lib:repositories.bzl", "register_coreutils_toolchains")
 register_coreutils_toolchains()
 
 ############################################
-## CC toolchain using zig
-load("@hermetic_cc_toolchain//toolchain:defs.bzl", zig_toolchains = "toolchains")
```
alexeagle, line 34: nit: I think we could remove a fetch for this? will do in a follow-up, anyhow we're likely to change the cc toolchain again for musl

alexeagle (approved):

PR #286, opened 2024-02-29
feat: remove checked-in Rust binaries

alexeagle:
All users must currently build the tools from sources with rules_rust.

As a separate PR I will introduce a new toolchain that uses our typical repository rule constructs from bazel-lib to fetch the pre-built release binaries.

---

### Type of change

Users must now register a rust toolchain, or else wait for the next release of rules_py.

### Test plan

e2e/smoke covers this, though it sadly needs a bunch of additions.


internal_deps.bzl
```diff
@@ -10,6 +10,14 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 def http_archive(name, **kwargs):
     maybe(_http_archive, name = name, **kwargs)
 
+def rules_py_rust_deps():
```
mattem, line 13: nit: perhaps this should be `rules_py_tool_deps`? The implementation language is just that, an implementation detail that doesn't need exposing here, and if we add others, this doesn't need to change.

.bcr/patches/rust_dev_dep.patch
```diff
@@ -0,0 +1,36 @@
+diff --git a/MODULE.bazel b/MODULE.bazel
```
mattem, line 1: Does this have to live here? Or can it live in the name spaced third_party folder?

internal_deps.bzl
alexeagle: sure - I wish rules_rust would change so we could tuck the dependency in here altogether, right now the rust is quite leaky since it's a three-time load/init dance in WORKSPACE

alexeagle: looking at this again, I think it's not even worth having this in its own function. WORKSPACE users on a prerelease will still have to load/call all of these:

```
rules_rust_dependencies()
rust_register_toolchains()
crate_universe_dependencies()
crates_repository()
```

may as well force them to have the trivial `http_archive` for rules_rust too

.bcr/patches/rust_dev_dep.patch
alexeagle: it has to be here for the publish-to-bcr app to locate it and apply the patch to the MODULE file that gets published to the registry.

mattem (approved):

PR #287, opened 2024-03-01
feat: don't require users to build tools from source

alexeagle:
Wire up the toolchains along two different codepaths, based on whether our version was stamped as a release, or otherwise a pre-release.

When a release:
- fetch tools from GitHub releases using `rctx.download`
- each fetched tool goes into a separate external repo with its own `toolchain` declaration, so Bazel fetches only the needed ones
- the resulting repos are returned to repositories.bzl which registers them

When a pre-release:
- continue to build the tools from source, using existing toolchains

---

### Type of change

### Test plan

Use my personal fork to test against a release I performed there.


.github/workflows/ci.yaml
```diff
@@ -29,3 +29,9 @@ jobs:
           {"folder": ".", "bazelversion": "6.4.0"},
           {"folder": ".", "bzlmodEnabled": true}
         ]
+
+  verify-bcr-patches:
```
alexeagle, line 33: oops, this should have been in the prior PR where I added a patch file - we want to check when it needs to be recreated.

mattem (approved):
🚀

alexeagle:
Follow-ups that we ought to do before publishing a release, maybe in another PR:
- the e2e/smoke, bzlmod=True case is still using from-source binaries because it doesn't use this new conditional codepath
- we should test that we only fetch tools for the required platform. Currently the external/ folder ends up with all the tools.

PR #292, opened 2024-03-02
Publish binaries on releases

### LINKED ISSUE #289: [Bug]: downloads too many pre-built binaries from release
### What happened?

Follow-up to #287 
I believe our toolchain setup still downloads the pre-built rust binaries for unneeded platforms.

Should add some test (maybe manual, outside-bazel) to confirm that problem, and use it to prevent regressions.

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

### LINKED ISSUE #290: [Bug]: Under bzlmod, always builds rust binaries from source
### What happened?

Followup to #287 which added conditional logic for WORKSPACE users to download pre-built binaries instead, and register them in the toolchain.

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
Fixes #290 
Fixes #289 

---

### Type of change

### Test plan


MODULE.bazel
```diff
@@ -13,14 +13,15 @@ bazel_dep(name = "bazel_skylib", version = "1.4.2")
...
-    urls = "https://github.com/bazelbuild/rules_python/archive/52381415be9d3618130f02a821aef50de1e3af09.tar.gz",
-    integrity = "sha256-pYfEFNWqygSEElDYgJsuIeDYn9oll/rZB0GcR+6rirA=",
-    strip_prefix = "rules_python-52381415be9d3618130f02a821aef50de1e3af09",
-)
+tools = use_extension("//py:extensions.bzl", "py_tools")
+tools.rules_py_tools()
+use_repo(tools, "rules_py_tools")
 
 register_toolchains(
+    "@rules_py_tools//:all",
```
thesayyn, line 21: are concerned with the order of toolchains being registereted?

py/private/toolchain/repo.bzl
```diff
@@ -0,0 +1,77 @@
...
+        for [platform, meta] in TOOLCHAIN_PLATFORMS.items():
+            build_content += """
+# Declare a toolchain Bazel will select for running {tool} on the {cfg} platform.
+toolchain(
+    name = "{tool}_{platform}_toolchain",
+    {cfg}_compatible_with = {compatible_with},
+    # Bazel does not follow this attribute during analysis, so the referenced repo
+    # will only be fetched if this toolchain is selected.
+    toolchain = "@{user_repository_name}.{platform}//:{tool}_toolchain",
+    toolchain_type = "@aspect_rules_py//py/private/toolchain:{tool}_toolchain_type",
```
thesayyn, line 40: Do we need a new toolchain type for exec and target platforms? if any of these tools are invoked at runtime and compile time?

thesayyn (approved):

MODULE.bazel
alexeagle: The ones in this package are guaranteed to be mutually exclusive with each other, so the order shouldn't matter here.

py/private/toolchain/repo.bzl
alexeagle: Nope, the reason we have two tools is that venv is only used at runtime in a binary, and unpack is only used in an action. Commented that over here https://github.com/aspect-build/rules_py/pull/292/files#diff-d06e56acb93ae623783dbfa6032fe99d955510112ef0fdaf59286b3d3e564ae4R8

Issue #276, opened 2024-02-28
[Bug]: py_library instances from generated repositories are not found by py_test (bzlmod)


wade-arista:
### What happened?

I'm using bzlmod in bazel 7.0.2.

I created a basic `repository_rule` that creates a `rules_py` `py_library`. I then created a simple test that specifies the external repository in its deps and tries to import this `py_library` (`foo`). It doesn't find the python package. If I access other `rules_py` `py_library` instances from my `_main` repo, instead, those are found as expected.

```mermaid
graph TD;
test(:test) --> baz(direct:baz)
test --> foo("@myrepo//:foo")
test --> bar("@myrepo//bar")
    subgraph "@myrepo"
    foo
    bar
    end
```

### Version

Development (host) and target OS/architectures: linux x86_64 "Ubuntu 22.04.3 LTS"

Output of `bazel --version`:
```
Bazelisk version: v1.19.0
Build label: 7.0.2
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Thu Jan 25 16:13:35 2024 (1706199215)
Build timestamp: 1706199215
Build timestamp as int: 1706199215
```

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

```
bazel_dep(name = "rules_python", version = "0.31.0")
bazel_dep(name = "aspect_rules_py", version = "0.5.0")
```

Language(s) and/or frameworks involved:
Python3

### How to reproduce

I have a minimal repro here:

https://github.com/wade-arista/bazel-demo/tree/63fc4391a7c30be0587a642649b0c16514fab302/rules_py_reporule

test:

```python
import baz
import foo
import bar
```

BUILD.bazel
```starlark
load("@aspect_rules_py//py:defs.bzl", "py_test")

py_test(
    name = "test",
    srcs = ["test.py"],
    deps = [
        "//rules_py_reporule/direct/baz",
        "@myrepo//:foo",
        "@myrepo//bar",
    ],
)

py_test(
    name = "all_direct",
    srcs = ["test.py"],
    main = "test.py",
    deps = [
        "//rules_py_reporule/direct/baz",
        "//rules_py_reporule/imported:foo",
        "//rules_py_reporule/imported/bar",
    ],
)
```

for `@myrepo`:
```starlark
importer = use_extension("//rules_py_reporule/rules:import.bzl", "importer")
use_repo(importer, "myrepo")
```

### Any other information?

[...]

PR #299, opened 2024-03-05
fix: use correct runfiles path for venv import paths

wade-arista:
In `_make_import_path`, `ctx.workspace_name` is always `_main` when evaluating the deps of targets in the main repo. This includes deps that are in external repos, e.g.

    importer = use_extension("//:import.bzl", "importer")
    use_repo(importer, "myrepo")

with

    py_test(..., deps = ["@myrepo//foo"])

In this case, we actually want to use the label's `workspace_name` attr which will be the same as the runfiles path that we need for the venv import paths.

Also added a new e2e test that creates this external repo and verifies the fix. Without the fix in place, the test does fail as described in the e2e/repository-rule-deps/README.md "Repro" section.

---

### Test plan
- New test case added in `e2e/repository-rule-deps`.
- Manually tested against bazel 6.5 and 7.0.2.

Fixes: #276


wade-arista:
I would like to see that the new e2e/repository-rule-deps is properly executed via the new "repository-rule-deps" check before this merges as well. I don't expect that I have permissions to run this, though.

alexeagle (approved):
Wow, nicely documented!

.github/workflows/ci.yaml
```diff
@@ -30,6 +30,13 @@ jobs:
           {"folder": ".", "bzlmodEnabled": true}
         ]
 
+  repository-rule-deps:
```
alexeagle, line 33: this doesn't need a new job, you can just add to the `folders` array of the `test` job above

scasagrande:
thanks @wade-arista , I was in the process of fixing this same issue, but you did a much better job than I had time for 😄

wade-arista:
> thanks @wade-arista , I was in the process of fixing this same issue, but you did a much better job than I had time for 😄

sure thing! Sorry, I had meant to post a [link to discourse](https://bazelbuild.slack.com/archives/CA306CEV6/p1709085258803439) where we were discussing a PR, but then I forgot.

Hopefully this is the same thing fix you had in mind?

scasagrande:
Yeah I believe this covers my case but I will have to double check just to make sure. I had a quick fix for my testing so I'll just double check that it covers my case.

I did notice a further issue in my testing. After patching importing a py_libary from another workspace, I saw that the dependencies of this library not being picked up correctly. I'll take a look if that is still the case and raise a separate issue if so.

PR #310, opened 2024-03-18
Fix for external deps that use repo root relative imports

scasagrande:
---

### Type of change

### Test plan

### Details

As a follow-up to https://github.com/aspect-build/rules_py/pull/299 , this PR enables the use of external dependencies that use workspace root relative imports.

This just expands current functionality to also cover the case where the dependency is from an external workspace.

### DIFF
--- a/py/private/py_library.bzl
+++ b/py/private/py_library.bzl
@@ -135,7 +135,11 @@ def _make_imports_depset(ctx, imports = [], extra_imports_depsets = []):
     ] + [
         # Add the workspace name in the imports such that repo-relative imports work.
         ctx.workspace_name,
+        ctx.label.workspace_name,
     ]
+
+    # Filter out any empty strings
+    import_paths = [x for x in import_paths if x]
 
     return depset(
         direct = import_paths,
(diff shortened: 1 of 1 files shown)


py/private/py_library.bzl
```diff
@@ -135,8 +135,12 @@ def _make_imports_depset(ctx, imports = [], extra_imports_depsets = []):
     ] + [
         # Add the workspace name in the imports such that repo-relative imports work.
         ctx.workspace_name,
+        ctx.label.workspace_name,
     ]
 
+    # Filter out any empty strings
+    import_paths = [x for x in import_paths if x]
```
mattem, line 142: Is this because `ctx.label.workspace_name` could be an empty string?

scasagrande: yes

mattem: Instead of iterating over the full list in each call to `_make_imports_depset`, could instead this be added conditionally (rather than added then removed)?

scasagrande: If that's what you really want, then sure

PR #311, opened 2024-03-18
fix: use PyInfo from rules_python

mattem:
Switches to using `PyInfo` from rules_python, rather than relying on the builtin one from Java.

Note that this doesn't add any interoperability between the **native** `py_*` rules (java based) and the exports from rules_python. As such, using rules_python with the native java based rules and providers is no longer supported after this commit.

If we want, we could support it, but being able to consume and return both providers.

---

### Type of change

- Suggested release notes are provided below:

Use `PyInfo` provider as loaded from rules_python, and drop support for dependency edges from native `py_*` rules.

### Test plan

### DIFF
--- a/py/private/py_binary.bzl
+++ b/py/private/py_binary.bzl
@@ -1,5 +1,6 @@
 """Implementation for the py_binary and py_test rules."""
 
+load("@rules_python//python:defs.bzl", "PyInfo")
 load("@aspect_bazel_lib//lib:paths.bzl", "BASH_RLOCATION_FUNCTION", "to_rlocation_path")
 load("@aspect_bazel_lib//lib:expand_make_vars.bzl", "expand_locations", "expand_variables")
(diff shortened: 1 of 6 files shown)


alexeagle:
> If we want, we could support it

Don't we have to do that before landing? We can't just break interop with rules_python py_library's right?

mattem:
> > If we want, we could support it
> 
> Don't we have to do that before landing? We can't just break interop with rules_python py_library's right?

Perhaps the description wasn't clear. This doesn't break compatibility with the rules_python py_library, but it'll break compatibility with a `py_library` that isn't loaded from rules_python, vs the native java one _if_ pystar is enabled, as both rules will be referencing a different `PyInfo` symbol.

The line in rules_python is:

```
PyInfo = _starlark_PyInfo if config.enable_pystar else BuiltinPyInfo
```

So, if pystar is enabled, and the user uses `native.py_library`, then this will cause an error.

alexeagle (approved):

PR #323, opened 2024-04-02
refactor: split toolchains and repositories

thesayyn:
### Type of change

Currently, rules_py_dependencies and rules_py_toolchains are exported from the same bzl file which makes it impossible to depend on another rule and load from at the same since both macros are in the same file.  This changes this to allow adding a dependency and loading from it.

- Relevant documentation has been updated

### Test plan


thesayyn:
Done

alexeagle (approved):

Issue #236, opened 2023-12-28
[Bug]: `--build_python_zip` seems to not work in py_binary


ChangFu2000:
### What happened?

When building a py_binary with `--build_python_zip` specified, expected .zip artifact is not generated.

### Version

Development (host) and target OS/architectures:
Ubuntu 20.04

Output of `bazel --version`:
bazel 7.0.0 (bazelisk)

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
aspect_rules_py, version: 0.5.0
rules_python, version: 0.22.0

Language(s) and/or frameworks involved:
Python

### How to reproduce

```starlark
# MODULE.bazel

bazel_dep(name = "aspect_rules_py", version = "0.5.0")

bazel_dep(name = "rules_python", version = "0.22.0", dev_dependency = True)

python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(
    name = "python3",
    configure_coverage_tool = True,
    python_version = "3.11",
)
use_repo(python, "python3_toolchains")

register_toolchains(
    "@python3_toolchains//:all",
)
```

```starlark
# BUILD

load("@aspect_rules_py//py:defs.bzl", "py_binary")

py_binary(
    name = "script",
    srcs = ["script.py"],
    main = "script.py",
)
```

Running with: `bazel build //:script --build_python_zip`
```


### Any other information?

_No response_

Cartman75:
If there is a self executing python file, that would seem like a upgrade and much better then zip.. Others can tell me if I am wrong, but currently, the zip is just the best we got right now, sorta deal. I would much prefer to just execute a file like most other code can do.

ChangFu2000:
@alexeagle Thanks for the reply. We are using a docker environment so both .pyz and .pex satisfy our use case. I'm not faimliar with Bazel implementation details so please follow your decision.

mattem:
Closed via 7a9e4b2, rules_py provides pex as the zip implementation.

PR #324, opened 2024-04-02
feat: implement --build_python_zip pex

thesayyn:
### Type of change

- New feature or functionality (#236)

### Test plan

- New test cases added


MODULE.bazel
```diff
@@ -8,7 +8,7 @@ module(
 
 # Lower-bound versions of direct dependencies.
 # When bumping, add a comment explaining what's required from the newer release.
-bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
+bazel_dep(name = "aspect_bazel_lib", version = "1.42.2")
```
thesayyn, line 11: This is needed to get tar toolchain.

thesayyn:
This is kind of working now, 

Missing things i can think of;

- inject env provided via `ctx.attr.env`
- set runfiles directory so the runfiles lookup works.
- bzlmod runfiles mapping is not working. 
- only workspace relative imports (`import wksp.xyz`) and third-party (`import 3p.xyz`) imports work, we need to find a way to inject more entries into sys.path without using venv.

py/private/py_pex.bzl
```diff
@@ -0,0 +1,51 @@
...
+    workspace_name = str(ctx.workspace_name)
+
+    args.add_all(
+        srcs, 
+        map_each = lambda f: _map_srcs(f, workspace_name),
+        uniquify = True,
+        allow_closure = True,
+    )
+    args.add(ctx.file.main, format = "--executable=%s")
+    args.add("#!/usr/bin/env python3", format="--python-shebang=%s")
```
dizzy57, line 37:
```suggestion
    args.add("/usr/bin/env python3", format="--python-shebang=%s")
```

nit: `--python-shebang`'s help line says to pass shebang without `#!` (it works in both cases, whether you pass it or not)

py/private/py_binary.bzl
```diff
@@ -6,13 +6,15 @@ load("@aspect_bazel_lib//lib:expand_make_vars.bzl", "expand_locations", "expand_
 load("//py/private:py_library.bzl", _py_library = "py_library_utils")
 load("//py/private:py_semantics.bzl", _py_semantics = "semantics")
 load("//py/private/toolchain:types.bzl", "PY_TOOLCHAIN", "VENV_TOOLCHAIN")
+load("//py/private:py_pex.bzl", "build_pex")
```
mattem, line 9: Can the convention to rename on import be followed for this? _build_pex.

py/tools/pex/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
+    name = "pex",
+    srcs = ["main.py"],
+    main = "main.py",
+    deps = ["@pypi_pex//:pkg"],
```
mattem, line 7: Where does the user get this dependency from? 

rules_py currently doesn't have any Python dependencies, this will be the first, so there is no call to a pip install in the setup. This is working here as it's using the internal test time dependencies that rules_py uses.

Ideally we wouldn't have a Python dependency at all, is there another way to interact with the PEX library, can we vendor the dependency?

mattem (changes requested):

py/tools/pex/BUILD.bazel
thesayyn: You are right, my latest commit fixes this by introducing a slightly complicated bootstrapping logic to allow pex builder to be either a py_binary from rules_python with a pip dependency (for local dev only) or a py_binary from rules_python with no external dependencies but a  pex builder packaged into a pex zip to prevent a python dependency.

thesayyn: > Ideally we wouldn't have a Python dependency at all, is there another way to interact with the PEX library, can we vendor the dependency?

vendoring is another approach we could take, it's open for a debate.

py/private/py_pex.bzl
```diff
@@ -0,0 +1,51 @@
...
+def _runfiles_path(file, workspace):
+    if file.short_path.startswith("../"):
+        return file.short_path[3:]
+    else:
+        return workspace + "/" + file.short_path
+
+def _map_srcs(f, workspace):
+    site_packages_i = f.path.find("site-packages")
+    # determine if the src if a third party.
+    if site_packages_i != -1 and f.path.count("/",  site_packages_i) == 2:
```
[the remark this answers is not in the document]
thesayyn: I will explain this in detail once it is close to the finish line. i am not sure if this is the way to install third-party deps yet. I'll wait for some feedback from some internal users first.

py/tools/pex/BUILD.bazel
```diff
@@ -0,0 +1,71 @@
...
+    name = "use_dev_pex",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "should_use_dev_pex",
+    flag_values = {
+        ":use_dev_pex": "true",
+    },
+)
```
mattem, line 62: There's an existing mechanism to select between the release and dev version of tools by switching on the `IS_PRERELEASE` symbol in `tools/version.bzl`, can we hook into the same structure here, rather than adding something new?

py/tools/pex/BUILD.bazel
```diff
@@ -0,0 +1,71 @@
+load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_file")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@rules_python//python:defs.bzl", "py_binary")
+load("//py:defs.bzl", _py_binary = "py_binary")
```
mattem, line 4: Perhaps rename one symbol explicitly to `rules_python_py_binary`? It's hard to follow with just the _ rename.

py/tools/pex/BUILD.bazel
mattem: Given that pex is a single wheel with no dependencies, I'm not clear why we need to bootstrap it via a zip? 
Could we instead fetch the wheel via an `htt_file`, have rules_py expand it with the existing tools, then depend on that? rules_py has a fetch dependencies function where it declares other rule dependencies, so we could add the `http_file` into there.

thesayyn: Yeah, that sounds better, i'll do that once get back to this again.

py/tools/pex/BUILD.bazel
thesayyn: with the http_file approach, we don't need this as the pex builder always will be the same target.

dizzy57:
I was unable to make it work with bzlmod, here's a minimal repro: https://github.com/dizzy57/bazel_py_minimal

py/tools/pex/BUILD.bazel
thesayyn: This is now fixed.

examples/py_pex_binary/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
...
+        "TEST": "1"
+    },
+    deps = [
+        "@pypi_cowsay//:pkg",
+        "@bazel_tools//tools/python/runfiles",
+    ],
+)
+
+
+
```
mattem, line 18: Can these be tidied? I think I've commented on empty space in this diff before, perhaps a pass over would be good.

py/private/py_pex_binary.bzl
```diff
@@ -0,0 +1,119 @@
...
+    args = ctx.actions.args()
+
+    # Copy workspace name here to prevent ctx
+    # being transferred to the execution phase.
+    workspace_name = str(ctx.workspace_name)
+
+    args.add_all(
+        ctx.attr.inject_env.items(), 
+        map_each = lambda e: "--inject-env=%s=%s" % (e[0], e[1]),
+        allow_closure = True,
```
mattem, line 59: Does this need to capture the closure? Seems like it only references the var in the lambda.

py/private/py_pex_binary.bzl
```diff
@@ -0,0 +1,119 @@
...
+    ctx.actions.run(
+        executable = ctx.executable._pex,
+        inputs = runfiles.files,
+        arguments = [args],
+        outputs = [output],
+        mnemonic = "PyPex",
+        progress_message = "Building PEX binary %{label}",
+        # Unfortunately there is no way to disable pex cache, so just set it to . allow
+        # bazel to discard cache once the action is done. 
+        # TODO: this is probably not the right thing to do if the action is unsandboxed.
```
mattem, line 87: What is the right thing to do? Should we create an issue rather than an TODO in the code that will never be todone?

py/repositories.bzl
```diff
@@ -42,4 +53,40 @@ def rules_py_dependencies():
...
+    Args:
+        name: prefix used in created repositories
+        register: whether to call the register_toolchains, should be True for WORKSPACE and False for bzlmod.
+        is_prerelease: True iff there are no pre-built tool binaries for this version of rules_py
+    """
+
+    # the url and digest information can be found at https://pypi.org/pypi/pex/json
+    # WARNING: when updated, also update MODULE.bazel
+    http_file(
+        name = "pex_2_3_1",
```
mattem, line 73: Do you think we should namespace this more? `rules_py_pex_231` for example?

py/repositories.bzl
```diff
@@ -42,4 +53,40 @@ def rules_py_dependencies():
...
+def rules_py_toolchains(name = DEFAULT_TOOLS_REPOSITORY, register = True, is_prerelease = IS_PRERELEASE):
+    """Create a downloaded toolchain for every tool under every supported platform.
+
+    Args:
+        name: prefix used in created repositories
+        register: whether to call the register_toolchains, should be True for WORKSPACE and False for bzlmod.
+        is_prerelease: True iff there are no pre-built tool binaries for this version of rules_py
+    """
+
+    # the url and digest information can be found at https://pypi.org/pypi/pex/json
```
mattem, line 70:
```suggestion
    # The url and digest information can be found at https://pypi.org/pypi/pex/json
```

py/private/py_pex_binary.bzl
thesayyn: i must address this before landing this PR.

py/private/py_pex_binary.bzl
thesayyn: nope, left over. will remove

thesayyn:
Also, CPython=={major}.{minor}.* should be the default constraint.

py/private/py_pex_binary.bzl
thesayyn: done

thesayyn:
@mattem I have removed some unrelated diff from my failed attempt before using PEX.  Now diff should be smaller and easier to understand.

py/private/py_pex_binary.bzl
```diff
@@ -0,0 +1,158 @@
...
+        DefaultInfo(files = depset([output]), executable = output),
+        # See: https://github.com/bazelbuild/bazel/blob/b4ab259fe1cba8a108f1dd30067ee357c7198509/src/main/starlark/builtins_bzl/common/python/py_executable_bazel.bzl#L265
+        OutputGroupInfo(
+            python_zip_file = depset([output])
+        )
+    ]
+
+
+_attrs = dict({
+    "binary": attr.label(executable = True, cfg = "target"),
```
dizzy57, line 123:
```suggestion
    "binary": attr.label(executable = True, mandatory = True, cfg = "target"),
```
Building a pex without a `binary` doesn't make much sense.

njlr: I think that a PEX can be used as an environment only - IIUIR the entry point(s) are option in the PEX tool https://docs.pex-tool.org/buildingpex.html#specifying-entry-points

py/private/py_pex_binary.bzl
```diff
@@ -0,0 +1,158 @@
...
+def _runfiles_path(file, workspace):
+    if file.short_path.startswith("../"):
+        return file.short_path[3:]
+    else:
+        return workspace + "/" + file.short_path
+
+exclude_paths = [
+    "toolchain",
+    "aspect_rules_py/py/tools/",
+    "rules_python~~python~",
```
dizzy57, line 16: > `rules_python~~python~`

Will this path change if we set `--incompatible_use_plus_in_repo_names`? It's going to be set to true by default in bazel 8.

context: https://github.com/bazelbuild/bazel/issues/23127

py/private/py_pex_binary.bzl
dizzy57: @njlr there is no `deps` attribute in this rule (and I don't think we need it), so setting `binary` is the only way to specify what goes into PEX. In any way, the rule currently fails if you don't specify `binary` attribute:
```
ERROR: [...]/cli_tool/BUILD.bazel:19:14: in py_pex_binary rule //cli_tool:empty_pex:
	File "[...]/external/aspect_rules_py~/py/private/py_pex_binary.bzl", line 61, column 22, in _py_python_pex_impl
		runfiles = binary[DefaultInfo].data_runfiles
Error: type 'NoneType' has no operator [](Provider)
```

py/private/py_pex_binary.bzl
thesayyn: Yes, this will stop working.

thesayyn: Sent a filter for plus version as well. LMK if that does it.

py/private/py_pex_binary.bzl
thesayyn: done

py/private/py_pex_binary.bzl
thesayyn: Please see the comment below the argument. 

https://github.com/aspect-build/rules_py/pull/324/files#diff-bdbfb557a8b870d00314cf49db0fad2611b37d9d20d6c055034aef05b30a6e65R77

py/private/py_pex_binary.bzl
```diff
@@ -0,0 +1,160 @@
...
+    "rules_python++python+",
+    "aspect_rules_py~/py/tools/",
```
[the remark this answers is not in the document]
thesayyn: It changes based on `--incompatible_use_plus_in_repo_names` flag. Left a comment above them.

py/private/py_pex_binary.bzl
thesayyn: > # this is needed to allow passing a lambda (with workspace_name) to map_each

mattem: But this isn't that call, the `map_each` here is `lambda e: "--inject-env=%s=%s" % (e[0], e[1]),` so how is the call below related?

dizzy57: Lambda for `inject_env`'s `map_each` does not contain any references to the `workspace_name`.

thesayyn: i could get rid of it by creating a global `def`, this feels more localized and i did not see any performance degradation.

py/private/py_pex_binary.bzl
```diff
@@ -0,0 +1,163 @@
...
+        outputs = [output],
+        mnemonic = "PyPex",
+        progress_message = "Building PEX binary %{label}",
+    )
+
+    return [
+        DefaultInfo(files = depset([output]), executable = output),
+        # See: https://github.com/bazelbuild/bazel/blob/b4ab259fe1cba8a108f1dd30067ee357c7198509/src/main/starlark/builtins_bzl/common/python/py_executable_bazel.bzl#L265
+        OutputGroupInfo(
+            python_zip_file = depset([output])
```
dizzy57, line 122: Do we need to return `python_zip_file` output group here?

Looks like `py_pex_binary` is a rule on it's own  by now and does not need any output groups. I think it's fine to return just the `DefaultInfo` provider.

py/private/py_pex_binary.bzl
dizzy57: https://bazel.build/rules/lib/builtins/Args#parameters_1 says
> To avoid unintended retention of large analysis-phase data structures into the execution phase, the map_each function must be declared by a top-level def statement; [...]

so I think it's better to rewrite it as a global `def`

mattem: In that case, can the comment be updated, as it's unrelated to what is happening in his call.

py/private/py_pex_binary.bzl
thesayyn: you are right, it's not necessary.

py/private/py_pex_binary.bzl
thesayyn: fixed the comment.

py/private/py_pex_binary.bzl
thesayyn: removed.

mattem (approved):

Issue #331, opened 2024-04-16
[Bug]: cannot use `py_venv`


KoltesDigital:
### What happened?

I've just added `bazel_dep(name = "aspect_rules_py", version = "0.7.1")` to make a venv, in a project where there already was `rules_python` with pip and so on: was working well.

But I'm unable to declare a `py_venv`:

```text
ERROR: error loading package '@@aspect_rules_py~//py/tools/venv_bin': Unable to find package for @@[unknown repo 'rules_rust' requested from @@aspect_rules_py~]//rust:defs.bzl: The repository '@@[unknown repo 'rules_rust' requested from @@aspect_rules_py~]' could not be resolved: No repository visible as '@rules_rust' from repository '@@aspect_rules_py~'.
ERROR: X:/external/aspect_rules_py~/py/private/toolchain/venv/BUILD.bazel:3:17: error loading package '@@aspect_rules_py~//py/tools/venv_bin': Unable to find package for @@[unknown repo 'rules_rust' requested from @@aspect_rules_py~]//rust:defs.bzl: The repository '@@[unknown repo 'rules_rust' requested from @@aspect_rules_py~]' could not be resolved: No repository visible as '@rules_rust' 
from repository '@@aspect_rules_py~'. and referenced by '@@aspect_rules_py~//py/private/toolchain/venv:venv_toolchain_source'
```

I tried to untangle the code to find the dependency chain, but I'm out of luck.

### Version

Development (host) and target OS/architectures: **windows amd64**

Output of `bazel --version`: **bazel 7.1.1**

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:  
bazel_dep(name = "**aspect_rules_py**", version = "**0.7.1**")
bazel_dep(name = "**rules_python**", version = "**0.31.0**")

Language(s) and/or frameworks involved: N/A

### How to reproduce

Simply put

```starlark
load("@aspect_rules_py//py:defs.bzl", "py_venv")

py_venv(
    name = "venv",
)
```

and build it.

### Any other information?

I'm migrating to bzlmod: my WORKSPACE is empty.

mattem:
Do you have the toolchain setup? https://github.com/aspect-build/rules_py/blob/main/e2e/use_release/MODULE.bazel#L8-L10

It's missing from the setup on the releases page, will add.

alexeagle:
I'm not able to reproduce.

```
rules_py_snip % bazel build :all
[...]

PR #332, opened 2024-04-19
fix: add toolchain setup to bzlmod release snippet

mattem:
Adds the bzlmod `use_extension` to the release instructions snippet. Will look something like:

## Using [Bzlmod] with Bazel 6:

Add to your `MODULE.bazel` file:

```starlark
bazel_dep(name = "aspect_rules_py", version = "")

tools = use_extension("@aspect_rules_py//py:extensions.bzl", "py_tools")
tools.rules_py_tools()
use_repo(tools, "rules_py_tools")
```

And also register a Python toolchain, see rules_python. For example:

```starlark
# Minimum version needs:
# feat: add interpreter_version_info to py_runtime by @mattem in #1671
bazel_dep(name = "rules_python", dev_dependency = True, version = "0.29.0")

python.toolchain(
    configure_coverage_tool = True,
    python_version = "3.11",
)
```

[Bzlmod]: https://bazel.build/build/bzlmod

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -53,6 +53,13 @@ Add to your \`MODULE.bazel\` file:
 
 \`\`\`starlark
 bazel_dep(name = "aspect_rules_py", version = "${TAG:1}")
+
+EOF
+
+# Collect the tools snippet from the use_release module file.
+awk 'f;/--SNIP--/{f=1}' e2e/use_release/MODULE.bazel
+
+cat << EOF
 \`\`\`
 
 And also register a Python toolchain, see rules_python. For example:
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,5 +1,5 @@
 bcr_test_module:
-  module_path: "e2e/smoke"
+  module_path: "e2e/use_release"
   matrix:
     bazel: ["7.x", "6.x"]
     # TODO(#9): add windows to this list
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,13 @@ jobs:
   test:
     uses: bazel-contrib/.github/.github/workflows/bazel.yaml@v6
     with:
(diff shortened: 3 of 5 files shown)


.github/workflows/release_prep.sh
```diff
@@ -53,6 +53,13 @@ Add to your \`MODULE.bazel\` file:
 
 \`\`\`starlark
 bazel_dep(name = "aspect_rules_py", version = "${TAG:1}")
+
+EOF
+
+# Collect the tools snippet from the use_release module file.
+awk 'f;/--SNIP--/{f=1}' e2e/use_release/MODULE.bazel
```
alexeagle, line 60: I expected this would just take the snippet from the use_release e2e INSTEAD of the smoke e2e, as each one should be fully self-contained there's no reason we have to compose more than one of them

alexeagle (approved):

.github/workflows/release_prep.sh
mattem: The `use_release` didn't have any of the Python toolchain stuff in. I don't know if it needs it, although, it's working.

alexeagle: That sounds like a bug in `use_release` then. If it's not complete then it doesn't work. I'll TAL

alexeagle: Yup it's already red, I missed adding it to the test matrix https://github.com/aspect-build/rules_py/blob/main/.github/workflows/ci.yaml#L23

alexeagle:
Also the error in #331 is about rules_rust being absent, so I don't think this is the fix for it. Maybe we should repro first and figure out if we do have a bug and why CI is green?

mattem:
> Also the error in #331 is about rules_rust being absent, so I don't think this is the fix for it. Maybe we should repro first and figure out if we do have a bug and why CI is green?

Right, because the toolchain isn't registered, so it's attempting to build from source, which it can't.

alexeagle:
Ah okay I will look again. I was not able to repro yet.

alexeagle:
We register the toolchain for users here:

PR #334, opened 2024-04-29
repo for symlink collision

vors:
This is a repro for the following problem: when we create venv and symlink different unpackaged wheels into it we can run into some silly (could be ignored) and perhaps not so silly (could not be ignored and should be a legit error) collisions.


mattem:
Thanks!

PR #336, opened 2024-04-29
fix: ignore conflicting symlinks at site-packages root on venv creation

### LINKED ISSUE #335: [Bug]: venv symlink creation can run into collisions
### What happened?

If there are wheels who define the same files then venv creation will fail.
Sometimes this is desirable, but often collision is on something silly like CHANGELOG.md and it would be good to somehow allow such problem to resolve without patching wheels

Repro https://github.com/aspect-build/rules_py/pull/334

### Version

Output of `bazel --version`:

bazel 7.0.2

0.7.2

### How to reproduce

```shell
https://github.com/aspect-build/rules_py/pull/334
```

### Any other information?

_No response_

mattem:
Handle conflicting symlinks when creating the venv directory so that the last link wins if the existing link can be removed.
Also bails early with better error message when conflicts that aren't at the root, or caused by namespace packages are found.

Closes #335

---

Handle conflicting symlinks at venv root during venv creation.

### Test plan

- New test cases added

Force a conflict and observe error message.


alexeagle (approved):

vors:
🙌

PR #347, opened 2024-06-11
feat: support python_version attribute

thesayyn:
This PR adds support python_version attribute the same way rules_python does, with the exception that we generate no `defs.bzl` file that users can just replace their with. Instead we encourage them to create their own macros that sets the desired version on py_binary/py_test rule.

 
### Test plan

- New test cases added


MODULE.bazel
```diff
@@ -13,6 +13,14 @@ bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_python", version = "0.29.0")
 bazel_dep(name = "platforms", version = "0.0.7")
 
+
+# Custom python version for testing only
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(
+    is_default = False,
```
alexeagle, line 20: I was expecting this one to be explicitly named something ... I guess that's not how rules_python works eh?

py/private/py_binary.bzl
```diff
@@ -150,17 +185,33 @@ _attrs = dict({
...
 
+def _python_version_transition_impl(_, attr):
+    if not attr.python_version:
+        return {}
+    return {"@rules_python//python/config_settings:python_version": str(attr.python_version)}
+
+_python_version_transition = transition(
+    implementation = _python_version_transition_impl,
+    inputs = [],
+    outputs = ["@rules_python//python/config_settings:python_version"],
```
alexeagle, line 204: I dont see this https://github.com/bazelbuild/rules_python/blob/main/python/config_settings/BUILD.bazel - does that mean we need a change to rules_python to land and get released first?

MODULE.bazel
thesayyn: this directly from their example, i don't see any name attribute, so no 🤷🏻‍♂️

py/private/py_binary.bzl
thesayyn: it is created by this macro, https://github.com/bazelbuild/rules_python/blob/49cdf7d3fe000076d6432a34238e5d25f5b598d0/python/config_settings/BUILD.bazel#L28-L30

https://github.com/bazelbuild/rules_python/blob/49cdf7d3fe000076d6432a34238e5d25f5b598d0/python/private/config_settings.bzl#L173-L182

thesayyn: We did not need to introduce anything new to rules_python, everything works with existing rules_python version.

alexeagle: oh, neat, they were so close to getting this right :)

MODULE.bazel
alexeagle: Yeah they must have their own scheme for avoiding naming collisions in the external/ folder and also for referencing the version. Okay...

alexeagle (approved):
So simple 😍

WORKSPACE
```diff
@@ -24,6 +24,15 @@ register_toolchains("//:container_py_toolchain")
 
 load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
 
+python_register_toolchains(
+    name = "python_toolchain_3_8",
+    python_version = "3.8.12",
+    # set set_python_version_constraint makes it so that only matches py_* rule that has 
+    # this exact version set in the `python_version` attribute.
```
mattem, line 31: I don't follow this comment?

py/private/py_binary.bzl
```diff
@@ -139,6 +139,41 @@ _attrs = dict({
...
+
+python_register_toolchains(
+    name = "python_toolchain_3_8",
+    python_version = "3.8.12",
+    # setting set_python_version_constraint makes it so that only matches py_* rule  
+    # which has this exact version set in the `python_version` attribute.
+    set_python_version_constraint = True,
+)
+
+# It's important to register the default toolchain last it will match any py_* target.
```
mattem, line 161:
```suggestion
# It's important to register the default toolchain last as it will match any py_* target.
```

py/private/py_binary.bzl
```diff
@@ -139,6 +139,41 @@ _attrs = dict({
...
+)
+
+# It's important to register the default toolchain last it will match any py_* target. 
+python_register_toolchains(
+    name = "python_toolchain",
+    python_version = "3.9",
+)
+```
+
+Whereas configuring it for MODULE.bazel may look like this:
```
mattem, line 168:
```suggestion
Configuring for MODULE.bazel may look like this:
```

py/private/py_binary.bzl
```diff
@@ -139,6 +139,41 @@ _attrs = dict({
...
+
+When using WORKSPACE, this may look like this,
+
+```
+load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
+
+python_register_toolchains(
+    name = "python_toolchain_3_8",
+    python_version = "3.8.12",
+    # setting set_python_version_constraint makes it so that only matches py_* rule
```
mattem, line 156: Can this be updated to?

mattem:
Please also update the PR description with more text to explain the change

py/private/py_binary.bzl
thesayyn: i don't follow.

thesayyn: do you mean, can it be updated to some other version?

thesayyn:
> Please also update the PR description with more text to explain the change

done

WORKSPACE
thesayyn: reworded it.

py/private/py_binary.bzl
mattem: It's the similar comment that I didn't understand from above, so if that one is reworded, can this one be updated too?

WORKSPACE
```diff
@@ -24,6 +24,17 @@ register_toolchains("//:container_py_toolchain")
 
 load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
 
+python_register_toolchains(
+    name = "python_toolchain_3_8",
+    python_version = "3.8.12",
+    # setting `set_python_version_constraint` will set special constraints on the registered toolchain
+    # this means that this toolchain registration will only be selected for py_binary/py_test targets 
+    # that has `python_version = "3.8.12"` attribute set, targets that has no python_attribute will get
+    # the default toolchain resolved for the which can be seen down below.
```
mattem, line 33:
```suggestion
    # Setting `set_python_version_constraint` will set special constraints on the registered toolchain.
    # This means that this toolchain registration will only be selected for `py_binary` / `py_test` targets 
    # that have the `python_version = "3.8.12"` attribute set. Targets that have no `python_attribute` will use
    # the default toolchain resolved which can be seen below.
```

examples/multi_version/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+        ":__test__",
+    ],
+    main = ":__test__.py",
+    deps = [
+        ":__test__",
+    ],
+    python_version = "3.8.12"
+)
+
+
```
mattem, line 31:
```suggestion

```

examples/multi_version/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+py_binary(
+    name = "multi_version",
+    srcs = ["say.py"],
+    deps = [
+        "@pypi_cowsay//:pkg",
+    ],
+    python_version = "3.8.12"
+)
+
+
```
mattem, line 12:
```suggestion

```

mattem (approved):

py/private/py_binary.bzl
thesayyn: done

PR #357, opened 2024-06-27
Draft: Windows

peakschris:
This is a very early cut of windows support. 

Three test cases are passing.

I'm finding rules_rust very flaky on windows, particularly around its generation of very long filenames that do not work on windows or with msvc. The rules_rust maintainers have mitigated this to some extent in their builds by using D:\b as output_root, but the issues are hitting this repo too.

It may be that if the binaries can be cross compiled on linux, then rules_rust does not need to be a dependency on the windows platform and things might go more smoothly.

The current blocker is zstd crate that is failing to build on windows.

@alexeagle This is as far as I can go right now with this; it needs a CI or rust expert to get past the current hurdle. If that can be solved I can have another look.

One question: Can we avoid generating zip files with this ruleset for py_binary? I was hoping that this ruleset already does that, it seems not (the zip files and hermetic toolchain together are *extremely slow* to create and unzip on windows). Are there any plans to directly invoke python with system interpreter?


peakschris:
The zstd failure is very similar to this: https://gist.github.com/sbabcoc/28f3b53ae2765444100ee6a183a939b2

PR #367, opened 2024-07-12
refactor: don't default imports to current folder

### LINKED ISSUE #368: [Bug]: "." added to imports by default
### What happened?

`py_library`, `py_binary` and `py_test` macros set `imports` attribute to `["."]` by default. This pollutes top-level import namespace.
https://github.com/aspect-build/rules_py/blob/c37e67963502a401de4b117fa8104da2091b5d1c/py/defs.bzl

### Version

Output of `bazel --version`: `bazel 7.0.2`

`WORKSPACE` or `MODULE.bazel` file: `HEAD` (`v0.7.3+`)

Language(s) and/or frameworks involved: none

### How to reproduce

```shell
Check out `rules_py` repository. Apply the following patch:


diff --git a/py/tests/repo_relative_imports/test/test.py b/py/tests/repo_relative_imports/test/test.py
index 8e87756..60a33da 100644
--- a/py/tests/repo_relative_imports/test/test.py
+++ b/py/tests/repo_relative_imports/test/test.py
@@ -1,3 +1,3 @@
-from py.tests.repo_relative_imports.lib import adder
+import adder

 assert adder.add(2, 3) == 5


Run `bazel test //py/tests/repo_relative_imports/test`.

Expected result: test fails since no top-level module `adder` is installed.
Actual result: test passes
```

### Any other information?

This is extremely problematic if instead of `adder.py` your users create `numpy.py`, since `import numpy` will import this file instead of `numpy` library.
`rules_python` defaults to `import = []`

alexeagle:
This causes errors for one of our clients. @dizzy57 is planning to file a corresponding issue.

Fixes #368 

---

`py_library` no longer defaults to `imports=["."]`. we recommend using longer-form import statements to import first-party libraries. You can manually restore this value on the `imports` attribute of your libraries if you still want to import them with shorthand.

### Test plan

### DIFF
--- a/py/defs.bzl
+++ b/py/defs.bzl
@@ -14,12 +14,12 @@ py_pytest_main = _py_pytest_main
 py_venv = _py_venv
 py_binary_rule = _py_binary
 py_test_rule = _py_test
-py_library_rule = _py_library
-py_unpacked_wheel_rule = _py_unpacked_wheel
+py_library = _py_library
+py_unpacked_wheel = _py_unpacked_wheel
 
 resolutions = _resolutions
 
-def _py_binary_or_test(name, rule, srcs, main, imports, deps = [], resolutions = {}, **kwargs):
+def _py_binary_or_test(name, rule, srcs, main, deps = [], resolutions = {}, **kwargs):
     # Compatibility with rules_python, see docs in py_executable.bzl
     main_target = "_{}.find_main".format(name)
     determine_main(
@@ -34,7 +34,6 @@ def _py_binary_or_test(name, rule, srcs, main, imports, deps = [], resolutions =
         name = name,
         srcs = srcs,
         main = main_target,
-        imports = imports,
         deps = deps,
         resolutions = resolutions,
         **kwargs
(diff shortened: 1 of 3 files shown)


dizzy57:
Closes #368

py/defs.bzl
```diff
@@ -73,14 +73,14 @@ def py_binary(name, srcs = [], main = None, imports = ["."], **kwargs):
...
-def py_test(name, main = None, srcs = [], imports = ["."], **kwargs):
+def py_test(name, main = None, srcs = [], imports = [], **kwargs):
     "Identical to py_binary, but produces a target that can be used with `bazel test`."
 
     # Ensure that any other targets we write will be testonly like the py_test target
     kwargs["testonly"] = True
     _py_binary_or_test(name = name, rule = _py_test, srcs = srcs, main = main, imports = imports, **kwargs)
 
-def py_library(name, imports = ["."], **kwargs):
+def py_library(name, imports = [], **kwargs):
```
mattem, line 83: Given that nothing reads `imports` now, let's remove it from the method signatures and default it to an empty list on the rule attr. It can be passed here via kwargs.

mattem (approved):

Issue #358, opened 2024-06-29
[Bug]: Conflicting symlinks found when attempting to create venv.


anuragkanungo:
### What happened?

Using version 0.7.3 I am seeing

```
-----------------------------------------------------------------------------
Error:   × Unable to run command:
  ├─▶ Conflicting symlinks found when attempting to create venv. More than one
  │   package provides the file at these paths
  ├─▶ pip_deps_mcap_protobuf_support/site-packages/tests/test_writer.py
  ╰─▶ site-packages/tests/test_writer.py
  ```
  
  These are the files
  
  ```
./bazel-voxel/external/pip_deps_mcap/site-packages/tests/test_writer.py
./bazel-voxel/external/pip_deps_mcap_protobuf_support/site-packages/tests/test_writer.py
```
  
  In a typical pip installation the second file overwrites the first one.

### Version

Development (host) and target OS/architectures: Linux x86-64

Output of `bazel --version`: bazel 7.1.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 0.7.3

Language(s) and/or frameworks involved:

### How to reproduce

```shell
Adding these pip packages

mcap==0.0.15
mcap-protobuf-support==0.0.9
```
```


### Any other information?

_No response_

**Comment by mattem** [comment]
> In a typical pip installation the second file overwrites the first one.

Are the file contents checked before overwriting, or only the name taken into consideration? If so, then which one wins? My concern is that the ordering of the dependencies matter here, and rules_py will need to ensure the same dependency ordering as pip, which may not be possible.

In this case, both those libraries should not be shipping test cases into the distribution package.

**Comment by hofbi** [comment]
I get a similar error

```shell
Error:   × Unable to run command:
  ├─▶ Conflicting symlinks found when attempting to create venv. More than one
  │   package provides the file at these paths
  ├─▶ rules_python~~pip~rules_ros_pip_deps_38_pyyaml/site-packages/yaml/
  │   loader.py
  ╰─▶ site-packages/yaml/loader.py
```

Allowing these conflicts and printing a warning seems like a reasonable approach.

antspy:
Is the intended solution here to use package_collisions='ignore' any time this happens? Or should we wait for a long term fix? 

(I am hitting this when installing `pandera[polars]` and `polars`).

PR #369, opened 2024-07-17
fix: allow symlink collision error to be downgraded

mattem:
Introduces an attr allowing the symlink collision errors to be downgraded to a warning or ignored.

Closes https://github.com/aspect-build/rules_py/issues/358

---

Allow the "conflicting symlinks" error when creating a venv to be downgraded to a warning and ignored.

### Test plan

Run `//py/tests/virtual/django:manage` with the `pyqt6` packages installed.


py/private/py_binary.bzl
```diff
@@ -172,7 +173,18 @@ python = use_extension("@rules_python//python/extensions:python.bzl", "python")
 python.toolchain(python_version = "3.8.12", is_default = False)
 python.toolchain(python_version = "3.9", is_default = True)
 ```
-"""
+""",
+    ),
+    "package_collision": attr.string(
```
mattem, line 178: 🧵 Naming?

Should we allow this to be set via a Starlark flag globally?

py/private/py_binary.bzl
```diff
@@ -172,7 +173,18 @@ python = use_extension("@rules_python//python/extensions:python.bzl", "python")
...
+        doc = """The action that should be taken when a symlink collision is encounted when creating the venv.
+A collision can occour when multiple packages providing the same file are installed into the venv. The possible values are:
+
+* "error": (default) When conflicting symlinks are found, an error is reported and venv creation halts.
+* "warning": When conflicting symlinks are found, an warning is reported, however venv creation continues.
+* "ignore": When conflicting symlinks are found, no message is reported and venv creation continues.
+        """,
+        default = "error",
```
mattem, line 186: The current default is to error, but perhaps it should be warning?

py/tools/py/src/pth.rs
```diff
@@ -144,11 +148,45 @@ fn create_symlink(e: &DirEntry, root_dir: &Path, dst_dir: &Path) -> miette::Resu
...
+        let d2 = MietteDiagnostic::new(format!("site-packages/{}", path_to_conflict))
+            .with_severity(severity);
+        let d3 = MietteDiagnostic::new(next_conflict).with_severity(severity);
+
+        miette!(d3).wrap_err(d2).wrap_err(d)
+    }
+
+    if link.exists() {
+        // If the link already exists and is the same file, then there is no need to link this new one.
+        // Assume that if the files are the same, then there is no need to warn or error.
```
mattem, line 162: Should this still warning even if they are the same file?

py/private/py_binary.bzl
alexeagle: Starlark flags discard the analysis cache, so I think we try to avoid. I think the common case is that one venv has a collision but others in that repository do not? In that case I think it's okay to mark it on each one.

py/private/py_binary.bzl
alexeagle: if the error message tells you how to suppress, that's most of the DX right there. It's nice to force users to acknowledge they're aware of the collision, so if it actually breaks their program they are more likely to remember a plausible cause.

py/tools/py/src/pth.rs
alexeagle: There would be no action for the user to take, right? Then I don't see the use in warning them unnecessarily.

py/private/py_binary.bzl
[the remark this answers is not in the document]
mattem: What would true / false be given that the option is a tri state?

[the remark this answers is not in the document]
alexeagle: it would either be error, then when I acknowledge it would be silent. No warning.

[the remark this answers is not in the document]
mattem: Warning might be useful to discover new collisions, as error will bail at the first issue, where as warning will show all collisions. Perhaps we make error show all too?

py/tools/py/src/pth.rs
```diff
@@ -144,11 +148,45 @@ fn create_symlink(e: &DirEntry, root_dir: &Path, dst_dir: &Path) -> miette::Resu
             .and_then(|s| s.split_once(".runfiles/"))
             .map(|s| s.1)
             .unwrap();
-        return Err(
-            miette!(format!("site-packages/{}", path_to_conflict))
-            .wrap_err(format!("{}", next_conflict))
-            .wrap_err("Conflicting symlinks found when attempting to create venv. More than one package provides the file at these paths")
-        );
+
+        let d = MietteDiagnostic::new("Conflicting symlinks found when attempting to create venv. More than one package provides the file at these paths".to_string()).with_severity(severity);
```
mattem, line 152: Include how to ignore this message.

py/private/py_binary.bzl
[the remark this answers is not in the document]
mattem: Keeping it as a enum / string and presenting all warnings to the user I think gives the most flexibility. If users would like this to be a bool, they are wrap the rule in a macro that maps true / false to the desired values.

py/private/py_binary.bzl
```diff
@@ -172,7 +173,18 @@ python = use_extension("@rules_python//python/extensions:python.bzl", "python")
 python.toolchain(python_version = "3.8.12", is_default = False)
 python.toolchain(python_version = "3.9", is_default = True)
 ```
-"""
+""",
+    ),
+    "package_collisions": attr.string(
+        doc = """The action that should be taken when a symlink collision is encounted when creating the venv.
```
thesayyn, line 179:
```suggestion
        doc = """The action that should be taken when a symlink collision is encountered when creating the venv.
```

py/private/py_binary.bzl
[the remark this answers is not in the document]
alexeagle: yeah this is a fine answer, my "WDYT" there was intended literally

PR #374, opened 2024-07-31
example: Bump versions in uv example

hofbi:
---

### Test plan

- Covered by existing test cases

PR #375, opened 2024-07-31
example: Add example for pybind11

hofbi:
Add an example for pybind11 with rules_py which is a modified version of https://github.com/pybind/pybind11_bazel/tree/master/examples/basic.

---

### Test plan

- New test cases added


alexeagle:
Thanks! but it doesn't appear this is being tested by our CI. We also don't have sub-modules in the other examples folders. Could you move the new dependency to the root MODULE.bazel (with `dev_dependency=True`) and verify this is getting tested?

hofbi:
Can do.

Isn't the UV example I am bumping in #374 a submodule?

alexeagle (approved):

Issue #391, opened 2024-09-09
[Bug]: py_pex_binary-rule fails to include single-file module pypi-deps


ekfeet:
### What happened?

Hello,

I am trying to use the relatively new `py_pex_binary`-rule, but the generated pex fails when I try to run it. The error I get:
```
Examined the following interpreters:
1.) /usr/bin/python3.11 CPython==3.11.2

No interpreter compatible with the requested constraints was found:

  Failed to resolve requirements from PEX environment @ /home/[usr]/.pex/unzipped_pexes/6b1ca5c199fa0394c40fdcd89179767251885ff9.
  Needed cp311-cp311-manylinux_2_36_x86_64 compatible dependencies for:
   1: six
      Required by:
        isodate 0.6.1
      But this pex had no ProjectName(raw='six', validated=False, normalized='six') distributions.
```
I've seen this issue when depending on packages that depends on `six` and `typing-extensions`. The regular binary runs just fine.

The .deps/ in the unzipped pex only shows isodate:
```
ls ~/.pex/unzipped_pexes/6b1ca5c199fa0394c40fdcd89179767251885ff9/.deps/
isodate-0.6.1
```
The metadata of isodate shows that six is a dependency
```
rg six ~/.pex/unzipped_pexes/6b1ca5c199fa0394c40fdcd89179767251885ff9/.deps/isodate-0.6.1/isodate-0.6.1.dist-info/
/home/[usr]/.pex/unzipped_pexes/6b1ca5c199fa0394c40fdcd89179767251885ff9/.deps/isodate-0.6.1/isodate-0.6.1.dist-info/METADATA
26:Requires-Dist: six
```

Any help is much appreciated! 

### Version

Development (host) and target OS/architectures: Debian GNU/Linux 12/x86_64   

Output of `bazel --version`: bazel 7.2.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
```
# https://github.com/bazelbuild/rules_python/releases/
bazel_dep(name = "rules_python", version = "0.35.0")

# https://github.com/aspect-build/rules_py/releases
bazel_dep(name = "aspect_rules_py", version = "0.8.0")

python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(
    configure_coverage_tool = True,
    python_version = "3.11",
)

pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "pypi",
[...]

PR #392, opened 2024-09-13
fix: Handle single-file module pypi-deps in py_pex_binary

ekfeet:
Addresses the bug in https://github.com/aspect-build/rules_py/issues/391.   

Single-file modules such as `six` and `typing-extensions` does not work with the current `py_pex_binary` rule, since it's assumed that there exist a sub-directory within `site-packages` that is not dist-info. 

An example of what a single-file module pypi-dep looks like: 
```
ls bazel-reppro_pex_err/external/rules_python~~pip~pypi_311_six/site-packages/
__init__.py  six-1.16.0.dist-info  six.py
```

Since `Distribution.load(..)` takes in the `site-packages` directory, we only emit this part of the path, and let `uniquify=True` handle deduplication after `_map_srcs` is applied. 

---

### Test plan

Add `six` to requirements and `"@pypi_six//:pkg"` as a dep to the py_pex_binary example; then import `six` in py_pex_binary's `say.py`. Printing the module or `cowsay.cow(f"{six}")` shows that the previous example and single-files modules now also work.


mattem:
Thanks! Would you be able to turn the manual test case into an automated one?

@thesayyn Can you take a look?

ekfeet:
> Thanks! Would you be able to turn the manual test case into an automated one?
Looks to be a bit more work than just adding a test case, as I don't see any tests for the py-pex-binary rule. I'll leave test automation up to you guys if that's okay

py/tools/pex/main.py
```diff
@@ -159,7 +159,7 @@ def __call__(self, parser, namespace, value, option_str=None):
 ]
 
 for dep in options.dependencies:
-    dist = Distribution.load(dep + "/../")
+    dist = Distribution.load(dep)
```
thesayyn, line 162: As i remember, Distribution.load has always wanted to read a dist-info, how does this work now?

thesayyn:
Could you also add a test for this?

py/tools/pex/main.py
ekfeet: It should be functionally equivalent to the previous version I believe, but I might be missing something.  `_map_srcs` now emits the path up to  (and including) 'site-packages', instead of 'site-packages/some-package', so `+/../` isn't needed anymore

thesayyn: right, `options.dependencies` in fact  just site-packages due to `/../`, so this change ought to be functionally equivalent. I am just curious if pex is able to package bare deps like this one where there is no directory but a single file.

thesayyn:
@ekfeet could you rebase this?

thesayyn:
> Gave it a shot in [a11b203](https://github.com/aspect-build/rules_py/pull/392/commits/a11b2031ea595d65e432914664ff7e694090f56a)

it seems to be failing due to host not having a specific version of python,  could we just remove the interpreter constraints so it works with any version of python?

ekfeet:
> @ekfeet could you rebase this?

Rebased and ran `bazel run //:gazelle_python_manifest.update` 

> it seems to be failing due to host not having a specific version of python, could we just remove the interpreter constraints so it works with any version of python?

Like this https://github.com/aspect-build/rules_py/pull/392/commits/4765a62386efc98423eecd178a281b29202c18a6? Seems to work

alexeagle (approved):
Thanks! @thesayyn says this is good to go

PR #396, opened 2024-10-01
refactor: py_venv macro improved defaults for usability

### LINKED ISSUE #395: [Bug]: VSCode has trouble with virtualenvs in monorepo subfolder
### What happened?

In an example app I `bazel run //app:app_bin.venv` to create a `/app/.app_bin.venv` folder in my source tree.

However pointing VSCode there does not work. I think https://github.com/microsoft/vscode-python/issues/14592#issuecomment-895945452 explains it. we use `rip` Rust crate to create the virtualenv and don't have a `--copies` option.

https://code.visualstudio.com/docs/python/environments explains how the integration is meant to work.

If the `py_binary` target is in the repository root, then it works. However it still doesn't provide a nice name for the venv in VSCode bottom bar, making it hard to see "am I in the right one".

### Version

0.8.0 (latest)

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
After testing real developer workflows in vscode, this way works correctly and is much more practical.

Fixes #395
Screenshot 2024-10-01 at 2 55 13 PM

---

### Test plan

Try a few things in VScode:

1. in the root, run `bazel run app:app_bin.venv` and see VSCode offers to open the resulting virtual env, labels it well
2. open the app folder in the editor, then run `bazel run app_bin.venv` which also is detected
3. run `bazel run app_test.venv` and the editor offers to switch
4. open a terminal in vscode and run `python -m pytest` and the tests run correctly, meaning they should also be debuggable

### DIFF
--- a/py/private/py_venv.bzl
+++ b/py/private/py_venv.bzl
@@ -140,9 +140,12 @@ A collision can occour when multiple packages providing the same file are instal
 )
 
 def py_venv(name, **kwargs):
+    # By default, support VSCode (and other tools) by creating virtualenv's in the root of the workspace.
+    # Editors expect to locate them here, and provide a nice name to see "which one is open".
+    # See https://github.com/aspect-build/rules_py/issues/395
+    default_venv_name = ".{}".format(paths.join(native.package_name(), name).replace("/", "+"))
     _py_venv(
         name = name,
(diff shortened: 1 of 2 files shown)


gregmagolan (approved):
🌮

PR #399, opened 2024-10-02
fix: add missing python version flag attr to pex

mattem:
Adds the missing `_interpreter_version_flag` attr to the pex binary rule that allows it to resolve the correct Python toolchain in older versions of bazel.

---

Ensure `py_pex_binary` can locate the Python version on the toolchain correctly on older bazel versions.

### Test plan

Run test suite with bazel 6.5.0

### DIFF
--- a/py/private/py_pex_binary.bzl
+++ b/py/private/py_pex_binary.bzl
@@ -19,12 +19,12 @@ exclude_paths = [
     "aspect_rules_py~/py/tools/",
     # these will match in bzlmod setup with --incompatible_use_plus_in_repo_names flag flipped.
     "rules_python++python+",
-    "aspect_rules_py+/py/tools/"
+    "aspect_rules_py+/py/tools/",
 ]
 
 # determines if the given file is a `distinfo`, `dep` or a `source`
 # this required to allow PEX to put files into different places.
-# 
+#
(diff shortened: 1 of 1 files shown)


py/private/py_pex_binary.bzl
```diff
@@ -37,23 +37,24 @@ def _map_srcs(f, workspace):
...
     # No Match: `external/rules_python~~pip~pypi_39_rtoml/site-packages/rtoml-0.11.0/src/mod/parse.py`
     # Reason: It has three `/` after first `site-packages` substring.
     if site_packages_i != -1 and f.path.count("/", site_packages_i) == 2:
         if f.path.find("dist-info", site_packages_i) != -1:
             return ["--distinfo={}".format(f.dirname)]
         return ["--dep={}".format(f.dirname)]
 
-    # If the path does not have a `site-packages` in it, then put it into 
-    # the standard runfiles tree.
+        # If the path does not have a `site-packages` in it, then put it into
```
CaerusKaru, line 55: Is the spacing correct here?

CaerusKaru (approved):

py/private/py_pex_binary.bzl
mattem: 🤷 The formatter did it.

PR #400, opened 2024-10-02
chore(docs): improved navigation and structure for API docs

alexeagle:
Follows pattern from our other rulesets, makes it easier to link users to docs.

---

### Test plan
None


docs/migrating.md
```diff
@@ -0,0 +1,10 @@
+# Migrating rules from rules_python to rules_py
+
+rules_py tries to closely mirror the API of rules_python, so a migration is a "drop-in replacement" for the majority of use cases.
```
mattem, line 3: Remove filler words "so", if a word there is really necessary, use "therefore"

```suggestion
rules_py tries to closely mirror the API of rules_python, a migration is a "drop-in replacement" for the majority of use cases.
```

docs/migrating.md
```diff
@@ -0,0 +1,10 @@
...
+
+## Replace load statements
+
+Instead of loading from `@rules_python//python:defs.bzl`, you load from `@aspect_rules_py//py:defs.bzl`.
```
mattem, line 7: Same rule as "we", remove "you" for technical writing.

```suggestion
Instead of loading from `@rules_python//python:defs.bzl`, load from `@aspect_rules_py//py:defs.bzl`.
```

docs/migrating.md
```diff
@@ -0,0 +1,10 @@
...
+The rest of the BUILD file can remain the same.
+
+If you use Gazelle, see the note on [using with Gazelle](/README.md#using-with-gazelle)
```
mattem, line 10:
```suggestion
If using Gazelle, see the note on [using with Gazelle](/README.md#using-with-gazelle)
```

py/private/py_venv.bzl
```diff
@@ -1,4 +1,10 @@
-"""Implementation for the py_binary and py_test rules."""
+"""Create a Python virtualenv directory structure.
+
+Uses https://github.com/prefix-dev/rip.
```
mattem, line 3: Do we need to mention this? Seems like an implementation detail that's hidden from the end user.

alexeagle: We don't have to, but I think it's very interesting. I'll make it clear that you cannot depend on it.

mattem (approved):

PR #401, opened 2024-10-03
feat: port pytest_test macro

alexeagle:
Comes from https://github.com/caseyduquettesc/rules_python_pytest/pull/13 It's much easier to properly gazelle-generate a BUILD file in this shape rather than the awkward way you're forced to hold our py_pytest_main.

### DIFF
--- a/docs/py_test.md
+++ b/docs/py_test.md
@@ -95,7 +95,7 @@ py_pytest_main wraps the template rendering target and the final py_library.
 ## py_test
 
 <pre>
-py_test(<a href="#py_test-name">name</a>, <a href="#py_test-srcs">srcs</a>, <a href="#py_test-main">main</a>, <a href="#py_test-kwargs">kwargs</a>)
+py_test(<a href="#py_test-name">name</a>, <a href="#py_test-srcs">srcs</a>, <a href="#py_test-main">main</a>, <a href="#py_test-pytest_main">pytest_main</a>, <a href="#py_test-kwargs">kwargs</a>)
 </pre>
 
 Identical to [py_binary](./py_binary.md), but produces a target that can be used with `bazel test`.
@@ -108,6 +108,7 @@ Identical to [py_binary](./py_binary.md), but produces a target that can be used
 | <a id="py_test-name"></a>name |  Name of the rule.   |  none |
 | <a id="py_test-srcs"></a>srcs |  Python source files.   |  <code>[]</code> |
 | <a id="py_test-main"></a>main |  Entry point. Like rules_python, this is treated as a suffix of a file that should appear among the srcs. If absent, then <code>[name].py</code> is tried. As a final fallback, if the srcs has a single file, that is used as the main.   |  <code>None</code> |
+| <a id="py_test-pytest_main"></a>pytest_main |  If set, generate a [py_pytest_main](#py_pytest_main) script and use it as the main. The deps should include the pytest package (as well as the coverage package if desired).   |  <code>False</code> |
 | <a id="py_test-kwargs"></a>kwargs |  additional named parameters to <code>py_binary_rule</code>.   |  none |
 
 
--- a/e2e/use_release/src/BUILD.bazel
+++ b/e2e/use_release/src/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@aspect_rules_py//py:defs.bzl", "py_binary", "py_pytest_main", "py_test")
+load("@aspect_rules_py//py:defs.bzl", "py_binary", "py_test")
 
 py_binary(
     name = "main",
@@ -9,25 +9,14 @@ py_binary(
     main = "__main__.py",
 )
 
-# TODO(alex): land https://github.com/aspect-build/rules_py/pull/401 and shorten this
-py_pytest_main(
-    name = "__test__",
-    data = ["//:.coveragerc"],
-    deps = [
-        "@pip//coverage",
-        "@pip//pytest",
-    ],
-)
-
 py_test(
     name = "test",
-    srcs = [
-        "my_test.py",
-        ":__test__",
-    ],
-    main = ":__test__.py",
+    srcs = ["my_test.py"],
+    data = ["//:.coveragerc"],
+    pytest_main = True,
     deps = [
-        ":__test__",
         ":main",
+        "@pip//coverage",  # keep
+        "@pip//pytest",  # keep
     ],
 )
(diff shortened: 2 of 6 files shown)


pytest/pytest_shim.py
```diff
@@ -0,0 +1,71 @@
+"""A shim for executing pytest that supports test filtering, sharding, and more.
+
+Copied from https://github.com/caseyduquettesc/rules_python_pytest/blob/331e0e511130cf4859b7589a479db6c553974abf/python_pytest/pytest_shim.py
```
alexeagle, line 3: FYI @caseyduquettesc - thank you for writing this!

pytest/pytest_shim.py
```diff
@@ -0,0 +1,71 @@
...
+    # Remove any __init__.py file to avoid that.
+    # pytest.version_tuple is available since pytest 7.0
+    # https://github.com/pytest-dev/pytest/issues/9313
+    if not hasattr(pytest, "version_tuple") or pytest.version_tuple < (8, 0):
+        args = [arg for arg in args if arg.startswith("-") or os.path.basename(arg) != "__init__.py"]
+
+    if os.environ.get("XML_OUTPUT_FILE"):
+        pytest_args.append("--junitxml={xml_output_file}".format(xml_output_file=os.environ.get("XML_OUTPUT_FILE")))
+
+    # Handle test sharding - requires pytest-shard plugin.
```
mattem, line 26: Can the presence of the plugin be detected, and an error printed if it's missing?

pytest/pytest_shim.py
```diff
@@ -0,0 +1,71 @@
...
+"""
+
+import sys
+import os
+
+import pytest
+
+
+if __name__ == "__main__":
+    pytest_args = ["--ignore=external"]
```
mattem, line 13: The other template had different args around disabling caching, check if they are needed here.

pytest/defs.bzl
```diff
@@ -0,0 +1,65 @@
...
+            pytest_deps.append("@{}//pytest_shard".format(pip_repo))
+
+    py_test(
+        name = name,
+        srcs = [
+            shim_label,
+        ] + srcs,
+        main = shim_label,
+        args = [
+            "--capture=no",
```
mattem, line 61: What's the reason to have this defined here, rather than in the shim?

pytest/defs.bzl
```diff
@@ -0,0 +1,65 @@
...
+        ] + args + ["$(location :%s)" % x for x in srcs],
```
mattem, line 62: Do these automatically spill? The arg limit length could be easily hit here.

pytest/defs.bzl
```diff
@@ -0,0 +1,65 @@
...
+      pip_repo: Name of the external repository where Python packages are installed.
+        It's typically created by `pip.parse`.
+        This attribute is used only when `pytest_deps` is unset.
+      **kwargs: Additional named parameters to py_test.
+    """
+    shim_label = Label("//pytest:pytest_shim.py")
+
+    if pytest_deps == None:
+        pytest_deps = ["@{}//pytest".format(pip_repo)]
+        if kwargs.get("shard_count", 1) > 1:
```
mattem, line 51: The 1 seems unnecessary? Check for None, which would be falsey?

pytest/defs.bzl
```diff
@@ -0,0 +1,65 @@
...
+    py_test(
+        name = name,
+        srcs = [
+            shim_label,
+        ] + srcs,
+        main = shim_label,
+        args = [
+            "--capture=no",
+        ] + args + ["$(location :%s)" % x for x in srcs],
+        deps = deps + pytest_deps,
```
mattem, line 63: Might want to swap the ordering, so that this levels pytest dependencies will "win" over any defined transitively (perhaps there is a test helper as a `py_library` somewhere in the graph that consumes a different pytest version)

pytest/defs.bzl
```diff
@@ -0,0 +1,65 @@
...
+    shard_count = 2,
+    srcs = ["test.py"],
+    pytest_deps = [requirement("pytest"), requirement("pytest-shard"), ...],
+)
+```
+"""
+
+load("//py:defs.bzl", "py_test")
+
+def pytest_test(name, srcs, deps = [], args = [], pytest_deps = None, pip_repo = "pip", **kwargs):
```
mattem, line 31: I don't love the name here, the rest of the rules are `py_*`, representing the language, so technically this should be `py_pytest_test`, even if it is a mouthful.

pytest/pytest_shim.py
```diff
@@ -0,0 +1,71 @@
...
+            for arg in args:
+                if not arg.startswith("--"):
+                    # arg is a src file. Add test class/method selection to it.
+                    # test.py::TestClass::test_fn
+                    arg = "{arg}::{module_fn}".format(arg=arg, module_fn=test_filter.replace(".", "::"))
+                pytest_args.append(arg)
+    else:
+        pytest_args.extend(args)
+
+    print(pytest_args, file=sys.stderr)
```
mattem, line 70: This seems a little spammy, print it if there's some flag set?

pytest/pytest_shim.py
alexeagle: Can I file and leave as a follow-up? Right now I'm looking for parity with v1.1.1 https://registry.bazel.build/modules/caseyduquettesc_rules_python_pytest

pytest/defs.bzl
alexeagle: The user can explicitly say `shard_count = 1` which I hoped to account for

pytest/defs.bzl
alexeagle: done

alexeagle:
I have very little time to prepare a talk for Monday so I'm going to try the https://github.com/bazelbuild/rules_python/pull/2044 approach instead

pytest/defs.bzl
alexeagle: I don't know, I didn't write this code. Maybe Casey is willing to be the PR author.

caseyduquettesc: This absolutely can be in the shim. In the very beginning, the shim was a trivial wrapper around `pytest.main` and adding default arguments here was less code than in the shim. As more pytest features were hooked up to native Bazel features, the shim gathered more argument parsing logic and I simply missed calling this out in the pull request that first began adding default arguments in the shim. I see no reason this shouldn't be part of `pytest_args = ["--ignore=external"]` in the shim.

PR #349, opened 2024-06-16
feat: py_image_layers

alexeagle:
Fixes #212

---

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: yes
- Breaking change (forces users to change their own code or config): kind of - needs bazel_lib 2.x
- Suggested release notes appear below: yes

A helper function to create image layers for a `py_binary` is now included.

### Test plan

- New test cases added

PR #402, opened 2024-10-03
feat: add py_image_layer

thesayyn:
Replaces https://github.com/aspect-build/rules_py/pull/349

Add `py_image_layer` macro for creating py container images.

### Test plan

I will add a test in a follow-up.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,7 +8,7 @@ module(
 
 # Lower-bound versions of direct dependencies.
 # When bumping, add a comment explaining what's required from the newer release.
-bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
+bazel_dep(name = "aspect_bazel_lib", version = "2.9.1") # py_image_layer requires 2.x for the `tar` rule.
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_python", version = "0.29.0")
 bazel_dep(name = "platforms", version = "0.0.7")
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -33,10 +33,7 @@ stardoc_with_diff_test(
 
 stardoc_with_diff_test(
     name = "py_image_layer",
-    bzl_library_target = "//py:defs",
-    symbol_names = [
-        "py_image_layer",
-    ],
+    bzl_library_target = "//py/private:py_image_layer",
 )
 
 stardoc_with_diff_test(
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -32,6 +32,14 @@ stardoc_with_diff_test(
 )
 
 stardoc_with_diff_test(
+    name = "py_image_layer",
+    bzl_library_target = "//py:defs",
+    symbol_names = [
+        "py_image_layer",
+    ],
+)
+
+stardoc_with_diff_test(
     name = "venv",
     bzl_library_target = "//py/private:py_venv",
 )
--- a/py/private/py_image_layer.bzl
+++ b/py/private/py_image_layer.bzl
@@ -1,0 +1,138 @@
+"py_image_layer"
+
+load("@aspect_bazel_lib//lib:tar.bzl", "mtree_spec", "tar")
+
+default_layer_groups = {
+    # match *only* external pip like repositories that contain the string "pip_deps"
+    "packages": "\\.runfiles/pip_deps.*",
+    # match *only* external repositories that begins with the string "python"
+    # e.g. this will match
+    #   `/hello_world/hello_world_bin.runfiles/rules_python~0.21.0~python~python3_9_aarch64-unknown-linux-gnu/bin/python3`
+    # but not match
+    #   `/hello_world/hello_world_bin.runfiles/_main/python_app`
+    "interpreter": "\\.runfiles/python.*-.*/",
+}
+
+def py_image_layer(name, py_binary, root = None, layer_groups = {}, compress = "gzip", tar_args = ["--options", "gzip:!timestamp"], **kwargs):
+    """Produce a separate tar output for each layer of a python app
+
+    > Requires `awk` to be installed on the host machiner/rbe runner.
+
+    For better performance, it is recommended to split the output of a py_binary into multiple layers.
+    This can be done by grouping files into layers based on their path by using the `layer_groups` attribute.
+
+    The matching order for layer groups is as follows:
+        1. `layer_groups` are checked first.
+        2. If no match is found for `layer_groups`, the `default layer groups` are checked.
+        3. Any remaining files are placed into the default layer.
+    
+    The default layer groups are:
+    ```
+    {
+        "packages": "\\.runfiles/pip_deps.*", # contains third-party deps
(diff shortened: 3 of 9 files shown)


MODULE.bazel
```diff
@@ -8,7 +8,7 @@ module(
 
 # Lower-bound versions of direct dependencies.
 # When bumping, add a comment explaining what's required from the newer release.
-bazel_dep(name = "aspect_bazel_lib", version = "1.40.0")
+bazel_dep(name = "aspect_bazel_lib", version = "2.9.1") # py_image_layer requires 2.x for the `tar` rule.
```
alexeagle, line 11: this is an important change before 1.0 actually. Would be technically breaking to increase minimum later

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,138 @@
...
+    mtree_spec(
+        name = name + ".manifest",
+        srcs = [py_binary],
+        **kwargs
+    )
+
+    groups = dict(**layer_groups)
+    group_names = groups.keys() + ["default"]
+
+    mtree_begin_blocks = "\n".join([
```
alexeagle, line 79: nit: extract function `generate_awk_program` or sth

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,138 @@
...
+    #   `/hello_world/hello_world_bin.runfiles/rules_python~0.21.0~python~python3_9_aarch64-unknown-linux-gnu/bin/python3`
+    # but not match
+    #   `/hello_world/hello_world_bin.runfiles/_main/python_app`
+    "interpreter": "\\.runfiles/python.*-.*/",
+}
+
+def py_image_layer(name, py_binary, root = None, layer_groups = {}, compress = "gzip", tar_args = ["--options", "gzip:!timestamp"], **kwargs):
+    """Produce a separate tar output for each layer of a python app
+
+    > Requires `awk` to be installed on the host machiner/rbe runner.
```
alexeagle, line 19: nit: spelling "machine"

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,138 @@
+"py_image_layer"
+
+load("@aspect_bazel_lib//lib:tar.bzl", "mtree_spec", "tar")
+
+default_layer_groups = {
+    # match *only* external pip like repositories that contain the string "pip_deps"
+    "packages": "\\.runfiles/pip_deps.*",
```
alexeagle, line 7: what if the `pip_install` was named something else? Even in this repo we use `pypi` as the name rather than `pip` - so if you had an example you'd notice.

Maybe it should be a placeholder that can be separately filled with an attribute like `deps_repo_name`

alexeagle (commented):
thanks!

py/private/py_image_layer.bzl
thesayyn: hmm, i should probably capture, `site-packages/`, similar to what i did for `.runfiles/python`

MODULE.bazel
thesayyn: yes, we should have made this change nevertheless.

py/private/py_image_layer.bzl
thesayyn: okay i replaced it with `\\.runfiles/.*/site-packages` which will match regardless of repo name.

py/private/py_image_layer.bzl
thesayyn: agreed, done

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,138 @@
+"py_image_layer"
```
alexeagle, line 1: make the module docstring longer

Also maybe claim that it's unstable, not semver, for 1.0.0

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,138 @@
...
+
+    The matching order for layer groups is as follows:
+        1. `layer_groups` are checked first.
+        2. If no match is found for `layer_groups`, the `default layer groups` are checked.
+        3. Any remaining files are placed into the default layer.
+    
+    The default layer groups are:
+    ```
+    {
+        "packages": "\\.runfiles/pip_deps.*", # contains third-party deps
```
alexeagle, line 32: oops this got out of date already.

Is it possible to make them reused somehow?

alexeagle (approved):

py/private/py_image_layer.bzl
thesayyn: docstring seem to be ignored, because of how we do docs here. I just put a big letter experimental in the beginning of the docs.

alexeagle: That means the stardoc_with_diff_test points at the wrong file actually.

docs/BUILD.bazel
```diff
@@ -31,6 +31,14 @@ stardoc_with_diff_test(
     bzl_library_target = "//py/private:py_pex_binary",
 )
 
+stardoc_with_diff_test(
+    name = "py_image_layer",
+    bzl_library_target = "//py:defs",
```
alexeagle, line 36: this should point to the py_image_layer.bzl so you get the module docstring too

py/private/py_image_layer.bzl
thesayyn: fixed

py/private/py_image_layer.bzl
thesayyn: not really... it's annoying...

thesayyn: fixed it

docs/BUILD.bazel
thesayyn: yup, just fixed it

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,156 @@
+"""py_image_layer macro for creating multiple layers from a py_binary
+
+>> [!WARNING]
```
alexeagle, line 3: double `>` intentional?

thesayyn: nope, i was trying to get stardoc leave it alone, it seem to escaping it, aware of any workarounds?

alexeagle: hm that is very interesting, I haven't had that problem before. WHatever it's just cosmetic and inconsequential, ship it :)

py/private/py_image_layer.bzl
```diff
@@ -0,0 +1,156 @@
...
+    # Produce the manifest for a tar file of our py_binary, but don't tar it up yet, so we can split
+    # into fine-grained layers for better pull, push and remote cache performance.
+    mtree_spec(
+        name = name + ".manifest",
+        srcs = [py_binary],
+        **kwargs
+    )
+
+    groups = dict(**layer_groups)
+    group_names = groups.keys() + ["default"]
```
hartikainen, line 138: I hope I'm not adding any unintentional noise here, especially since I'm still very new to Bazel. I wonder if these two lines should be something like:
```
    groups = dict(default_layer_groups.items() + layer_groups.items())
    group_names = default_layer_groups.keys() + groups.keys()
```

PR #412, opened 2024-10-16
feat: drop rattler_installs_packages for uv crates

### LINKED ISSUE #344: [FR]: include (de)activation scripts in venvs
### What is the current behavior?

I've a repository with multiple python distribution packages. I have this rule at the root of my repo meant to create a venv for use by my IDE:
```bazel
py_venv(
    name = "venv",
    deps = [ 
        ...
    ]
)
```
The venv produced by `rules_py` does not include (de)activation scripts. I'm using version 0.7.3 of the rules.
 

### Describe the feature

Add activate/deactivate scripts in the venv as per regular venv creation.

### LINKED ISSUE #345: [FR]: Switch to uv for venv creation and for pip compile
### What is the current behavior?

* As of [#233](https://github.com/aspect-build/rules_py/pull/233), to create virtualenv's, aspect_rules_py has maintained its own Rust implementation (that predates uv by only a [couple months](https://astral.sh/blog/uv) -- what timing!).

  aspect_rules_py's `venv` binary does not work on older Linux systems such as RHEL7, resulting in an error like `venv: /lib64/libc.so.6: version 'GLIBC_2.18' not found` at runtime.  Unfortunately this blocks adoption of aspect_rules_py by some organizations that are still stuck on older Linux systems (such as my own).

* For pip compile functionality, aspect_rules_py currently does not provide a solution, leaving this up to the user to handle with some other tool.[1]

### Describe the feature

Switching to `uv` (perhaps even via [rules_uv](https://github.com/theoremlp/rules_uv)) could provide the following benefits:

* Compatibility: `uv` (and likewise [rules_uv](https://github.com/theoremlp/rules_uv)) works on older Linux systems.

* Reduced maintenance costs associated with maintaining aspect_rules_py's own implementation of `venv`.

* Given how popular `uv` has already become, `uv` represents a more "standard" and more field-tested implementation.

* If `uv` is adopted for venv creation, using it to also provide `pip_compile` functionality would make aspect_rules_py even more useful.

Subsumes #300.

---------------

[1] A previous version of this description incorrectly assumed from #300 that aspect_rules_py provides this functionality directly by depending on pip-tools.

mattem:
Drops the dependency on `rattler_installs_packages` in favour of the uv crates.

Note that while this works, the creates of uv do not officially support depending directly on the crates, however there is little overhead here, and is pinned to a fixed SHA.

Closes #307
Closes #344
Closes #345

Due to compilation issues under llvm, this diff also switches to musl for the toolchain binaries. We need to keep llvm for proc macros, unfortunately. 

---

Drop dependency on `rattler_installs_packages` in favour of the uv crates.
Produce statically linked toolchain binaries via musl.

### Test plan

- New test cases added

### DIFF
--- a/py/private/py_binary.bzl
+++ b/py/private/py_binary.bzl
@@ -73,11 +73,6 @@ def _py_binary_rule_impl(ctx):
             "{{ARG_COLLISION_STRATEGY}}": ctx.attr.package_collisions,
             "{{ARG_PYTHON}}": to_rlocation_path(ctx, py_toolchain.python) if py_toolchain.runfiles_interpreter else py_toolchain.python.path,
             "{{ARG_VENV_NAME}}": ".{}.venv".format(ctx.attr.name),
-            "{{ARG_VENV_PYTHON_VERSION}}": "{}.{}.{}".format(
-                py_toolchain.interpreter_version_info.major,
-                py_toolchain.interpreter_version_info.minor,
-                py_toolchain.interpreter_version_info.micro,
-            ),
             "{{ARG_PTH_FILE}}": to_rlocation_path(ctx, site_packages_pth_file),
             "{{ENTRYPOINT}}": to_rlocation_path(ctx, ctx.file.main),
             "{{PYTHON_ENV}}": "\n".join(_dict_to_exports(env)).strip(),
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,14 +1,14 @@
 {
-  "checksum": "d106312d2f6264156c52e22eaf1d544f39fa5380fa99cb923f75c2320aaafd25",
+  "checksum": "81a92b86099c7c9b831ce831b33ad24c552a0c97ef3d8dc37836db3d08529ec4",
   "crates": {
-    "addr2line 0.21.0": {
+    "addr2line 0.24.2": {
       "name": "addr2line",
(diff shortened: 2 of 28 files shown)


py/private/py_binary.bzl
```diff
@@ -73,11 +73,6 @@ def _py_binary_rule_impl(ctx):
             "{{ARG_COLLISION_STRATEGY}}": ctx.attr.package_collisions,
             "{{ARG_PYTHON}}": to_rlocation_path(ctx, py_toolchain.python) if py_toolchain.runfiles_interpreter else py_toolchain.python.path,
             "{{ARG_VENV_NAME}}": ".{}.venv".format(ctx.attr.name),
-            "{{ARG_VENV_PYTHON_VERSION}}": "{}.{}.{}".format(
```
alexeagle, line 76: I don't remember why this was needed before. Is it desirable that we no longer need to tell it the python version? Or is this undesirable and now something can drift?

alexeagle (approved):

py/private/py_binary.bzl
mattem: The UV crate query Python before creating the venv for version and a bunch of other info, and as the venv is created at runtime (JIT style), we no longer need to pass this, nothing will drift.
It's still needed for unpacking a wheel however, as we don't want to run the interpreter as part of the action.

PR #415, opened 2024-10-22
Use uv

illicitonion:


mattem:
Included changes for musl in #412

PR #432, opened 2024-11-12
fix: flip unused_inputs=1 by default

thesayyn:

### Test plan

- New test cases added


thesayyn:
Needs: https://github.com/bazelbuild/bazel-central-registry/pull/3167 first

PR #441, opened 2024-11-19
Add backwards compatible types to file generated by __test__.py

hofbi:
As suggested in https://github.com/aspect-build/rules_py/pull/435#issuecomment-2482658330 a backwards compatible (< Python 3.9) type hint.

Regardless of merging this PR or not, the current 1.0.0 release is broken for < Python 3.9. Since Python 3.8 is already EOL, this is in theory valid, but I guess there are still a few people using Python 3.8 because of Ubuntu 20.04.

---

### Test plan

- New test cases added

### DIFF
--- a/py/private/pytest.py.tmpl
+++ b/py/private/pytest.py.tmpl
@@ -14,6 +14,7 @@
 
 import sys
 import os
+from typing import List
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


hofbi:
When looking for test cases, I found https://github.com/aspect-build/rules_py/blob/97aff2461348c841aeb13d3e2cea0bcb27c26845/examples/multi_version/BUILD.bazel#L26 which seems to test with Python 3.8, hence I would expect that this test should have failed.

mattem:
Thanks 🙌  
Yeah the tests never ran on #428 as the author hadn't submitted a PR to rules_py before, and I missed clicking the "Approve Workflow Run" to run the tests, and it seems the settings on this repo allows merging with no CI results. If this is green I can merge this and tag a v1.0.1.

mattem (approved):

Issue #436, opened 2024-11-14
[FR]: Add ability to pass additional user defined flags to interpreter


muravev-vasilii:
### What is the current behavior?

Currently bootstrap script doesn't allow to pass custom interpreter flags to Python interpreter:
https://github.com/aspect-build/rules_py/blob/main/py/private/run.tmpl.sh#L61

Two flags are added by rules_py implementation (https://github.com/aspect-build/rules_py/blob/main/py/private/py_semantics.bzl#L6-L14), but I didn't find an option to add more flags defined by user to interpreter

**args** argument for py_binary target allows to pass custom arguments to invoked script, but not to interpreter itself.

### Describe the feature

I would like to have ability to pass custom user flags to Python interpreter used by rules_py.

Main motivation for it to allow to run py_binary targets under debugpy debugger via bazel.
Instead of having following line in bootstrap file

`exec "{{EXEC_PYTHON_BIN}}" {{INTERPRETER_FLAGS}} "$(rlocation {{ENTRYPOINT}})" "$@"`

I want to have something like
`+exec "{{EXEC_PYTHON_BIN}}" {{INTERPRETER_FLAGS}} $RULES_PY_USER_INTERPRETER_FLAGS "$(rlocation {{ENTRYPOINT}})" "$@"` 

Example usage will be:
`RULES_PY_USER_INTERPRETER_FLAGS="-m debugpy --listen 12345 --wait-for-client" bazel run //path/to/rules_py/py_binary/target`

I've seen and tried proposed way to debug via creating venv from https://github.com/bazelbuild/rules_python/issues/1401#issuecomment-2389768034

This way doesn't work for us in our monorepo setting for couple of reasons:

- We use custom pytest_runner
- We have our own VSCode extension which uses bazel to run tests and debug in TestExplorer and other things and we would prefer to keep interaction with tests and codebase through bazel, not via python interpreter from rules_py venv
- It creates one extra step for our user to be able to interact with IDE, comparing with current setup

Similar feature request from bazel slack:
https://bazelbuild.slack.com/archives/CA31HN1T3/p1730573900089359

Another alternative I was thinking is to use run_under flag (https://bazel.build/reference/command-line-reference#flag--run_under), but it seems to be not feasible for debugpy use case, as bootstrap script is shell and not Python.

If there is a way to specify custom toolchain for this use case I am happy to learn how to do it as well!

Thanks!

muravev-vasilii:
[...]

PR #442, opened 2024-11-20
feat: add ability to pass user flags to python interpreter

muravev-vasilii:
This change is meant to demonstrate the proposed update for the following feature request: https://github.com/aspect-build/rules_py/issues/436

---

Users can influence behaviour of interpreter by setting flags via setting env variable RULES_PY_USER_INTERPRETER_FLAGS

### Test plan

- Manual testing;

Ran bazel target with

`RULES_PY_USER_INTERPRETER_FLAGS="-m debugpy --listen 12345 --wait-for-client" bazel run //path/to/rules_py/py_binary/target`

From VSCode side:

- installed Python Debugger extension
- updated launch.json with
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Python Debugger: Attach",
            "type": "debugpy",
            "request": "attach",
            "connect": {
              "host": "localhost",
              "port": 12345
            }
          }
    ]
}
```

Used Run&Debug tab in VSCode to set breakpoints, connect to program and debug


mnoah1:
Also, one other change for the debug use case is that the run template should set the `PYDEVD_RESOLVE_SYMLINKS` environment variable to true before starting debugpy.  This resolves symlinks when sending paths to the client.  Without this set, VS Code will open any paths within the Bazel cache as their own tab instead of showing the breakpoint in the original source file.   More info:  https://github.com/microsoft/debugpy/issues/743

py/private/py_binary.bzl
```diff
@@ -149,6 +149,10 @@ A collision can occour when multiple packages providing the same file are instal
         default = "error",
         values = ["error", "warning", "ignore"],
     ),
+    "interpreter_args": attr.string_list(
```
alexeagle, line 152: nit: from https://docs.python.org/3/using/cmdline.html#command-line it seems like it should be named `interpreter_options` rather than args

py/private/py_binary.bzl
```diff
@@ -149,6 +149,10 @@ A collision can occour when multiple packages providing the same file are instal
...
+        doc = "Additional arguments to pass to the Python interpreter.",
```
alexeagle, line 153: add what options rules_py always adds (`-I`, `-B` or whatever)

alexeagle (commented):
optional: maybe we should document a bit more about the `--compilation_mode dbg` use case you'll use this for

py/private/py_binary.bzl
muravev-vasilii: > nit: from https://docs.python.org/3/using/cmdline.html#command-line it seems like it should be named `interpreter_options` rather than args

agree, updated

py/private/py_binary.bzl
muravev-vasilii: > add what options rules_py always adds (`-I`, `-B` or whatever)

updated doc

muravev-vasilii:
> optional: maybe we should document a bit more about the `--compilation_mode dbg` use case you'll use this for

This one I would prefer to add in a separate example, as I need to create minimalistic example not applied to our monorepo first

alexeagle (approved):

PR #443, opened 2024-11-20
chore(docs): document virtual_deps and resolutions

alexeagle:
Fixes a FIXME

---

### Test plan


alexeagle:
I no longer have push access to this branch :(

PR #444, opened 2024-11-20
Revert "example: Add example for pybind11"

alexeagle:
Reverts aspect-build/rules_py#375

Broke the `main` branch.

As noted earlier we don't have the right branch protections on this repo to prevent merging red PRs. @mattem please fix it or grant me permissions so I can fix it.

PR #446, opened 2024-11-20
chore(rollforward): example: Add example for pybind11

alexeagle:
Reverts aspect-build/rules_py#444


alexeagle:
@hofbi FYI - this broke because we also test the ruleset with WORKSPACE setup.

hofbi:
Thanks for letting me know, I created https://github.com/aspect-build/rules_py/pull/447 including pybind in the workspace approach. Just need to fix a new error there.

PR #447, opened 2024-11-20
chore(rollforward): example: Add example for pybind11

hofbi:
Reverts #444 and adds pybind dependency to `WORKSPACE` file as mention in #446.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -85,3 +85,5 @@ crate.from_cargo(
 )
 
 use_repo(crate, "crate_index")
+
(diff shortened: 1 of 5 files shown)


hofbi:
The new error looks like the C++ std lib is not correctly linked. Looking at your toolchain, https://github.com/aspect-build/rules_py/blob/main/third_party/com.github/bazel-contrib/toolchains_llvm/clang_ldd.patch looks like it could be related. However, I don't know why this patch is applied. Removing it led to other errors, so maybe @mattem could you explain what was the reason for this?

PR #448, opened 2024-11-22
docs: document virtual_deps and resolutions

alexeagle:
Fixes a docs FIXME.

Also show how a dependency can be dropped altogether using an empty filegroup, as requested by a client.

---

### Test plan


alexeagle:
FYI @mattem (I can't add reviewers)

mattem (approved):

PR #451, opened 2024-11-25
Canonicalize target link in venv

tgeng:
This way the created venv can be stable across sandboxes.

Internally we implement a test daemon that reuses the same venv (and python process) across tests whenever possible. Hence, if the target link is canonicalized, the same venv can be used across different targets.

---

### Test plan


tgeng:
Actually we found a way to workaround this. So feel free to just close it unless you think this would be a useful thing by itself.

PR #452, opened 2024-11-29
docs/example-pytest-fixtures

njlr:
Adds an example of test fixtures and `chdir`.

This option was quite hard to discover, however test fixtures are common.


alexeagle:
Cool thanks

alexeagle (approved):

Issue #457, opened 2024-12-04
[FR]: py_test doesn't support `env_inherit`


keith:
### Describe the feature

rules_python's `py_test` supports the `env_inherit` attribute which allows you to specify a list of environment variables which, if present, will be passed through to the test run. Currently rules_py's py_test doesn't support this attribute.

Repro case: https://github.com/aspect-build/rules_py/pull/458

PR #459, opened 2024-12-04
Add env_inherit to py_test

keith:
Fixes https://github.com/aspect-build/rules_py/issues/457

### DIFF
--- a/py/private/py_binary.bzl
+++ b/py/private/py_binary.bzl
@@ -1,8 +1,8 @@
 """Implementation for the py_binary and py_test rules."""
 
-load("@rules_python//python:defs.bzl", "PyInfo")
-load("@aspect_bazel_lib//lib:paths.bzl", "BASH_RLOCATION_FUNCTION", "to_rlocation_path")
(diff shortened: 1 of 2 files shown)


py/private/py_binary.bzl
```diff
@@ -205,7 +214,7 @@ py_binary = rule(
 py_test = rule(
     doc = "Run a Python program under Bazel. Most users should use the [py_test macro](#py_test) instead of loading this directly.",
     implementation = py_base.implementation,
-    attrs = py_base.attrs,
+    attrs = py_base.attrs | py_base.test_attrs,
```
alexeagle, line 217: huh, I didn't know about this merge operator on structs, neat

alexeagle: oh actually the spec says 

> - x | y means either x or y.

Should this be merging them instead?

keith: This is for merging dictionaries, the right takes precedence but the result is otherwise the union

PR #464, opened 2024-12-05
Don't bake env into binary

keith:
This mirrors the behavior of rules_python's py_binary, and other
*_binary rules. Previously any env set on the binary was baked into the
script, and therefore included if the script was run with --run_under,
or through a genrule. Now this env is passed through to
RunEnvironmentInfo so it only applies in the default cases.


alexeagle (approved):
Nice thanks, I had forgotten about RunEnvironmentInfo :)

PR #480, opened 2024-12-13
chore: Autoformat with buildifier

hofbi:
Follow up of #474 running buildifier on all files. A follow up PR will fix the remaining `buildifier-lint` warnings and add pre-commit to CI similar to https://github.com/bazel-contrib/rules-template/pull/130.


WORKSPACE
```diff
@@ -115,10 +115,11 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+
 ############################################
 # Gazelle, for generating bzl_library targets
 load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
```
hofbi, line 123: This change is the only one where I am not sure if the order has an impact

alexeagle (approved):

PR #486, opened 2024-12-19
pytest - add suite name to junit.xml report

vinnybod:
The `junit.xml` files generated by pytest use the name `pytest` by default. For any tools that consume the xml files like IDEs or CI system UIs, this makes the output unfriendly because every test case rolls up the the same test suite named `pytest`.

The changes here add the `BAZEL_TARGET` name to the suite using standard pytest command line args.

Here is an example before and after when running the tests with and without the change in IntelliJ. The example only runs two test targets. The issue becomes more apparent as the number of targets scales up.

Screenshot 2024-12-18 at 5 13 47 PM

Screenshot 2024-12-18 at 5 12 37 PM

---

### Test plan

Run a `py_test` target that uses `py_pytest_main` and observe the test suite name.

### DIFF
--- a/py/private/pytest.py.tmpl
+++ b/py/private/pytest.py.tmpl
@@ -23,6 +23,7 @@ if __name__ == "__main__":
     $$CHDIR$$
 
     os.environ["ENV"] = "testing"
+    suite_name = os.environ.get("BAZEL_TARGET")
 
     args = [
         "--verbose",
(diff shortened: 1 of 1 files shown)


fzakaria:
This has been a great QoL improvement for us; highly recommend the merge.

py/private/pytest.py.tmpl
```diff
@@ -23,6 +23,7 @@ if __name__ == "__main__":
     $$CHDIR$$
 
     os.environ["ENV"] = "testing"
+    suite_name = os.environ.get("BAZEL_TARGET")
```
mattem, line 26: bit, move this closer to where it's used.

py/private/pytest.py.tmpl
```diff
@@ -32,6 +33,9 @@ if __name__ == "__main__":
         "no:cacheprovider",
     ]
 
+    if suite_name:
```
mattem, line 36: Can this be moved within the conditional that sets `junit_xml`, as without that set this flag is moot?

mattem (approved):

Issue #466, opened 2024-12-05
[Bug]: py_test does not accept the same resolutions as py_binary


njlr:
### What happened?

Trying to use virtual_deps works with py_binary but not with py_test.

It complains that the type of `resolutions` is wrong.

### Version

Development (host) and target OS/architectures: Linux

Output of `bazel --version`:

```
bazel 6.2.0
```

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

```starlark
http_archive(
  name = "aspect_rules_py",
  sha256 = "2ce48e0f3eaaf73204b623f99f23d45690b862a994b5b3c2464a2e361b0fc4ae",
  strip_prefix = "rules_py-1.0.0",
  url = "https://github.com/aspect-build/rules_py/releases/download/v1.0.0/rules_py-v1.0.0.tar.gz",
)
```

Language(s) and/or frameworks involved:
Python

### How to reproduce

```shell
Create a py_test that uses resolutions like so:


# ...

resolutions = resolutions.from_requirements(all_whl_requirements_by_package, requirement)

py_test(
  name = "py_test_tests",
  srcs = glob([
    "tests/**/*.py",
  ]) + [
    ":__test__.py"
  ],
  main = ":__test__.py",
  package_collisions = "error",
  resolutions = resolutions,
  python_version = "3.9.16",
  size = "small",
)
```

njlr:
As a work-around:

```starlark
r = resolutions.from_requirements({
  "boto3": None,
  "torch": None,
  # etc...
}, requirement).to_label_keyed_dict()

py_test(
  name = "py_test_tests",
  srcs = glob([
    "tests/**/*.py",
  ]) + [
    ":__test__.py"
  ],
  main = ":__test__.py",
  package_collisions = "error",
  resolutions = r,
  python_version = "3.9.16",
  size = "small",
)
```

PR #487, opened 2024-12-19
fix: make py_binary#resolutions and py_test#resolutions identical. 

njlr:
Closes https://github.com/aspect-build/rules_py/issues/466

 - Make resolutions consistent between `py_binary` and `py_test`
 - Allow a list to be passed to `from_requirements`
 - Add an example of `virtual_deps` feature

---

> - Fix: Make resolutions consistent between `py_binary` and `py_test`
> - Feature: Allow a list to be passed to `from_requirements`
> - Docs:  Add an example of `virtual_deps` feature

### Test plan

- Manual testing; please provide instructions so we can reproduce: Added an example

### DIFF
--- a/docs/py_binary.md
+++ b/docs/py_binary.md
@@ -16,12 +16,12 @@ load("@rules_python//python:repositories.bzl", "py_repositories", "python_regist
 python_register_toolchains(
     name = "python_toolchain_3_8",
     python_version = "3.8.12",
-    # setting set_python_version_constraint makes it so that only matches py_* rule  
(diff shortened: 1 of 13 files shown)


thesayyn (approved):
Aha, great finding.  It all looks good to me, while technically breaking change, we claimed that py_binary and py_test are identical, so this counts as a bug.

thesayyn:
@njlr can you force push to kick the CI?

thesayyn:
Thank you.

PR #491, opened 2025-01-17
docs: fix example of py_image_layer

alexeagle:
It's named `binary` rather than `py_binary`


alexeagle:
I'm not sure the breaking change was intentional? https://github.com/aspect-build/rules_py/commit/b9a2c70a56d80b767102660644b7ce3e69d42e7f

It's breaking `aspect init` template as well.

thesayyn (approved):
ouch

PR #493, opened 2025-01-21
pytest test sharding

vinnybod:
Adds test sharding support to the pytest runner.

I took some inspiration from https://github.com/caseyduquettesc/rules_python_pytest . In particular, I used it as a reference for which Bazel environment variables need to be pulled in the shim.

The pytest plugin is a derivative work of https://github.com/AdamGleave/pytest-shard . That dependency hasn't been updated in a long time, and the sharding was based off hashing which often results in unbalanced shards or errors from empty shards. I vendored the code into the repo, modified it to use a plain round robin strategy for selection, and converted the code to be class based so it can be used with `pytest.main()`.

### Test plan

I added an example to the examples directory and I tested it with my own repository here: https://github.com/vinnybod/bazel-examples/blob/python-test-sharding/test-shard-python/BUILD.bazel


py/private/pytest_shard/pytest_shard.py
```diff
@@ -0,0 +1,73 @@
...
+    @staticmethod
+    def pytest_report_collectionfinish(config, items: Sequence[nodes.Node]) -> str:
+        """Log how many and, if verbose, which items are tested in this shard."""
+        msg = f"Running {len(items)} items in this shard"
+        if config.option.verbose > 0 and config.getoption("num_shards") > 1:
+            msg += ": " + ", ".join([item.nodeid for item in items])
+        return msg
+
+    @staticmethod
+    def pytest_collection_modifyitems(config, items: List[nodes.Node]):
```
fzakaria, line 64: What if it's a class with methods? Do those become 1 item or multiple?
Is this list also stable? Would be good to confirm since it's round robin.

If it's not, we can enforce a sort order.

vinnybod: I think this is covered. This hook is invoked by pytest after all tests have been collected, and it knows how to collect `unittest` classes. I can add another example file to prove this.

https://docs.pytest.org/en/stable/reference/reference.html#std-hook-pytest_collection_modifyitems

alexeagle:
I haven't had time to study this yet, note I had https://github.com/aspect-build/rules_py/pull/401 which may or may not overlap.

vinnybod:
I did see #401 , but there hadn't been activity on it in a while. I think this could still be merged in the meantime.

If you do want to commit it with #401 instead, I think it would still be worth while to vendor in the pytest-shard plugin with the round-robin patch, always loading it, since its a better UX for developers.

alexeagle:
round-robin is not the best strategy, because it tends to mean a shard has to do many different setup routines.

Take as example (invented from what I remember of JUnit, but I think it applies)

file 1
  test class A
    setup A
    test case i
    test case ii
  test class B
    setup B
    test case iii
    test case iv
file 2
  test class C
    setup C
    test case v
    test case vi

If this is meant to have --shard_count=2, you'd rather have test cases [i, ii, iii] and [iv, v, vi] so that the first shard can skip setup C and the second shard can skip setup A. 

IIUC, round robin means you'll have both shards having to run all three setups.

We did test sharding in Jasmine, https://github.com/aspect-build/rules_jasmine/blob/main/jasmine/private/runner.cjs#L32-L53 where we sort and then partition them

py/private/py_pytest_main.bzl
```diff
@@ -88,6 +88,6 @@ def py_pytest_main(name, py_library = default_py_library, deps = [], data = [],
         srcs = [test_main],
         tags = tags,
         visibility = visibility,
-        deps = deps,
+        deps = deps + ["@aspect_rules_py//py/private/pytest_shard"],
```
alexeagle, line 91: this is technically incorrect in a macro, where a string will turn into a label in the caller's context and they might not use this name for the apparent repo.

Instead you should `Label("//py/private/pytest_shard")` so you explicitly turn into a label in the macro's context where it will be "the repository I'm in"

py/private/pytest_shard/LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright 2019 Adam Gleave
```
alexeagle, line 1: also add a README pointing to the precise location you vendored these files from, so we can later reason about the delta from upstream changes

py/private/pytest_shard/pytest_shard.py
```diff
@@ -0,0 +1,73 @@
...
+https://github.com/AdamGleave/pytest-shard
+The main difference is that the plugin has been converted to a class
+so that it can be passed to pytest.main's plugins attribute, and the
+shard selection was changed from a hash-based approach to a round-robin
+approach for a more balanced distribution with smaller numbers of tests.
+"""
+
+from typing import Iterable, List, Sequence
+
+from _pytest import nodes  # for type checking only
```
alexeagle, line 12: Is it worth including type hints in this repo if we never run any type-checking? I'm genuinely not sure, maybe it's wise just for documentation, or in case we add type-checking later.

alexeagle (commented):
Nice, thanks for bringing this in

py/private/pytest_shard/pytest_shard.py
vinnybod: it is nice to have for IDEs

py/private/py_pytest_main.bzl
vinnybod: TIL! Thanks. I originally just had `//py/private/pytest_shard` which didn't obviously work when used from an external repository.

vinnybod:
@alexeagle I think that is less of a problem when you only have one test file per `py_test` target. But I suppose if we want to make this more optimal for use cases where developers are globbing a bunch of `*_test.py` into a single target, the change could help.

If I'm understanding your explanation correctly, instead of a round robin, we should just take the full list [i, ii, iii, iv, v, vi] and partition it by the number of shards, keeping the default ordering (n=3, [i, ii], [iii, iv], [v, vi]). Does that sound right?

I did look at [rules_jvm's implementation](https://github.com/bazel-contrib/rules_jvm/blob/main/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSharding.java#L39) and they use hashes, so they don't take into account any ordering.

alexeagle:
Yes that's exactly right. Obviously this makes a difference for tests with heavy fixtures, with a larger number of test cases and a larger shard count the more we can "group by fixture" the less duplicate work is performed. I think rules_jvm probably has this wrong too, but I'm pretty sure the google3 implementation gets it right.

py/private/py_pytest_main.bzl
fzakaria: also TIL

py/private/pytest_shard/pytest_shard.py
```diff
@@ -0,0 +1,73 @@
...
+            type=positive_int,
+            default=1,
+            help="Total number of shards.",
+        )
+
+    @staticmethod
+    def pytest_report_collectionfinish(config, items: Sequence[nodes.Node]) -> str:
+        """Log how many and, if verbose, which items are tested in this shard."""
+        msg = f"Running {len(items)} items in this shard"
+        if config.option.verbose > 0 and config.getoption("num_shards") > 1:
```
alexeagle, line 59: how does one set this verbose option?

py/private/pytest.py.tmpl
```diff
@@ -52,7 +65,7 @@ if __name__ == "__main__":
     if len(cli_args) > 0:
         args.extend(cli_args)
 
-    exit_code = pytest.main(args)
+    exit_code = pytest.main(args, plugins=[ShardPlugin()])
```
alexeagle, line 68: maybe we should avoid installing the plugin if sharding wasn't enabled (or the shard count is 1) ?
Helps to reduce the blast radius in case there's a problem with the plugin, it's easy for users to avoid going in that codepath

py/private/pytest_shard/pytest_shard.py
vinnybod: It looks like we already do https://github.com/aspect-build/rules_py/pull/493/files#diff-eab700c56364dc4619c882e7e89bf26cd4cd1cd9efe61284e3bd0607bfe152bcR31

alexeagle:
> The discussion about sharding strategy seems like an implementation detail that can be further refined

Sure, it's always possible to leave a TODO - but you have to appreciate that typically on OSS projects you have a contributor who's on the hook to land the change, and later is likely to disappear. As the maintainer you then end up with that tech debt on your own head. So we're usually a bit defensive about making it right at the beginning. I could trust you guys to come back to this optimization later though.

vinnybod:
Looks like a new change in https://github.com/aspect-build/rules_py/tree/main/examples/virtual_deps broke these changes. Investigating.

alexeagle (approved):

PR #500, opened 2025-01-23
ci: Add pre-commit GitHub action

hofbi:
Follow up of #480 adding a pre-commit GH action, fixing last formatting, and remaining buildifier warnings (split in 3 commit)

---


.github/workflows/ci.yaml
```diff
@@ -67,7 +73,7 @@ jobs:
   # For branch protection settings, this job provides a "stable" name that can be used to gate PR merges
   # on "all matrix jobs were successful".
   conclusion:
-    needs: test
+    needs: [test, pre-commit]
```
hofbi, line 76: @alexeagle While adding this here, I noticed that `verify-bcr-patches` and `test-release` are not listed here as deps. Is this expected?

alexeagle: No, that's a (minor) bug.

alexeagle (approved):
thanks :)

PR #501, opened 2025-01-23
ci: Add all jobs as conclusion deps

hofbi:
Follow up from https://github.com/aspect-build/rules_py/pull/500#discussion_r1926686289


.github/workflows/ci.yaml
```diff
@@ -73,7 +73,11 @@ jobs:
   # For branch protection settings, this job provides a "stable" name that can be used to gate PR merges
   # on "all matrix jobs were successful".
   conclusion:
-    needs: [test, pre-commit]
+    needs:
+      - test
+      - verify-bcr-patches
+      - test-release
+      - pre-commit
```
alexeagle, line 80: this isn't sufficient - this job will wait for them to finish, but won't consider their pass/fail in the aggregate, and the conclusion can therefore pass even if one of them fails.

hofbi: This is very suprising. I made a test with https://github.com/aspect-build/rules_py/pull/501/commits/ed1036fb86f45d9c42cad8acf42b40184ef40816 where the pre-commit step fails and the conclusion fails as well.

alexeagle: Oh, okay I'm wrong then! I guess it's only matrix jobs where we needed to call a helper in this job to get the right behavior. Thanks for testing.

alexeagle (approved):

Issue #506, opened 2025-01-24
[Bug]: aspect_rules_py 1.2.0 cannot be downloaded from BCR


bu3:
### What happened?

I tried to use aspect_rules_py `1.2.0` from BCR a few hours after it was published and I see the following error caused by the `rust_dev_dep.patch` patch mentioned in this [commit](https://github.com/bazelbuild/bazel-central-registry/commit/3e1a25afa015d29b20ecd8403579a17c4c65f889) 

The error I get is:
```
INFO: Repository aspect_rules_py+ instantiated at:
  <builtin>: in <toplevel>
Repository rule http_archive defined at:
  /private/var/tmp/_bazel_fmangione/fa14ff5f5627de19544fde8f05b3187d/external/bazel_tools/tools/build_defs/repo/http.bzl:392:31: in <toplevel>
INFO: repository @@aspect_rules_py+' used the following cache hits instead of downloading the corresponding file.
 * Hash '63bd7db0d2d4b17b107c96025423d0c882f6d72f73f9ab7a54899e122d7f746d' for https://bcr.bazel.build/modules/aspect_rules_py/1.2.0/patches/rust_dev_dep.patch
 * Hash '0ca24365412239cc8242664f6c0efd210ede8bc7b8d2a306036ccaf5ba09d81e' for https://github.com/aspect-build/rules_py/releases/download/v1.2.0/rules_py-v1.2.0.tar.gz
If the definition of 'repository @@aspect_rules_py+' was updated, verify that the hashes were also updated.
ERROR: /private/var/tmp/_bazel_fmangione/fa14ff5f5627de19544fde8f05b3187d/external/bazel_tools/tools/build_defs/repo/utils.bzl:196:18: An error occurred during the fetch of repository 'aspect_rules_py+':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_fmangione/fa14ff5f5627de19544fde8f05b3187d/external/bazel_tools/tools/build_defs/repo/http.bzl", line 150, column 10, in _http_archive_impl
		patch(ctx)
	File "/private/var/tmp/_bazel_fmangione/fa14ff5f5627de19544fde8f05b3187d/external/bazel_tools/tools/build_defs/repo/utils.bzl", line 196, column 18, in patch
		ctx.patch(patchfile, remote_patch_strip)
[...]

PR #507, opened 2025-01-24
fix: fix bcr rust dev patch

hofbi:
Closes https://github.com/aspect-build/rules_py/issues/506

I think this was introduced by https://github.com/aspect-build/rules_py/pull/480. We should maybe double check the `verify-bcr-patches` job and why this was not found by CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
     # Will need to bump this version in the future when there are breaking changes.
     # NB: update tools/integrity.bzl at the same time!
     env:
-      RULES_PY_RELEASE_VERSION: 1.1.0
+      RULES_PY_RELEASE_VERSION: 1.2.0
     steps:
       - uses: actions/checkout@v4
       - run: ./minimal_download_test.sh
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
     # Will need to bump this version in the future when there are breaking changes.
     # NB: update tools/integrity.bzl at the same time!
     env:
-      RULES_PY_RELEASE_VERSION: 1.1.0
+      RULES_PY_RELEASE_VERSION: 1.2.0
```
hofbi, line 67: The update to the patch does not longer work with `1.1.0`. But `1.2.0` is broken because of the patch. Not sure if we can get the `test-release` job green.

alexeagle:
I don't think you want to bump the e2e/use_release - it won't help to verify this fix.

I also spent time last night trying to understand why the patch applies cleanly in the CI test, and also locally. What a mess :/

alexeagle:
> we should maybe double check

Actually I think that's a required step 1. If we can't trust CI, then we can't know if this stays fixed and will chase it again.
I found it's actually easy to reproduce the problem first:
https://github.com/aspect-build/rules_py/pull/508

hofbi:
Closing in favor of #508

PR #508, opened 2025-01-24
fix(ci): assert that patches apply exactly

### LINKED ISSUE #506: [Bug]: aspect_rules_py 1.2.0 cannot be downloaded from BCR

alexeagle:
From `man patch`

```
     -F max-fuzz, --fuzz max-fuzz
             Sets the maximum fuzz factor.  This option only applies to
             context diffs, and causes patch to ignore up to that many lines
             in looking for places to install a hunk.  Note that a larger fuzz
             factor increases the odds of a faulty patch.  The default fuzz
             factor is 2, and it may not be set to more than the number of
             lines of context in the context diff, ordinarily 3.
```

The hint was in our CI job which passes but prints 

```
Run patch --dry-run -p1 < .bcr/patches/*.patch
checking file MODULE.bazel
Hunk #1 succeeded at 45 with fuzz 2 (offset 4 lines).
Hunk #2 succeeded at 61 with fuzz 2 (offset 3 lines).
```

When Bazel or the Publish to BCR app apply the patches, they don't use the `patch` tool so they don't allow any variation in the patch lines. Our assertion needs to be strict.

Fixes #506

### DIFF
--- a/.bcr/patches/rust_dev_dep.patch
+++ b/.bcr/patches/rust_dev_dep.patch
@@ -1,36 +1,36 @@
 diff --git a/MODULE.bazel b/MODULE.bazel
-index 841aed7..e1a1d6d 100644
+index 967dad2..5022390 100644
 --- a/MODULE.bazel
 +++ b/MODULE.bazel
-@@ -41,13 +41,13 @@ register_toolchains(
+@@ -45,13 +45,13 @@ register_toolchains(
  bazel_dep(
      name = "rules_rust",
      version = "0.53.0",
 -    # In released versions: dev_dependency = True
-+    dev_dependency = True
++    dev_dependency = True,
  )
  
  rust = use_extension(
      "@rules_rust//rust:extensions.bzl",
      "rust",
 -    # In released versions: dev_dependency = True
-+    dev_dependency = True
++    dev_dependency = True,
  )
- 
  rust.toolchain(
-@@ -58,13 +58,13 @@ use_repo(rust, "rust_toolchains")
+     edition = "2021",
+@@ -61,13 +61,13 @@ use_repo(rust, "rust_toolchains")
  
  register_toolchains(
      "@rust_toolchains//:all",
 -    # In released versions: dev_dependency = True
-+    dev_dependency = True
++    dev_dependency = True,
  )
  
  crate = use_extension(
      "@rules_rust//crate_universe:extension.bzl",
      "crate",
 -    # In released versions: dev_dependency = True
-+    dev_dependency = True
++    dev_dependency = True,
  )
- 
  crate.from_cargo(
+     name = "crate_index",
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,7 +50,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: patch --dry-run -p1 < .bcr/patches/*.patch
+      - run: patch --dry-run -p1 --fuzz 0 < .bcr/patches/*.patch
 
   test-release:
     runs-on: ubuntu-latest
--- a/e2e/use_release/minimal_download_test.sh
+++ b/e2e/use_release/minimal_download_test.sh
@@ -13,7 +13,8 @@ fi
 # Test bzlmod
 (
     cd ../..
-    patch -p1 < .bcr/patches/*.patch
+    # Create the .orig file, whether there's a mismatch or not
+    patch -p1 --backup < .bcr/patches/*.patch
 )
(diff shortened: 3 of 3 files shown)


thesayyn (approved):

alexeagle:
Also I don't understand why this test was passing before there was a "mismatch" introduced. You would expect it didn't create `.orig` files a week ago either.

hofbi:
> Also I don't understand why this test was passing before there was a "mismatch" introduced. You would expect it didn't create `.orig` files a week ago either.

Yep, that's strange. I tried both the one introducing the mismatch 1d6cd38948268fc5ea0a3d10c92e030883349825 and the one before e618b7574729f2cdead1e7c313008fe4a51d8f0d, and both applied and produce the `.orig` file. Are we looking at the wrong place?

alexeagle:
I'm moving on from this but I agree it would be better to understand what happened so we are sure we fixed the root cause

hofbi:
After some further history checking, I found that #429 disabled the release test while #496 enabled it again. #417 was the PR that made it red and updated the patch. Unfortunelty it is too long ago to check the CI log and verify if the error was the missing `.orig` file back then. Maybe there were 2 errors that compensated each other 🤷 . But seems like we won't find out without the CI logs, so let's move on.

PR #509, opened 2025-01-24
chore(ci): make it 8min faster

alexeagle:
The GHA job for the root folder is taking 12min while all the rest are 4min. We setup Aspect Workflows recently which gives us test coverage.

Note, we'll have to adjust the branch protection to gate PRs on AW passing

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,18 +25,17 @@ jobs:
   test:
     uses: bazel-contrib/.github/.github/workflows/bazel.yaml@646899086d7aaee8e532540480f3e91e00596234 # 11 Dec 2024
     with:
-      folders: '[".", "e2e/smoke", "e2e/repository-rule-deps", "e2e/system-interpreter", "examples/uv_pip_compile"]'
+      # NB: the root folder is tested with Aspect Workflows on BuildKite, see /.aspect/workflows/config.yaml
+      folders: '["e2e/smoke", "e2e/repository-rule-deps", "e2e/system-interpreter", "examples/uv_pip_compile"]'
       # TODO: Build Windows tools and add to toolchain
       # TODO(alex): switch the root folder to bzlmod
       # TODO: fix remaining folders on Bazel 8
       exclude: |
         [
           {"os": "windows-latest"},
-          {"folder": ".", "bzlmodEnabled": true},
           {"folder": "e2e/repository-rule-deps", "bzlmodEnabled": false},
           {"folder": "e2e/system-interpreter", "bzlmodEnabled": false},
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -25,18 +25,17 @@ jobs:
   test:
     uses: bazel-contrib/.github/.github/workflows/bazel.yaml@646899086d7aaee8e532540480f3e91e00596234 # 11 Dec 2024
     with:
-      folders: '[".", "e2e/smoke", "e2e/repository-rule-deps", "e2e/system-interpreter", "examples/uv_pip_compile"]'
+      # NB: the root folder is tested with Aspect Workflows on BuildKite, see /.aspect/workflows/config.yaml
+      folders: '["e2e/smoke", "e2e/repository-rule-deps", "e2e/system-interpreter", "examples/uv_pip_compile"]'
```
kormide, line 29: Do you want me to wire up any of the subworkspaces to run under workflows? I'm not sure we could do the bzlmod/workspace switch easily though, so it would only run one variant.

kormide (approved):

.github/workflows/ci.yaml
```diff
@@ -25,18 +25,17 @@ jobs:
...
       # TODO(alex): switch the root folder to bzlmod
       # TODO: fix remaining folders on Bazel 8
       exclude: |
         [
           {"os": "windows-latest"},
-          {"folder": ".", "bzlmodEnabled": true},
           {"folder": "e2e/repository-rule-deps", "bzlmodEnabled": false},
           {"folder": "e2e/system-interpreter", "bzlmodEnabled": false},
           {"folder": "examples/uv_pip_compile", "bzlmodEnabled": false},
-          {"folder": ".", "bazelversion": "8.0.0"},
```
kormide, line 39: Are you okay with not running the WORKSPACE variant on CI anymore?

.github/workflows/ci.yaml
alexeagle: possibly in a follow-up yes

.github/workflows/ci.yaml
alexeagle: i believe AW is testing workspace w bazel 7

alexeagle: Yup I'm pretty confident we aren't losing any test coverage here.

.github/workflows/ci.yaml
alexeagle: as discussed in meeting - let's not bother with this anytime soon. We got the value we want here already, and will always have some amount of GHA on this repo

PR #515, opened 2025-01-26
Add `strip_prefix` parameter to `py_image_layer`

jimmyt857:
By default, bazel lays out files using the target name as the root for all runfiles. For example, the target //a/b/c:my_bin has its runfiles directory located at $BAZEL_ROOT/a/b/c/my_bin.runfiles/. When placed into an oci_image this becomes /a/b/c/my_bin.runfiles/.

The problem with the above is that it prevents us from sharing layers between separate images. Even if no `layer_groups` are provided, at a minimum the contents of the interpreter layer are identical across images, but can't be shared because the path prefixes of the files are different.

The primary change in this PR is to introduce a strip_prefix option to the `py_image_layer` macro that allows us to remove prefixes from the directories laid out in the image. Combined with the existing `root` option, we can align directory paths across images and produce identical layer hashes.

Also changes the template strings to use named replacement syntax, for readability.

---

### Test plan

- New test cases added


py/private/py_image_layer.bzl
```diff
@@ -44,40 +44,40 @@ default_layer_groups = {
...
+{{
     # Exclude .whl files from container images
-    if ($$1 ~ ".whl") {
+    if ($$1 ~ ".whl") {{
         next
-    }
+    }}
     # Move everything under the specified root
-    sub(/^/, ".%s")
+    sub(/^{strip_prefix}/, ".{root}")
```
jimmyt857, line 74: Note: this line is the primary change. Here we append `strip_prefix` to the beginning matcher, so that it is taken out of the path.

jimmyt857:
CC @thesayyn

py/tests/py_image_layer/BUILD.bazel
```diff
@@ -26,3 +26,18 @@ assert_tar_listing(
     actual = [":my_app_layers"],
     expected = ":my_app_layers.listing",
 )
+
+# Case 2: With `strip_prefix` and `root`
+py_image_layer(
+    name = "my_app_layers_strip_prefix",
+    binary = ":my_app_bin",
+    platform = ":linux_amd64",
+    strip_prefix = "py\\/tests\\/py_image_layer\\/my_app_bin",
```
thesayyn, line 35: it should be possible to pass this string to awk `-v strip_prefix=` so that we don't have to escape slashes.

thesayyn: basically pass this information to awk using its `-v strip_prefix=/py/test/my_app_bin` and use `strip_prefix` variable within awk script.

py/private/py_image_layer.bzl
thesayyn: strip_prefix needs more than this i believe,  see an example logic. 

https://github.com/bazel-contrib/bazel-lib/blob/c6707f1de817296befc9c238ddb0c27ed43a7c46/lib/private/modify_mtree.awk#L3-L31

PR #518, opened 2025-01-27
Only remove the venv_root directory if it has contents

kriscfoster:
When a k8s pod is configured with readOnlyRootFilesystem, we hit the following error:
```
Unable to create base venv directory
Read-only file system.
```

We then mounted a volume under the path where the venv directory was being created. We then hit:
```
Unable to remove venv_root directory
Read-only file system.
```

I think this should solve the issue. Basically, only try to delete the existing venv directory if it is non-empty. In our case, it is empty (when we are mounting an empty directory as a volume).

However, I'm not totally sure how I can actually test this in our environment without a release. I tried adding a patch for `py/tools/py/src/venv.rs` but that didn't work because I guess this actually gets bundled as a binary as part of the release. Do you have any hints for us to test this change? 

---

### Test plan

- New test cases added

### DIFF
--- a/py/tools/py/src/venv.rs
+++ b/py/tools/py/src/venv.rs
@@ -21,7 +21,7 @@ pub fn create_venv(
         // With readOnlyRootFilesystem (k8s), we mount a writable volume here with an empty directory.
         // In that case, we cannot delete that directory.
         let is_empty = location.read_dir()?.next().is_none();
-        if is_empty.not() {
+        if !is_empty {
             // Clear down the an old venv if there is one present.
             fs::remove_dir_all(location)
                 .into_diagnostic()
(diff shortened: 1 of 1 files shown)


kriscfoster:
My knowledge of rust is limited so looks like there's some issue. I'll work on fixing it in this PR tomorrow.

alexeagle:
If you switch to pulling rules_py from a SHA rather than a release, you'll get the "pre-release" which exposes the rust sources to your build and you can test this from HEAD.

Or of course you can reproduce your problem within the examples folder here, which always uses a pre-release

thesayyn:
I am curious about your end goal, ultimately we want to generate the venv runtime so it works with readonly filesystems.

arrdem:
This should be obviated as of #551.

Issue #377, opened 2024-08-06
[Bug]: `bazel-runfiles` pypi library does not work


dizzy57:
### What happened?

`bazel-runfiles` library is unable to properly locate runfiles root when running under `rules_py`

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: `bazel 7.2.1`

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: `0.7.4`

Language(s) and/or frameworks involved: None

### How to reproduce

```shell
git clone https://github.com/dizzy57/bazel_py_minimal
cd bazel_py_minimal
bazel test //runfiles_from_pipy_do_not_work/...

Expected result: all tests pass
Actual result: //runfiles_from_pipy_do_not_work:runfiles_test fails with a message
ValueError: [PREFIX]/runfiles_test.runfiles/_main/runfiles_from_pipy_do_not_work/test_runfiles.py does not lie under the runfiles root [PREFIX]/runfiles_test.runfiles/.runfiles_test.venv/lib
```

### Any other information?

Same test passes with `rules_python`: see `bazel test //runfiles_from_pipy_do_not_work:runfiles_rp_test`

[Exception source in runfiles.py](https://github.com/bazelbuild/rules_python/blob/fa13b0138924245c55f79922509e4688f252ff71/python/runfiles/runfiles.py#L250) (`rules_python` v0.34.0)

The `python.runfiles` package from `@rules_python//python/runfiles` works, see `bazel test //runfiles_from_pipy_do_not_work:runfiles_rfrp_test //runfiles_from_pipy_do_not_work:runfiles_rp_rfrp_test`

This problem was introduced in rules_py v0.7.2

mattem:
This is a rules_python issue I think, as the runfiles helper is walking up the directory structure to attempt to determine the runfiles root, which under the venv (or any other layout) is not correct. The diff here shows a potential solution, but I'm not clear if this is correct, or what repercussions this has, I imagine Windows will have something to say about it.

https://github.com/bazelbuild/rules_python/pull/2115

Continuing the discussion with other rules_python maintainers.

Zemnmez:
I'm getting this same issue:
```
Traceback (most recent call last):
[...]

PR #519, opened 2025-01-27
fix: write __init__ shim for pip runfiles helper

thesayyn:
Fixes https://github.com/aspect-build/rules_py/issues/377 

Even though messing with __init__.py is a little risky, in this case its safer since we only generate it while symlinking files from external repositories into venv, and since https://github.com/bazelbuild/rules_python/blob/18f76f9cc4c3c6c4470f7a881e6ea4e8b80d7bab/python/runfiles/__init__.py#L15 is simple line, i can't think of easier fix than this.

### Test plan

- New test cases added


py/tools/py/src/pth.rs
```diff
@@ -6,6 +6,23 @@ use std::{
 
 use miette::{miette, Context, IntoDiagnostic, LabeledSpan, MietteDiagnostic, Severity};
 
+const RULES_PYTHON_INIT_PATH: &str = "runfiles/__init__.py";
+const RULES_PYTHON_RUNFILES_INIT_SHIM: &str = r#"
```
thesayyn, line 10: i am not even sorry. https://www.youtube.com/watch?v=4-B2gUnL4SA

py/tools/py/src/pth.rs
```diff
@@ -6,6 +6,23 @@ use std::{
...
+# Generated by rules_py venv for rules_python compatibility
+# See: https://github.com/bazelbuild/rules_python/pull/2115
+# See: https://github.com/aspect-build/rules_py/issues/377
+from . import runfiles
+def _FindPythonRunfilesRoot():
```
alexeagle, line 15: I'm okay with it. Could you add a bunch more commenting for our future selves to understand what it does?

IN particular, this totally replaces a file from the runfiles package? Or is there no such file in that package?

alexeagle (approved):

py/tools/py/src/pth.rs
thesayyn: done

PR #521, opened 2025-01-28
fix: workaround Bazel lcov parsing bug

alexeagle:
https://github.com/bazelbuild/bazel/issues/25118 was causing our coverage reports to fail to parse into combined report.

Related to #520 which was missing testing that the report actually worked.
fyi @dizzy57

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,17 +67,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: ./minimal_download_test.sh
-      - run: bazel coverage //src/...
+      - run: bazel coverage --combined_report=lcov //src/...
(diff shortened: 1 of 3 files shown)


alexeagle:
Wow after all that fighting with it this morning, turns out it's a fundamental and obvious bug in Bazel's Lcov parser https://github.com/bazelbuild/bazel/issues/25118

dizzy57:
Oh yes, we hit same issue using recent versions of `coverage.py`, we downgraded to the one pinned in `rules_python`.

alexeagle:
@dizzy57 yup, https://github.com/nedbat/coveragepy/commit/6118798a5702a90dd89b726d44d4696bb0f82eec introduced this, but also a lot of other improvements that look worthwhile. Obviously rules_py doesn't want to force an upper-bound constraint on our user's requirements :)

Issue #522, opened 2025-01-28
[FR]: Generate venv at build time


thesayyn:
### What is the current behavior?

This is a tracking issue for all things related to venv generation at build time.  Biggest issue so far is when running py_binary put into a docker image on Lambda, it simply does not work since all the paths on Lambda is write protected. 

Current issues with venv

https://github.com/aspect-build/rules_py/issues/433
https://github.com/aspect-build/rules_py/issues/373
https://github.com/aspect-build/rules_py/issues/351
https://github.com/aspect-build/rules_py/issues/339

### Describe the feature

Generate the venv at Build time by making the venv relocatable.

thesayyn:
Blocker: https://github.com/aspect-build/rules_py/issues/339#issuecomment-2452525854

vinnybod:
I **think** I am blocked on this.

I am trying to create a python environment on an OCI image that has a few dependencies pre-installed. It doesn't seem possible to have the venv already created in the image. It also doesn't seem possible to symlink the venv in the `$PATH` and still contain the 3rd party dependencies.

https://github.com/vinnybod/bazel-examples/tree/python-oci/python-oci

Issue #339, opened 2024-05-08
[FR]: Remove need for runfiles directory to be writable


jsharpe:
### What is the current behavior?

The venv is recreated whenever the binary is executed. This prevents packaging up the binary and installing it to a system outside of bazel as an application deployment that can be run by other users from a common directory e.g. an /apps directory, possibly served read-only over nfs or similar.

### Describe the feature

I would like either for the venv creation to fallback to a user writable location e.g. XDG_CACHE or to remove the need to recreate the venv links so it can be done once by an admin user and then subsequent users can just execute from the precreated venv.

jsharpe:
Another place where this could be used is in an immutable container - the container is built and then at execution time the container image is made immutable and so writing to the runfiles directory would potentially not be possible in this case too.

jsharpe:
Also the current behaviour would prevent multiple copies of a binary running at the same time since it would delete files being used by the first process on the execution of the second process.

voxeljorge:
We've run into this issue with when running `py_binary` targets from this ruleset in an immutable environment. One example is when running code in a os-only aws lambda runtime, the lambda is unpacked in an immutable env and the py_binary fails to run.

voxeljorge:
@alexeagle  this issue is stopping us from migrating a couple of our lambdas to rules_py, and it seems somewhat straightforward to resolve. The runner could accept an env var as proposed above. Either `XDG_CACHE` or something specifically named after `rules_py` seems reasonable, any specific preference?

thesayyn:
Little bit of context and progress here; 

We have recently started using uv for venv generation which supports relative paths https://github.com/astral-sh/uv/pull/3717   it's still experimental but it's almost there. 

But we are still blocked on https://github.com/python/cpython/issues/83650 which still forces us to use an absolute path `pyenv.cfg` file, bummer. 


However, i have been hitting my head against the wall trying to solve this and we have solved this with @voxeljorge.  It's approximately https://github.com/aspect-build/rules_py/pull/427 but needs a static symlink from `/tmp/venv` back to runfiles tree to allow venv being created in `/tmp/venv` but available in the runfiles tree.

arrdem:
Some updates here. Removing the need for the virtualenv to be writable requires two things, a relocatable "python" a take at which is #546, and to assemble the `site-packages` and other required directory trees at build time which is what I was prototyping on last week.

[...]

PR #546, opened 2025-04-04
feat(venv): Create a relocatable venv shim

arrdem:
Related to #522 and associated issues.

This PR introduces a new venv tool, what for lack of better words I'm calling the "shim".

While standard interpreters do support the use of relative paths to the interpreter in `pyvenv.cfg` -- for instance `home = ./bin/python` is legal -- we still need a `bin/python` which is portable. Existing relocatable virtual environment solutions (uv, conda) still ultimately create both an absolute path reference in `pyvenv.cfg` and usually in the `bin/python` symlink to a specific interpreter on the filesystem. UV does this statically as is standard, Conda will also do this statically with an explicit update process as part of the conda unpack.

We can't create relocatable symlinks, and we also can't dynamically correct static symlinks without continuing to have #339 as a problem.

What we can do is dynamically identify a Python interpreter and hoodwink it with regards to its own path, causing it to load a virtualenv which is itself entirely relocatable.

As with many other tools, the Python interpreter introspects its `argv[0]` to figure out its own nominal path and identify the home, see [2], [3] for the gory details. This means that when the interpreter is invoked via a symlink under normal execution, the "path" of the interpreter per `argv[0]` is the path of the symlink with respect to which `pyvenv.cfg` can be identified and the virtualenv activated automatically. The Darwin platform provides the `_NSGetExecutablePath` libc call which is an alternative mechanism for determining the "path" by which the interpreter is invoked.

This PR introduces a shim tool which can be emplaced as the target of `bin/python`, which will consult the `pyvenv.cfg` of a conventional virtualenv to find the requested interpreter version, and will attempt to delegate to an identified interpreter in such a way as to make the interpreter believe that its path is that of the shim tool.

On a conventional unix this is as simple as lying about the value of `argv[0]`, on Darwin the `PYTHONEXECUTABLE` environment flag must be set to make Python disbelieve the value of `_NSGetExecutablePath` which is harder to hoodwink.

Using this tool, a relocatable virtualenv can be structured as follows

```
./pyvenv.cfg                        # conventional
./bin/python -> ./aspect_venv_shim  # customized "interpreter"
./bin/python3 -> ./python           # conventional
./bin/python3.{N} -> ./python        # conventional
./bin/aspect_venv_shim
./lib/python3.${N}/site-packages/...                 # conventional; standard contents
```

We should be able to create one of these venvs by updating our `uv` dependency, setting the `relocatable=True` flag when creating virtualenvs, and attempting to specify the custom interpreter path of `./bin/aspect_venv_shim` so that the "python" symlink and its siblings will enter this pipeline.

Using this as our "interpreter" will allow us to pull venv creation forwards from runtime to a normal build action, and allow us to create conda-pack like structures which require no unpack post-processing directly from that venv.

The downside of this approach is that as with other `exec` based Python launchers such as Pex or Bazel's `--run_under` it is likely to interfere with debugging tools that want to instrument a Python process, as the bootloader is not an interpreter and cannot be analyzed as such.

### Test plan

- [x] Built for MacOS; made a relocatable venv, customized the `pyvenv.cfg` to be `home = ./bin/python`, updated the `bin/python` link to be `bin/aspect_venv_shim`, confirmed that invoking the venv's python links did cause the venv to load.
- [x] Built for Linux; made a relocatable venv, customized the `pyvenv.cfg` to be `home = ./bin/python`, updated the `bin/python` link to be `bin/aspect_venv_shim`, confirmed that invoking the venv's python links did cause the venv to load.

### Notes
[1] https://discuss.python.org/t/interpreter-independent-isolated-virtual-environments/5378/53
[2] https://github.com/python/cpython/blob/main/Modules/getpath.py#L293
[3] https://github.com/python/cpython/blob/main/Modules/getpath.c#L774

### DIFF
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "81a92b86099c7c9b831ce831b33ad24c552a0c97ef3d8dc37836db3d08529ec4",
+  "checksum": "7c8cd2683de76ab9fae29a74881389aefee00f8d83f53bd006aba44b4dd3731d",
   "crates": {
     "addr2line 0.24.2": {
       "name": "addr2line",
@@ -23131,6 +23131,33 @@
       "license_ids": [],
       "license_file": null
     },
+    "venv_shim 0.1.0": {
+      "name": "venv_shim",
+      "version": "0.1.0",
+      "package_url": "https://github.com/aspect-build/rules_py",
+      "repository": null,
+      "targets": [],
+      "library_target_name": null,
+      "common_attrs": {
+        "compile_data_glob": [
+          "**"
+        ],
+        "deps": {
+          "common": [
+            {
+              "id": "miette 7.2.0",
+              "target": "miette"
+            }
+          ],
+          "selects": {}
+        },
+        "edition": "2021",
+        "version": "0.1.0"
+      },
+      "license": "Apache 2",
+      "license_ids": [],
+      "license_file": null
+    },
     "version_check 0.9.5": {
       "name": "version_check",
       "version": "0.9.5",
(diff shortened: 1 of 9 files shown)


thesayyn (approved):
LGTM, nice!

PR #551, opened 2025-04-22
feat(venv): Static site-packages trees

arrdem:
This PR introduces a new suite of virtualenv machinery building on the existing `py_binary` virtualenv tree builder to create such trees "statically" at build time rather than "dynamically" at runtime. The original reason this is difficult is that the `$VENV/bin/python` file is usually a link to a static interpreter, and the interpreter specified in `pyvenv.cfg` is also usually an absolute path to a real interpreter outside of the venv.

The introduction of #546 or something like it makes it possible to create a virtualenv tree which _does not_ refer to an interpreter with a fixed path. Instead an appropriate interpreter can be dynamically discovered by `$PATH` inspection. Support for other configuration ala `/etc/pexrc` to facilitate interpreter discovery is possible but beyond the scope of an initial implementation.

By using an interpreter shim, the `pyenv.cfg` file can use a relative path to `./bin/python` as its interpreter reference, and the `./bin/python3` and `./bin/python3.X` links can be created statically pointing to the shim. As long as other `bin/` entrypoint scripts reference only `#!/usr/bin/env python3` as is fairly conventional, the constructed virtual environment will contain no absolute references to a specific interpreter and will be fully relocatable.

A relocatable virtualenv tree containing dependencies installed into the `site-packages` tree may be built using the new experimental `py_venv` rule.

```
py_venv(
    name = "venv",
    deps = [
        ... # PyInfo targets to be integrated into the site-packages tree
    ],
)
```

A conda-pack like packaging of this virtualenv into a prefix-less zipfile may be requested using the `zip` output group, ala

```
filegroup(
    name = "venv-pack",
    srcs = [":venv"],
    output_group = "zip"
)
```

 A `py_venv` is not intended to be a `.runfiles`-style packaging of all source code, it is intended to be a packaging of dependency code which is compliant with being treated as if `pip install`ed. This may not be generally true of firstparty application code, depending on usage of `runfiles.Runfiles` or repository-relative logics.

Instead applications wishing to leverage a static virtualenv should leverage the `py_venv_binary` rule.

```
py_venv_binary(
    name = "demo",
    srcs = ["demo.py"],
    deps = [ ... ],
    main = "demo.py",
    virtualenv = ":venv",
)
```

A `py_venv_binary` defines a more conventional `.runfiles`-based Python application, but one which will use the `.runfiles` embedded virtualenv as the source of thirdparty packages. Firstparty packages will receive the normal `.runfiles` workspace structure preserving treatment.

### Test plan

- New test cases added


py/private/py_image_layer.bzl
```diff
@@ -141,24 +130,11 @@ def py_image_layer(
 
     # Produce the manifest for a tar file of our py_binary, but don't tar it up yet, so we can split
     # into fine-grained layers for better pull, push and remote cache performance.
-    manifest_name = name + ".manifest"
```
thesayyn, line 144: hmm why change this? we recently landed a change to fix this; https://github.com/bazel-contrib/bazel-lib/commit/8a9204ab17cbe337a64e838dd193b00f243764fb i think we should keep this as is.

py/private/py_venv.bzl
```diff
@@ -1,172 +1,401 @@
...
 load("@aspect_bazel_lib//lib:paths.bzl", "BASH_RLOCATION_FUNCTION", "to_rlocation_path")
-load("@bazel_skylib//lib:paths.bzl", "paths")
-load("@rules_python//python:defs.bzl", "PyInfo")
-load("//py/private:providers.bzl", "PyVirtualInfo")
 load("//py/private:py_library.bzl", _py_library = "py_library_utils")
 load("//py/private:py_semantics.bzl", _py_semantics = "semantics")
-load("//py/private/toolchain:types.bzl", "PY_TOOLCHAIN", "VENV_TOOLCHAIN")
+load("//py/private/toolchain:types.bzl", "PY_TOOLCHAIN")
+
+VirtualenvInfo = provider(
```
thesayyn, line 9:
```suggestion
VirtualEnvInfo = provider(
```

py/private/py_venv.bzl
```diff
@@ -1,172 +1,401 @@
...
+    # NB: rules_python ALSO includes this attribute on the py_binary rule, but we think that's a mistake.
+    # see https://github.com/aspect-build/rules_py/pull/520#pullrequestreview-25790761972
+    "_lcov_merger": attr.label(
+        default = configuration_field(fragment = "coverage", name = "output_generator"),
+        executable = True,
+        cfg = "exec",
+    ),
+})
+
+def _python_version_transition_impl(_, attr):
```
thesayyn, line 354: can this transition live in a single place?

py/private/venv.tmpl.sh
```diff
@@ -1,35 +1,15 @@
...
-    --location "${VIRTUAL_ENV}" \
-    --python "$(alocation $(rlocation {{ARG_PYTHON}}))" \
-    --pth-file "$(rlocation {{ARG_PTH_FILE}})" \
-    --pth-entry-prefix "${RUNFILES_DIR}" \
-    --collision-strategy "{{ARG_COLLISION_STRATEGY}}" \
-    --venv-name "{{ARG_VENV_NAME}}"
+{{PREEXEC}}
 
-echo "Created virtualenv in ${VIRTUAL_ENV}"
+exec "$(rlocation "{{VENV}}")"/bin/python {{INTERPRETER_FLAGS}} "$@"
```
thesayyn, line 15: previously this rule was used for IDE support via `bazel run :venv` to create the .venv symlink in the source tree. we are probably breaking those people here.

py/tools/py/src/venv.rs
```diff
@@ -62,7 +66,489 @@ pub fn create_venv(
...
+                major: major.parse().unwrap(),
+                minor: minor.parse().unwrap(),
+                patch: 0,
+            }),
+            [major, minor, patch] => Ok(PythonVersionInfo {
+                major: major.parse().unwrap(),
+                minor: minor.parse().unwrap(),
+                patch: patch.parse().unwrap(),
+            }),
+            _ => panic!("X.Y or X.Y.Z required!"),
```
thesayyn, line 95: return an Err instead?

py/tools/py/src/venv.rs
```diff
@@ -62,7 +66,489 @@ pub fn create_venv(
...
+    site_dir: PathBuf,
+    python_bin: PathBuf,
+}
+
+fn link(original: PathBuf, link: PathBuf) -> miette::Result<()> {
+    let build_dir = current_dir().into_diagnostic()?;
+    let original_abs = build_dir.clone().join(original.clone());
+    let link_abs = build_dir.clone().join(link.clone());
+
+    let original_relative = diff_paths(original_abs.clone(), link_abs.parent().unwrap()).unwrap();
```
thesayyn, line 133: you could probably avoid these `clone`s by passing them via reference.

thesayyn (commented):
Overall looking good, a few things, rust can be better but i think its good enough for now.

py/private/py_venv.bzl
```diff
@@ -1,172 +1,402 @@
...
+
+    py_toolchain = _py_semantics.resolve_toolchain(ctx)
+
+    # Check for duplicate virtual dependency names. Those that map to the same resolution target would have been merged by the depset for us.
+    virtual_resolution = _py_library.resolve_virtuals(ctx)
     imports_depset = _py_library.make_imports_depset(ctx, extra_imports_depsets = virtual_resolution.imports)
 
     pth_lines = ctx.actions.args()
-    pth_lines.use_param_file("%s", use_always = True)
+    pth_lines.use_param_file("%s/", use_always = True)
```
thesayyn, line 81:
```suggestion
    pth_lines.use_param_file("%s", use_always = True)
```

py/private/py_venv.bzl
```diff
@@ -1,172 +1,402 @@
...
     pth_lines.set_param_file_format("multiline")
-
-    # A few imports rely on being able to reference the root of the runfiles tree as a Python module,
-    # the common case here being the @rules_python//python/runfiles target that adds the runfiles helper,
-    # which ends up in bazel_tools/tools/python/runfiles/runfiles.py, but there are no imports attrs that hint we
-    # should be adding the root to the PYTHONPATH
-    # Maybe in the future we can opt out of this?
-    pth_lines.add(".")
```
thesayyn, line 36: lets double check if this is breaking

py/tools/venv_bin/src/main.rs
```diff
@@ -55,17 +70,63 @@ struct VenvArgs {
...
+                args.bin_dir.unwrap(),
+                args.collision_strategy.unwrap_or_default().into(),
+            )?;
+
+            Ok(())
+        }
+        VenvMode::StaticPth => {
+            let venv = py::venv::create_empty_venv(
+                &args.python,
+                py::venv::PythonVersionInfo::from_str(&args.version.unwrap())?,
```
thesayyn, line 118: .unwrap() here sounds dangerous. do we want to check and print an error message?

thesayyn (approved):

thesayyn:
LGTM, nice job!

thesayyn (approved):
LGTM

py/tools/py/src/venv.rs
```diff
@@ -62,7 +66,489 @@ pub fn create_venv(
...
+            let link_entry = link_dir.join(&relative_entry);
+
+            // link() makes dirs for us, but maybe it shouldn't. Do it manually here
+            if original_entry.canonicalize().unwrap().is_dir() {
+                continue;
+            }
+            // Handle collisions, probably conflicting empty __init__.py files from Bazel
+            else if link_entry.exists() {
+                return Err(miette!(
+                    "Collision resolution not yet implemented! Saw {} twice",
```
honnix, line 531: Is this a way to ignore this type of error?

honnix: Hmm, setting `package_collisions` doesn't seem to work.

PR #564, opened 2025-05-28
chore(publish-to-bcr): Update

arrdem:
For https://github.com/bazel-contrib/publish-to-bcr/pull/277

### Test plan
N/A


alexeagle:
Sorry, was in a rush on mobile yesterday so I didn't add that backlink ;)

PR #568, opened 2025-05-29
fix(release): Configure the interpreter shim via a prebuilt toolchain

arrdem:
This fixes a breaking regression in v1.5.0 caused by the new interpreter shim being referenced by source label and not via a toolchain which can be supplied with prebuild artifacts.

Included in this change is a refactor consolidating our source of truth for the list of prebuilt tools to configure down to `tools.bzl` which is now shared with `tools/release` among other things, preventing how I got myself into this mess.

As a further bit of paranoia this changeset also reworks the integration tests so that instead of doing e2e testing with assets from a checkpoint release (previously 1.1.0) we instead do a full build of the release candidate assets, update the asset integrity file, run a local fileserver hosting them and then do builds using the locally hosted assets.

### Test plan

- [x] Existing test cases still green
- [x] Preexisting and inadequate e2e tests updated to cover this case

### DIFF
--- a/e2e/use_release/minimal_download_test.sh
+++ b/e2e/use_release/minimal_download_test.sh
@@ -107,4 +107,4 @@ bazel test --test_output=streamed //...
 )
 
 # Shut down the devserver
-kill "$PIDFILE"
+kill "$(cat $PIDFILE)"


tools/release/BUILD.bazel
```diff
@@ -18,16 +22,15 @@
     ]
 ]
 
+# FIXME: This is kinda a mess. Should just be done with transitions? These can crossbuild.
```
alexeagle, line 25: yeah but it's simpler than transitions I think?

tools/release/BUILD.bazel
```diff
@@ -1,3 +1,7 @@
+# buildifier: disable=bzl-visibility
+# Since this is a load from private despite it being our private... sigh
+load("//py/private/toolchain:tools.bzl", "TOOL_CFGS")
```
alexeagle, line 3: maybe the TOOL_CFGS wants to move out of private for this reason?

PR #569, opened 2025-05-29
chore(ci): Mostly replace GHA with Aspect Workflows

arrdem:
As it says on the tin. Aspect Workflows benefits from stateful runners which can cache hit WAY faster than GHA's stateless runners will.

---

### Test plan

This is the test plan.


alexeagle:
I had setup GHA here to report the coverage data, we should ensure that continues to work with AW publishing the coverage data to codecov (or anywhere else)

PR #571, opened 2025-05-30
fix(py_venv): Need venv in the exec config

arrdem:
The 1.5.1 release was bad for similar reasons to the 1.5.0 release -- the new `py_venv` machinery was taking a source dep on a tool which rather than pull a prebuilt artifact created an unmet dep on `rules_rust` to implement a source build.

This second time it was because the `venv` builder tool rather than the interpreter `shim` wasn't depended on via a toolchain. This however is slightly harder to fix because while the `shim` tool is used in the target configuration strictly, the `venv` tool is used in _both_ the target and exec configurations. The original `py_binary` machinery uses it in target, and the new `py_venv` machinery uses it in exec to build the venv ahead of time.

That both of these problems were present in 1.5.1 wasn't detected in integration testing because the final batch of integration test checks were missing the `/external/` spot check to make sure that Rust hadn't been fetched.

This changeset does a couple things:
- (refactor) Introduce a new Info provider so that our toolchain bins are returning an Info as they should not raw Files
- (fix) Create a new rule label which refers to the resolved file of the `venv` toolchain, and depend on it in `exec` from `py_venv` so that we can get a prebuilt `venv` binary but will always choose one in the correct configuration for local usage without doing a source build 
- (fix) Register the new shim toolchains in our own MODULE.bazel
- (fix) Don't revert the bzlmod dev-dep patches before running the tests for integration
- (fix) Check that running the tests for integration didn't pull rust
- (fix) Force tests across platform boundaries to exercise that we didn't break crossbuilding

---

### Test plan

- [x] New test cases added
- [x] Existing cases extended to cover previously untested behavior


thesayyn (approved):
neat

Issue #573, opened 2025-06-01
[Bug]: py_venv rule does not create the virtual environment correctly in v 1.5.2


pouyana:
### What happened?

Hi,

I tried to use the latest version of rules_py (1.5.2). Sourcing the created venv, does not add the (.venv.venv) at the beginning of the prompt and my own Bazel built libraries (added as  deps to the py_venv) cannot be imported in the virtual environment. The latest version that worked with the exact same code is 1.4.0. 

Thanks,
Pouyan

### Version

Development (host) and target OS/architectures: Ubuntu 24.04/amd64

Output of `bazel --version`: bazel 8.2.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

```
"""Contains all the spider global information module"""

module(
    name = "spider",
    version = "0.0.1",
    compatibility_level = 1,
)

bazel_dep(name = "rules_python", version = "1.4.1")
bazel_dep(name = "aspect_rules_py", version = "1.5.2")

python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(
    python_version = "3.12",
)

pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "pypi",
    python_version = "3.12",
    requirements_lock = "//:requirements_lock.txt",
)

use_repo(pip, "pypi")
```

Language(s) and/or frameworks involved:

### How to reproduce

```shell
* create a simple py_library
* add the py_library as dependency to py_venv
* run the py_venv rule
* source the venv (activate)
-> (venv.venv) is not visible in the shell prompt after activation in v 1.5.2.
-> (venv.venv) is visible in the shell prompt after activation in v 1.4.0.
* try to import the py_library in the the venv
-> Will get module not found error in 1.5.2
-> imports the library without issue.
```

arrdem:
This specific issue was resolved in 1.6.0, although there are other challenges (requiring explicit venv activation) with that solution.

PR #579, opened 2025-06-04
test(py_venv): Assert re: relocations

arrdem:
Fails to repro #573. Add tests covering it nonetheless.

---

### Test plan

- [x] New test cases added

### DIFF
--- a/py/tests/py_venv_conflict/BUILD.bazel
+++ b/py/tests/py_venv_conflict/BUILD.bazel
@@ -1,5 +1,16 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
-load("//py/unstable:defs.bzl", "py_venv")
+load("//py:defs.bzl", "py_library")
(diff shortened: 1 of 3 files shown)


py/tools/venv_shim/src/main.rs
```diff
@@ -148,30 +187,27 @@ fn main() -> miette::Result<()> {
...
-    let mut cmd = Command::new(&interpreter_path);
+    let mut cmd = Command::new(actual_interpreter_path);
     cmd.args(exec_args);
 
     // Lie about the value of argv0 to hoodwink the interpreter as to its
     // location on Linux-based platforms.
-    if cfg!(target_os = "linux") {
-        cmd.arg0(&exe_path);
-    }
+    cmd.arg0(&venv_interpreter_path);
```
thesayyn, line 210: hmm, should this be cfg! not target_os  = "macos"?

arrdem: Testing shows that it doesn't have to be.

PR #597, opened 2025-06-16
chore: Down with WORKSPACE

arrdem:
This is a massive combined patch aimed at starting to address a whole pile of WORKSPACE related techdebt and migration pain points.

Major points:
- Delete the `WORKSPACE` file
- Delete `WORKSPACE` supporting private files
- Retain `WORKSPACE` supporting `py/repositories.bzl` for now; will drop in future 2.0 release
- Synchronize how we're configuring Rust and the crossbuild toolchain(s) with the Aspect CLI, other projects

As future items we can go through and consider bumping dependencies

### Test plan

- This is the test plan

### Remaining items
- [x] Eliminate dev proto dependencies
- [x] Drop source build Gazelle for prebuilt  
- [x] Get local tests to green
- [x] Inline out `include()` statements which block subrepo/e2e tests (cf. bazelbuild/bazel#27137)
- [x] Update the release generation flow to discard all the dev dependencies and retain only the desired prod MODULE.bazel content
- [x] Test the release artifact generation flow; ensure release artifact naming expectations are preserved

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: patch --dry-run -p1 --fuzz 0 < .bcr/patches/*.patch
-        
+
   # For branch protection settings, this job provides a "stable" name that can be used to gate PR merges
   # on "all matrix jobs were successful".
   conclusion:


Simon-1-1:
If you figure out why you are getting that first error about `dev_pip`, it would be much appreciated if you share the solution. I'm running into it in a different scenario. I'm setting up my original stardoc target, but in your case it seems you just triggered it by adding a pip_hub. I haven't found any other person having the same issue, which is why I'm posting here.

PR #612, opened 2025-07-23
feat(telemetry): Integrate tools_telemetry

arrdem:
This patch integrates [tools_telemetry](https://github.com/aspect-build/tools_telemetry) into `rules_py`. Depending on `rules_py` will register a dependency into the telemetry system and produce a usage report when external repositories are invalidated.

`aspect_tools_telemetry` is now used for coarse grained usage tracking.

### Test plan

N/A.


alexeagle (approved):
🚢

PR #613, opened 2025-07-24
Fix debug logging in release

arrdem:
debug_build was enabled for users who were using `--compilation_mode=dbg`, which lead to `set -x` being set in the terminal used to create the venv, and tons of logging users don't need to see. The help for the venv tool was also printed.

Replaces #596 which has conflicts.

### Test plan

- New test cases added


alexeagle (approved):

arrdem:
Thanks @keith!

PR #644, opened 2025-09-23
feat(py_venv): Replace untennable copying with symlinks

arrdem:
As reported by customers, the naive but correct strategy of using copies in `py_venv_*` can lead to laughable disk usage. Some clients are reporting order 10min slowdowns and order 100GiB disk usage wasted copying inputs into binaries. We need a more scalable strategy such as symlinking.

Thankfully we can generate symlinks from tools driven by Bazel into a TreeArtifact so long as the symlinks aren't dangling. By carefully crafting relative symlinks we're able to produce a tree of links which is valid both at and after action time. When relocating a `.runfiles` tree containing such links (for instance into a OCI later tar) these links must be dereferenced but that Just Works.

While I'm at it, refactor the venv machinery to operate in terms of strategies and combinators on strategies so that it's simpler to talk about the production-grade behavior we want which is:

* `site-packages` trees in 1stparty code get relocated/linked into the venv
* `bin` sibling trees in 1stparty code get relocated/patched into the venv
* General trees in 1stparty code are referred to by `.pth` file entries
* General trees in 3rdparty code get relocated/linked into the venv
* `bin` sibling trees in 3rdparty code get relocated/patched into the venv

This makes the venv builder significantly more flexible, allows for better error reporting and opens the door to more flexible error handling.

Incorporates an implementation of #606, but testing is required.
Should include an implementation of #635, but testing is required.

`py_venv_*` now use symlinks rather than hard file copies which radically reduce disk usage while improving venv building performance.

### Test plan

- New test cases added
TODO.

### Remaining work
- [x] Strip debug prints
- [x] Improve collision handling
- [x] Rework the command interpreter to implement the last-wins semantics
- [x] Mitigate spooky dangling symlink issues 
- [x] Fix a regression which can cause a `site-packages/__init__.py` file to be linked
- [x] Add sha256-sum based collision ignoring
- [ ] Add a test covering that a `site-packages/__init__.py` file will not be linked
- [ ] Add a test covering bin shebang patching
- [ ] Integrate the test case from #635
- [ ] Manually test that linked venvs still work; should just be fine

### DIFF
--- a/py/tools/py/src/venv.rs
+++ b/py/tools/py/src/venv.rs
@@ -934,7 +934,6 @@ pub fn populate_venv<A: PthEntryHandler>(
                 // joined to the dirname. Without explicitly taking the parent
                 // we're off by 1.
                 let resolved = diff_paths(&src, &dest.parent().unwrap()).unwrap();
-                eprintln!("Linking {dest:?} -> {src:?} ({resolved:?})");
                 unix_fs::symlink(&resolved, &dest).into_diagnostic()?;
             }
             Command::PthEntry { path } => {
--- a/py/tools/py/src/venv.rs
+++ b/py/tools/py/src/venv.rs
@@ -915,30 +915,27 @@ pub fn populate_venv<A: PthEntryHandler>(
                 copy_and_patch_shebang(src, dest)?;
             }
             Command::Symlink { src, dest } => {
-                let mut src = src;
                 fs::create_dir_all(&dest.parent().unwrap()).into_diagnostic()?;
 
-                // On Linux, the sandboxing strategy for actions is to create a
-                // forest of symlinks. If we create a symlink (dest) pointing to
-                // a symlink (src) we're assuming that the src won't be removed
-                // sometime down the line.
+                // The sandboxing strategy for actions is to create a forest of
+                // symlinks. If we create a symlink (dest) pointing to a symlink
+                // (src) we're assuming that the src won't be removed sometime
+                // down the line. But sandboxes are ephemeral, so this leaves us
+                // open to heisenbugs.
                 //
-                // However on Linux the symlink forrest sandboxes are ephemeral
-                // and per-action. Bazel does not appear to rewrite links in the
-                // output TreeArtifact to refer to stable locations for the
-                // input files. But we can do that ourselves by recognizing and
-                // dereferencing the links so that we continue to take links to
-                // the "real" output locations of the files we took as inputs
-                // but which Bazel is presenting to us as links.
+                // What Bazel does gurantee is the _relative tree structure_
+                // betwen our output file(s) and the input(s) used to generate
+                // them. So while we can't sanely just write absolute paths into
+                // symlinks we can write reative paths.
                 //
-                // Ideally we'd generate links to where these files WILL BE in
-                // our runfiles tree, but those links would be dangling at
-                // creation time and Bazel doesn't allow that.
-                while src.is_symlink() && src.to_str().unwrap().contains("sandbox") {
-                    src = src.read_link().into_diagnostic()?;
-                }
-
-                unix_fs::symlink(&src, &dest).into_diagnostic()?;
+                // Note that the relative path we need is the relative path from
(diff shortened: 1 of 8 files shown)


myrrlyn:
I'm having trouble running the tests on my machine; we can talk about it synchronously later. The meat of your question to me, I think, is about dyn-compatibility and how we can make the source-code simpler when branching on different types.

Short answer: Virtualization is mutually exclusive with generics. `fn plan<A: AsRef<Path>, B: AsRef<Path>>(&self, ...)` is not virtualizable _because it has generics_, not because of anything else in the arguments. If you strip the generics out of it, and take `&Path` in place of the type parameters, it virtualizes just fine.

Especially for functions where you only genericize because you're widening your Postel acceptance, you should get in the habit of writing your primary logic with a concretely lowered type (here `&Path`), and either explicitly converting at the call site (`.as_ref()`) or writing a wrapper function that serves as your "true API" with the generics which lowers and then calls the primary logic function.

The reason for this is that the ENTIRE function body of a generic function gets monomorphized for EVERY set of type arguments you use in it. If you have an inner non-generic function, it only exists once (function items are always global to the translation unit, even if they are syntactically nested inside another function; the only thing that does is affect symbol visibility in the privacy system), so the outer function with generics only monomorphizes as a couple function calls.

I'll put up a child PR into this one that illustrates switching the `PthEntryHandler` over to a virtualizable signature, virtualizing it into `Box<dyn PthEntryHandler>` in your use sites in `venv_bin::venv_cmd_handler`, and writing `populate_venv` to take `&dyn PthEntryHandler` instead of `impl PthEntryHandler` in my morning.

I also have style notes on other aspects of the code but I'll save that for after the virtualization.

myrrlyn:
Ok I'm not tired yet. See #648

py/tools/py/src/venv.rs
```diff
@@ -401,91 +466,309 @@ aspect-runfiles-repo = {1}
...
+        // Assumes that `create_empty_venv` has already been called to build out the virtualenv.
+        let dest = &venv.site_dir;
+        let action_src_dir = current_dir().into_diagnostic()?;
+        let action_bin_dir = action_src_dir.join(bin_dir);
+
+        let mut plan: Vec<Command> = Vec::new();
+
+        for prefix in [&action_src_dir, &action_bin_dir] {
+            let src_dir = prefix.join(entry_repo).join(&entry_path);
+            if src_dir.exists() {
```
myrrlyn, line 545: check this out:

```rust
for (src_dir, entry) in [&action_src_dir, &action_bin_dir]
  .iter()
  // conjure PathBufs from each of the provided action-dirs
  .map(|d| d.join(entry_repo).join(entry_path))
  // throw away ones that don't exist in the fs
  .filter(|p| p.exists())
  // turn surviving paths into a pair of (PathBuf, WalkDir) 
  .flat_map(|p| WalkDir::new(p)
    .ok() // Option<WalkDir>
    .map(|w| (p, w)) // Option<(PathBuf, WalkDir>
  ) // Iterator<(PathBuf, WalkDir)>
  // transform each single walkdir into a stream of (PathBuf, DirEnt)
  .flat_map(|(p, w)| w.into_iter().map(move |ent| (p.clone(), ent))) // (PathBuf, DirEnt)
  // and discard directories in the walk
  .filter(|(_, e)| !e.file_type().is_dir())
{
  plan.push(Command::Copy {
    // ...
  });
}
```

"To Me", these are equivalently legible; honestly, this might be _more_ legible. I've also been pickling my brain on this for eight years, so you might disagree. I encourage you to really try to think in terms of iterator combinators, though. Especially for cases like this, where you don't have unhappy-path handling, the rightward drift of for/if blocks can be replaced with vertical straight-line sequences of data transforms. I will happily grant that anything involving the word "flat" (`.flat_map()`, `.flatten()`) is an opportunity for pause and concern -- iterator expansion is really weird -- but the filters and plain maps are fairly straightforward.

arrdem: Slacked you about getting cute here. Can refine later.

PR #647, opened 2025-09-25
chore: use hermetic awk for mtree manipulation

alexeagle:
Testing locally isn't working on MacOS arm64:

```
ERROR: /private/var/tmp/_bazel_alexeagle/ac5c25dfb9cf2a4fcbea5ab9eb6f568e/external/gawk/BUILD.bazel:52:11: Linking external/gawk/libsupport.a [for tool] failed: (Exit 1): llvm-ar failed: error executing CppArchive command (from target @@gawk//:support) external/llvm_toolchain/bin/llvm-ar @bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/gawk/libsupport.a-2.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
external/llvm_toolchain/bin/llvm-ar: error: only one operation may be specified
OVERVIEW: LLVM Archiver
```

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -16,6 +16,7 @@ bazel_dep(name = "aspect_tools_telemetry", version = "0.2.6")
 bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "rules_python", version = "0.29.0")
 bazel_dep(name = "platforms", version = "0.0.7")
+bazel_dep(name = "gawk", version = "5.3.2.bcr.1")  # Required to manipulate mtree files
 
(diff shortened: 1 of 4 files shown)


UebelAndre (approved):
I will take a look at the build issue separately but the changes otherwise seem fine to me 😄

arrdem (changes requested):
Blocking until we can sort out the test issues; not willing to break local dev for the rest of the ruleset on this point.

alexeagle:
I'm not going to find time for this.

I updated https://github.com/alexeagle/awk.bzl/blob/main/README.md with the one useful bit.

PR #648, opened 2025-09-25
virtualize the trait Reid was stuck on

myrrlyn:
Code review per request

### DIFF
--- a/py/tools/venv_bin/src/main.rs
+++ b/py/tools/venv_bin/src/main.rs
@@ -115,76 +115,65 @@ struct VenvArgs {
 
 fn venv_cmd_handler(args: VenvArgs) -> miette::Result<()> {
     let pth_file = py::PthFile::new(&args.pth_file, args.pth_entry_prefix);
-    match args.mode {
+    if let VenvMode::DynamicSymlink = args.mode {
         // FIXME: Does this need to care about the repo?
-        VenvMode::DynamicSymlink => py::create_venv(
+        return py::create_venv(
             &args.python,
             &args.location,
             Some(pth_file),
             args.collision_strategy.unwrap_or_default().into(),
             &args.venv_name,
-        ),
+        );
+    }
 
-        it => {
-            let Some(version) = args.version else {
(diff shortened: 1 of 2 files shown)


py/tools/venv_bin/src/main.rs
```diff
@@ -115,76 +115,65 @@ struct VenvArgs {
 
 fn venv_cmd_handler(args: VenvArgs) -> miette::Result<()> {
     let pth_file = py::PthFile::new(&args.pth_file, args.pth_entry_prefix);
-    match args.mode {
+    if let VenvMode::DynamicSymlink = args.mode {
```
myrrlyn, line 118: By switching from a match with one arm acting and the other arm doing-work-then-matching-again to `if let` with a `return`, I took out one level of indentation with no other code changes. The second match still needs the `unreachable!` in it because rustc doesn't have the ability to statically strip enum variants in cases like these. Annoying but c'est la vie. Beats writing Haskal

py/tools/venv_bin/src/main.rs
```diff
@@ -115,76 +115,65 @@ struct VenvArgs {
...
+        py::venv::PythonVersionInfo::from_str(&version)?,
+        &args.location,
+        args.env_file.as_deref(),
+        args.venv_shim.as_deref(),
+        args.debug,
+        args.include_system_site_packages,
+        args.include_user_site_packages,
+    )?;
+
+    let strat: Box<dyn PthEntryHandler> = match args.mode {
```
myrrlyn, line 147: you have to annotate the virtual type here, because if you don't, the match will type-hint off the first real type it returns. The _first_ arm types as `!`, which isn't real. The second arm is `Box<py::venv::PthStrategy>`, which means all other arms need to type as that concrete instance. `!` types that way, but `Box<py::venv::FirstPartyThirdPartyStrategy>` does not.

You can dodge this by typing the binding site (`let strat: ...`) or by as-casting the box. `Box::new(...) as Box<dyn ...>` is more annoying, imo

PR #650, opened 2025-09-25
feat(pip): Add a clean-sheet pip implementation

arrdem:
[Design doc](https://docs.google.com/document/d/1UFessu8_BwmfMJXDvhyhaASrNMo7_3-gO1lTVASbJbs/edit?tab=t.0#heading=h.r69r7hmny141)

The `rules_python` strategy of implementing pip using installs as repository actions is fundamentally broken. Repository actions should be used only for downloading resources, build actions must be reserved for action time when toolchains and RBE can be employed.

This PR is a clean-sheet approach to implementing `pip` sourced dependencies under Bazel based on offloading much of the work `rules_python` must do in its repository actions to a more complete lockfile similar to what [rules_pycross](https://github.com/jvolkman/rules_pycross/) chose to do and for the same reasons.

This PR is very much a work in progress, but aims to make generalized Python crossbuilds and source builds tractable.

### Test plan

TBD


pip/private/sdist_build/rule.bzl
```diff
@@ -0,0 +1,82 @@
...
+    print(py_toolchain)
+    unpacked_sdist = ctx.actions.declare_directory(
+        "src",
+    )
+
+    archive = ctx.attr.src[DefaultInfo].files.to_list()[0]
+
+    # Extract the archive
+    ctx.actions.run(
+        executable = "tar",
```
dzbarsky, line 14: Grab the tar toolchain?

arrdem: yeah I need to drop in bsdtar here.

pip/private/constraints/abi/macro.bzl
```diff
@@ -0,0 +1,122 @@
...
+            #     name = "{}{}".format(interpreter, major),
+            #     match_all = [
+            #         "//pip/private/constraints/python/interpreter:{}".format(interpreter),
+            #         "//pip/private/constraints/python/major:{}".format(major),
+            #     ]
+            # )
+
+            for minor in MINORS:
+                selects.config_setting_group(
+                    name = "is_{}{}{}".format(interpreter, major, minor),
```
dzbarsky, line 90: generally `%s` is faster than `.format` because it avoids a function call and has simpler (faster) arg conversion. Also if you have only a single substitution it helps even more because you don't even need to construct a tuple. Up to you where to use it, but at least for the functions in the extension that are called repeatedly I would strongly consider it. (Honestly I pretty much always use it unless I have a complex/multiline template with 4+ args and/or using an arg multiple times)

pip/private/whl_install/rule.bzl
```diff
@@ -0,0 +1,78 @@
...
+        ],
+        outputs = [
+            install_dir,
+        ],
+    )
+    
+    return [
+        # FIXME: Need to generate PyInfo here
+        DefaultInfo(
+            files = depset([
```
dzbarsky, line 42: i guess you could create this depset once and reuse it below

dzbarsky (commented):
since you said you love nits...i left a bunch of nits. feel free to apply or disregard as you see fit!

PR #653, opened 2025-09-30
Fix bug in `py_venv` when using non runfiles python interpreter

kriscfoster:
We were hitting the following error when using a system python instead of a runfiles python.

`'--venv-shim=external/aspect_rules_py~~py_tools~rules_py_tools.linux_amd64/shim' /usr/bin/python3 '--pth-file=bazel-out/k8-fastbuild/bin/foo/bar/venv.pth' '--env-file=bazel-out/k8-fastbuild/bin/foo/bar/venv.env' '--bin-dir=bazel-out/k8-fastbuild/bin' '--collision-strategy=error' '--venv-name=.venv' '--mode=static-copy' '--version=3.12')`

`error: unexpected argument '/usr/bin/python3' found`

After debugging, realized that the `--python` argument identifier is being omitted when a non-runfiles python is used. The problem is in the `if` / `else` clause. Validated this works using a patch file.

### Test plan

### DIFF
--- a/py/private/py_venv/py_venv.bzl
+++ b/py/private/py_venv/py_venv.bzl
@@ -122,7 +122,7 @@ def _py_venv_base_impl(ctx):
             # user tries to consume a `py_venv_binary` across repo boundaries
             # which could cause repo mapping to become relevant.
             "--repo=" + (ctx.label.repo_name or ctx.workspace_name),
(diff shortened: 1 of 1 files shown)


arrdem (approved):
Thanks!

PR #576, opened 2025-06-02
fix(py_venv): Activate embedded interpreters

arrdem:
Previously the `bin/activate` script relied on Bazel (specifically `bazel run`) defined envvars, or being sourced directly from the binary entrypoint. This meant that if the user created a link (eg via the venv link verb or manually) to the venv defined by a given target, sourcing that venv's activate script would fail to locate a runfiles-based interpreter. On systems where there is no appropriately versioned non-hermetic interpreter available, this meant that venvs which should run with the Bazel-managed interpreter instead failed.

The fix is to harden venv initialization so that if the `activate` script is directly used AND the runfiles preconditions are not met an effort will be made to satisfy them.

---

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: no
- Breaking change (forces users to change their own code or config): no
- Suggested release notes appear below: no

### Test plan

- [x] Updated script works on BASH
```shellsession
$ /bin/bash
$ bazel run //examples/py_binary:py_binary.venv
INFO: Analyzed target //examples/py_binary:py_binary.venv (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //examples/py_binary:py_binary.venv up-to-date:
  bazel-bin/examples/py_binary/py_binary.venv
INFO: Elapsed time: 1.224s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/examples/py_binary/py_binary.venv
Linking: /private/var/tmp/_bazel_arrdem/93bfea6cdc1153cc29a75400cd38823a/execroot/aspect_rules_py/bazel-out/darwin_arm64-fastbuild/bin/examples/py_binary/.py_binary.venv -> /Users/arrdem/Documents/work/aspect/rules_py/.py_binary.venv

To activate the virtualenv run:
    source /Users/arrdem/Documents/work/aspect/rules_py/.py_binary.venv/bin/activate

Link is up to date!
$ ( source .py_binary.venv/bin/activate; echo $VIRTUAL_ENV; which python; which python3; echo $PYTHONHOME; deactivate; echo $VIRTUAL_ENV; which python; which python3 )
/private/var/tmp/_bazel_arrdem/93bfea6cdc1153cc29a75400cd38823a/execroot/aspect_rules_py/bazel-out/darwin_arm64-fastbuild/bin/examples/py_binary/.py_binary.venv
[...]

PR #669, opened 2025-10-15
tweak(py_venv): Don't modifying shell flags in `activate`

jgsogo:

When creating a virtual environment, the `activate` script no longer modifies bash flags. These flags were added in  https://github.com/aspect-build/rules_py/pull/576 (v1.6.0) (note, they don't appear in the scripts that Python's `venv` module generates).

- Searched for relevant documentation and updated as needed: yes (behaviour is not documented)
- Breaking change (forces users to change their own code or config):no

### Test plan

Potentially, users could have some tests running the following commands. The new flags introduced in https://github.com/aspect-build/rules_py/pull/576 broke this kind of tests:

```sh
#!/usr/bin/env bash
set -euo pipefail

# Create a virtualenvironment
bazel run //:venv

# Activate it (previous it was setting +e flag)
source .venv/bin/activate

# Some failing test -- users expects the script to end and fail here
false

# Some other command that doesn't fail
true

# !!!! Script returns 0 !!!!
```


arrdem:
Looks like GHA is having issues, tests passed I'm just gonna force merge.

PR #681, opened 2025-10-21
fix(release): Remove hardcoded -Clto flag from release builds

gibfahn:
The -Clto flag can conflict with parent module's Rust toolchain
configuration, particularly when embed-bitcode=no is set. This change
removes the hardcoded LTO flag to allow proper inheritance of toolchain
settings.

Related: https://github.com/bazelbuild/rules_rust/pull/3104/


arrdem:
I actually need to revisit exactly this today, but these flags were added for build determinism and to mitigate flapping of the two image layer tests that failed on this PR. Since our releases ship prebuilt Rust artifacts this isn't a detail that's visible unless you're running from source in which case you're already on your own.

PR #641, opened 2025-09-11
Remove determine_main rule

keith:
This inlines this into the executable rules instead.

Fixes https://github.com/aspect-build/rules_py/issues/605
Fixes https://github.com/aspect-build/rules_py/issues/621

PR #702, opened 2025-11-05
chore(semantics): Remove determine_main

### LINKED ISSUE #605: [Bug]: `py_test` / `py_binary` incorrectly handle an `exec_properties` attribute that is explicitly set to `None`
### What happened?

If `exec_properties` for a `py_binary` or `py_test` is explicitly set to `None`, an error of the form

```
Error: 'NoneType' value has no field or method 'items'
```

is raised on this line:

https://github.com/aspect-build/rules_py/blob/5968bcbde0cc7224b104bea3255a97bda29c89f1/py/defs.bzl#L67

In our case, the explicit value of `None` came from a symbolic macro that used `inherit_attrs = "common"`. This regression was found when trying upgrade from version 1.3.2 to version 1.6.0. Our workaround is to use

```starlark
if kwargs.get("exec_properties", None) == None:
        kwargs["exec_properties"] = {}
```

in symbolic macro implementations that create `py_*` targets.

This regression appears to have been introduced in https://github.com/aspect-build/rules_py/pull/529; CC @keith.

### Version

Development (host) and target OS/architectures: irrelevant

Output of `bazel --version`: bazel 8.1.1

`WORKSPACE` or `MODULE.bazel` file: 1.6.0

Language(s) and/or frameworks involved: irrelevant

### How to reproduce

```starlark
py_binary(
    name = "foo",
    srcs = ["foo.py"],
    exec_properties = None,
)
```

### Any other information?

_No response_

### LINKED ISSUE #621: [Bug]: exec_properties with select() statements fails with "'select' value has no field or method 'items'"
### What happened?

When using `exec_properties` with Bazel `select()` statements in `py_test` rules, the build fails with:
`Error: 'select' value has no field or method 'items'`

### Version

- rules_py version: v1.4.0+ (all versions since the fix)
- Bazel version: Any
- OS: Any

### How to reproduce

```shell
Minimal reproduction:

py_test(
    name = "my_test",
    srcs = ["test.py"],
    exec_properties = select({
        "//conditions:default": {"key": "value"},
    }),
)
```

### Any other information?

I believe this was introduced via commit 84f17d3 (PR #529) which was first released in v1.4.0

In https://github.com/aspect-build/rules_py/blob/main/py/defs.bzl#L67, the code assumes exec_properties is always a dict:
```python
non_test_exec_properties = {k: v for k, v in exec_properties.items() if not k.startswith("test.")}
```

arrdem:
Vendor `rules_python`'s `determine_main` and remove the need for a separate `determine_main` target which complicates exec properties.

Replaces #641 with thanks to Keith.
Fixes #605
Fixes #621

Note that `py_venv_*` already mandates that a `main=` label be provided, so this is for legacy `py_binary` only. Should be removed in 2.0.

### Test plan

### DIFF
--- a/py/defs.bzl
+++ b/py/defs.bzl
@@ -35,7 +35,6 @@ python.toolchain(python_version = "3.9", is_default = True)
 ```
 """
 
-load("@aspect_bazel_lib//lib:utils.bzl", "propagate_common_rule_attributes")
 load("//py/private:py_binary.bzl", _py_binary = "py_binary", _py_test = "py_test")
 load("//py/private:py_image_layer.bzl", _py_image_layer = "py_image_layer")
 load("//py/private:py_library.bzl", _py_library = "py_library")
--- a/py/private/py_binary.bzl
+++ b/py/private/py_binary.bzl
@@ -148,13 +148,13 @@ _attrs = dict({
 Script to execute with the Python interpreter.
 
 Must be a label pointing to a `.py` source file.
-If such a lable is provided, it will be honored.
+If such a label is provided, it will be honored.
 
 If no label is provided AND there is only one `srcs` file, that `srcs` file will be used.
 
 If there are more than one `srcs`, a file matching `{name}.py` is searched for.
 This is for historical compatibility with the Bazel native `py_binary` and `rules_python`.
-Relying on this behaivor is STRONGLY discouraged, may produce warnings and may
+Relying on this behavior is STRONGLY discouraged, may produce warnings and may
 be deprecated in the future.
 
 """,
--- a/py/private/py_semantics.bzl
+++ b/py/private/py_semantics.bzl
@@ -85,7 +85,6 @@ def _resolve_toolchain(ctx):
         flags = _INTERPRETER_FLAGS,
     )
 
-
 def _csv(values):
     """Convert a list of strings to comma separated value string."""
     return ", ".join(sorted(values))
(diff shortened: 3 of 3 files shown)


py/defs.bzl
```diff
@@ -126,6 +112,11 @@ def py_binary(name, srcs = [], main = None, **kwargs):
     if resolutions:
         resolutions = resolutions.to_label_keyed_dict()
 
+    if not main and len(srcs) == 1:
+        main = srcs[0]
+    elif main and main not in srcs:
```
dzbarsky: Perfect

dzbarsky (approved):

py/defs.bzl
```diff
@@ -104,7 +91,16 @@ def _py_binary_or_test(name, rule, srcs, main, data = [], deps = [], resolutions
...
 
-def py_binary(name, srcs = [], main = None, **kwargs):
+def _normalize_label(src):
+    if src.startswith("//"):
+        return Label(src)
+    else:
+        return Label("//%s:%s" % (native.package_name(), src.lstrip(":")))
+
+def _normalize_srcs(srcs):
+    return [_normalize_label(it) for it in srcs]
```
arrdem: @dzbarsky is there a better way here? The motivation is that if the user gives us `:main.py`, we can't immediately tell without normalization that `main.py` is already on the list. Applies to labels generally; consider absolute label to main.

Issue #718, opened 2025-11-19
[Bug]: UV locks may contain multiple versions of single artifacts


siscia:
### What happened?

```
% bazel query '@pypi//...'
ERROR: /private/var/tmp/_bazel_simone/d5ca606f12ab6c659b9a7bb351339261/external/aspect_rules_py+/uv/private/extension.bzl:553:28: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_simone/d5ca606f12ab6c659b9a7bb351339261/external/aspect_rules_py+/uv/private/extension.bzl", line 899, column 18, in _uv_impl
		_sbuild_repos(module_ctx, lock_specs, annotation_specs, override_specs)
	File "/private/var/tmp/_bazel_simone/d5ca606f12ab6c659b9a7bb351339261/external/aspect_rules_py+/uv/private/extension.bzl", line 553, column 28, in _sbuild_repos
		sdist_build(
Error in repository_rule: A repo named sbuild__pypi__default__aiohttp is already generated by this module extension at /private/var/tmp/_bazel_simone/d5ca606f12ab6c659b9a7bb351339261/external/aspect_rules_py+/uv/private/extension.bzl:553:28
ERROR: Target parsing failed due to unexpected exception: error evaluating module extension @@aspect_rules_py+//uv/unstable:extension.bzl%uv
```

I [follow the instructions](https://github.com/aspect-build/rules_py/blob/main/docs/uv.md) and added to our MODULE.bazel

```
...

uv = use_extension("@aspect_rules_py//uv/unstable:extension.bzl", "uv")
uv.declare_hub(
    hub_name = "pypi",      # Or whatever you wish
)
uv.declare_venv(
    hub_name = "pypi",      # Must be a declared hub
    venv_name = "default",  # Or whatever you wish
)
uv.lockfile(
    hub_name = "pypi",      # Must be a declared hub
    venv_name = "default",  # Must be a declared venv
    src = "//:uv.lock",
)
use_repo(uv, "pypi")
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:

% bazel --version
bazel 8.4.2

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

```
# Aspect Rules Py - Enhanced Python rules
bazel_dep(name = "aspect_rules_py", version = "1.7.0")

# Override to use git repository since not yet in BCR
git_override(
    module_name = "aspect_rules_py",
[...]

PR #720, opened 2025-11-19
fix multiple versions same name - #718

siscia:

Address #718 

### Test plan

- New test cases added


arrdem:
Closing; the fix isn't this simple.

PR #747, opened 2025-12-09
fix: unblock bcr

thesayyn:
Removes .bazelrc from e2e as it contains flags that does not work with some versions of 7.x


arrdem:
Looking.

PR #750, opened 2025-12-11
test: add test for failing use case of uv locks

lstanden:
This PR is not ready for merge. It simply creates a new test case for an issue encountered with using the UV lock files. The python library being used is older and unmaintained, and an error is generated when trying to run tests:

Error in fail: Error in lockfile @@//cases/legacy-deps/googlemaps:uv.lock

The requirements `build` and `setuptools` are missing from, but the following requirements only provide sdists.
Please update your lockfile to provide build tools in order to enable sdist support.

Problems:
 - googlemaps

### Test plan

- New test cases added


lstanden:
Update on findings:
- The original error I had was fixed with a `bazel clean`
- The new error which is in CI i have now run into locally as well

PR #751, opened 2025-12-11
fix: add tar.bzl as dep since its needed sdist_build

thesayyn:


tellett (approved):
LGTM

.bcr/presubmit.yml
```diff
@@ -1,7 +1,7 @@
 bcr_test_module:
   module_path: "e2e"
   matrix:
-    bazel: ["8.x", "7.x"]
+    bazel: ["8.x"]
```
alexeagle: is it broken on bazel 7 by this change?

thesayyn: no it was broken before. i am just here to have a green entry on BCR

thesayyn: CI is only testing with 8.x as well.

PR #755, opened 2025-12-16
Remove some usage of aspect_bazel_lib

dzbarsky:
We shouldn't pull in both bazel-libs.

### Test plan


arrdem:
Yep, this and ripping out the WORKSPACE file is the basis of the 2.0 I want to cut.

PR #798, opened 2026-02-13
fix(uv): invalidate extension state when lock/annotation files change

### LINKED ISSUE #797: [Bug]: uv extension: uv.lock changes can leave hub repos stale in normal lockfile mode
### Summary
When using `@aspect_rules_py//uv/unstable:extension.bzl` with Bzlmod, updating `uv.lock` can leave generated hub repos stale in normal lockfile mode (`update`/`refresh`).

In our case, `uv.lock` contains `hypothesis`, but `@hub//hypothesis` is missing unless we run with `--lockfile_mode=off`.

### Environment
- Bazel: `9.0.0`
- Bazelisk: `v1.28.1`
- `aspect_rules_py`: `1.8.4`
- `rules_python`: `1.8.4`
- uv: `0.10.0`
- extension usage: `use_extension("@aspect_rules_py//uv/unstable:extension.bzl", "uv")`

### Repro
1. Configure uv extension in `MODULE.bazel` with one hub+venv and lockfile:
   - `declare_hub(hub_name = "hub")`
   - `declare_venv(hub_name = "hub", venv_name = "default")`
   - `lockfile(src = "//:uv.lock", hub_name = "hub", venv_name = "default")`
2. Set up dependencies in pyproject.toml and uv.lock, and lock them in MODULE.bazel.lock. Do a build or a test that causes bazel to cache everything.
3.  Add any new dependency. For example, add hypothesis to pyproject.toml's dev dependencies and re-lock uv.
4. Run:
   ```bash
   bazelisk mod deps --lockfile_mode=refresh
   bazelisk query @hub//hypothesis:all
   ```

### Actual
`bazelisk query @hub//hypothesis:all` fails with:
`no such package '@@aspect_rules_py++uv+hub//hypothesis'`

And `bazelisk test //...` fails during analysis for the same reason.

### Expected
If `uv.lock` contains `hypothesis`, then `@hub//hypothesis:*` should exist in normal lockfile mode.

### Important observation
With lockfile mode disabled, it works immediately:

```bash
bazelisk query --lockfile_mode=off @hub//hypothesis:all
```

This returns targets (`hypothesis`, `lib`, `whl`) correctly.

So extension repos appear stale under normal lockfile mode, but recompute correctly when lockfile caching is bypassed.

### Additional investigation
`uv.lock` parsing itself appears correct:
- The `toml2json` binary used by the extension decodes our `uv.lock` with both `hypothesis` and `sortedcontainers` present.

Potential invalidation issue is in:
- `uv/private/tomltool/toml.bzl` (`_decode_file`), which parses via `ctx.execute([... content_path ...])`.

It looks like `content_path` may not be registered as a tracked input for extension invalidation, so changes to `uv.lock` can be missed by cached module-extension state.

If that diagnosis is right, explicitly reading/tracking lockfile content before parsing (so Bazel keys include lockfile content) would likely fix this.

### Workaround
Use `--lockfile_mode=off` for commands that need fresh extension evaluation, but that is not viable as normal workflow.

mkanat:
## Summary
Fixes #797 by explicitly watching uv lock/annotation source files in the module extension so file content changes invalidate cached extension state in normal lockfile modes.

## Reproduction
Reproduced from a current head checkout using a minimal bzlmod workspace that depends on published `aspect_rules_py@1.8.4`:
1. Run `bazelisk mod deps --lockfile_mode=refresh` and query an existing dep (`@hub//cowsay:all`).
2. Update `uv.lock` to add a new dep (`click`) and rerun `bazelisk mod deps --lockfile_mode=refresh`.
3. Query `@hub//click:all`.

Actual on `1.8.4`: missing package error (`@@aspect_rules_py++uv+hub//click`).

## Change
- Add `module_ctx.watch(lock.src)` before lockfile decode.
- Add `module_ctx.watch(ann.src)` before annotations decode.
- Include comments describing why these watches are required.

## Verification
- Reproduced failure from current head checkout against published `1.8.4`.
- Verified this branch resolves `@hub//click:all` correctly under `--lockfile_mode=refresh`.

### DIFF
--- a/uv/private/extension.bzl
+++ b/uv/private/extension.bzl
@@ -176,6 +176,9 @@ def _parse_locks(module_ctx, venv_specs):
             if lock.venv_name in lock_specs[lock.hub_name]:
                 fail("Multiple lockfiles detected for hub %s venv %s!" % (lock.hub_name, lock.venv_name))
 
+            # Keep module extension state keyed on lockfile content in normal
+            # lockfile modes so updated dependencies invalidate stale hub repos.
+            module_ctx.watch(lock.src)
             lockfile = toml.decode_file(module_ctx, lock.src)
             if lockfile.get("version") != 1:
                 fail("Lockfile %s is an unsupported format version!" % lock.src)
@@ -298,6 +301,8 @@ def _parse_annotations(module_ctx, hub_specs, venv_specs):
                 default_build_deps = [] + _default_annotations.default_build_deps,
             ))
 
+            # Track annotation file changes for module extension invalidation.
+            module_ctx.watch(ann.src)
             ann_content = toml.decode_file(module_ctx, ann.src)
             if ann_content.get("version") != "0.0.0":
                 fail("Annotations file %s doesn't specify a valid version= key" % ann.src)


mkanat:
One thing I don't know about this is whether this is the right place in the codebase to do this, since I'm not intimately familiar with the codebase.

PR #801, opened 2026-02-21
WIP: start proto/gRPC support

alexeagle:
/cc @acozzette


alexeagle:
Not particularly close to working yet. I'm using the new uv machinery and I think it doesn't produce a package that works with `py_console_script_binary` so I'm not sure how to execute the grpcio_tools package which contains the grpc codegen.

PR #806, opened 2026-02-27
Fix python interpreter lookup when space in path exists

Wito-1:
Currently there's an edge-case where if you deploy your python application to a path that contains a space, the python interpreter lookup only provides the path up until the space. 

This PR fixes that by surrounding what is returned with `$(rlocation <path>)` in quotes to pass the full location to `alocation`

---

### Test plan

- New test cases added


arrdem (approved):

arrdem:
Looks reasonable to me thanks for the fix!

PR #890, opened 2026-03-20
fix(uv): resolve unconstrained multi-version packages via lockfile dev-dependencies

arrdem:
When conflicting dependency groups both include a package without a version
specifier (e.g. just `"build"`), and the lockfile contains multiple versions,
resolution failed because `normalize_deps` only creates defaults for
single-version packages and specifier-based matching requires a non-empty
specifier.

The fix reads per-group resolved versions from the lockfile's root project
`dev-dependencies`/`dependency-groups` section and overlays them onto the
default version map before resolving each group.

### Test plan

- New e2e test case `uv-unconstrained-conflict-890` with intentionally stale lockfile
  reproducing the scenario (unconstrained `build` resolving to different versions
  per group)
- `test_unconstrained` asserts the pyproject.toml keeps `build` unpinned
- Existing conflict e2e tests pass
- Main test suite passes


uv/private/extension/projectfile.bzl
```diff
@@ -142,6 +142,25 @@ def collect_activated_extras(projectfile, lock_id, project_data, lock_data, defa
...
+    # Find the root project in lock_data to extract per-group resolved versions.
+    # The lockfile's dev-dependencies/dependency-groups section records the exact
+    # version the resolver picked for each group, which lets us resolve
+    # unconstrained multi-version packages (e.g. "build" with no specifier).
+    root_project_name = normalize_name(project_data["project"]["name"])
+    lock_group_versions = {}
+    for spec in lock_data.get("package", []):
+        if normalize_name(spec["name"]) == root_project_name:
+            # Handle both old ("dev-dependencies") and new ("dependency-groups") key names
+            group_deps_map = spec.get("dev-dependencies", spec.get("dependency-groups", {}))
```
jbedard: dev has priority over groups? Is that intentional? Should it be tested?

jbedard (approved):

PR #912, opened 2026-04-05
refactor(wheels): propagate wheels via PyWheelInfo instead of DefaultInfo

jbedard:
This was a robots work to "stop propagating all the wheels via DefaultInfo" which @dzbarsky mentioned on slack.

--

Previously, wheel files leaked into DefaultInfo.files of every target that transitively depended on a pip package. This happened because whl_requirements used whl_mode_transition to flip lib_mode=whl across the dep graph, causing install aliases to resolve to whl_lib (a py_library with the raw .whl file as srcs). Since py_library propagates DefaultInfo.files transitively, every intermediate target accumulated all transitive .whl files as visible build outputs.

The fix introduces PyWheelInfo, a dedicated provider that carries wheel files without putting them in DefaultInfo:

- whl_install now returns PyWheelInfo(files=depset([archive])) alongside its existing DefaultInfo and PyInfo
- py_library propagates PyWheelInfo transitively from its deps, so the full set of transitive wheels is reachable from any node in the dep graph
- whl_requirements reads PyWheelInfo.files directly, with no cfg transition

Because whl_requirements no longer needs whl_mode_transition to surface wheels, the entire lib_mode machinery becomes unnecessary:

- whl_mode_transition was only used by whl_requirements (now gone) and py_whl_library (which only existed to force whl mode)
- lib_mode_transition on pep517_whl existed solely to counteract whl_mode_transition in source-build dep chains; without whl_mode_transition there is nothing to counteract
- whl_lib in each package repo and the libs_are_whls arm of the install select are removed; install always resolves to actual_install
- py_whl_library in the hub becomes a plain alias; PyWheelInfo is available on the same target without any mode transition

libs_are_libs and libs_are_whls config_settings are retained as no-op stubs for API compatibility.

---

### Test plan

- New test cases added

### DIFF
--- a/py/private/py_library.bzl
+++ b/py/private/py_library.bzl
@@ -8,7 +8,7 @@ load("@bazel_skylib//lib:new_sets.bzl", "sets")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
 load("@rules_python//python:defs.bzl", "PyInfo")
-load("//py/private:providers.bzl", "PyVirtualInfo")
+load("//py/private:providers.bzl", "PyVirtualInfo", "PyWheelInfo")
 
 def _make_instrumented_files_info(ctx, extra_source_attributes = [], extra_dependency_attributes = []):
     return coverage_common.instrumented_files_info(
@@ -17,6 +17,13 @@ def _make_instrumented_files_info(ctx, extra_source_attributes = [], extra_depen
         dependency_attributes = ["data", "deps"] + extra_dependency_attributes,
         extensions = ["py"],
     )
+
+def _make_wheel_depset(ctx):
+    return depset(transitive = [
+        target[PyWheelInfo].files
+        for target in ctx.attr.deps
+        if PyWheelInfo in target
+    ])
 
 def _make_srcs_depset(ctx):
     return depset(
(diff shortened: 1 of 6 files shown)


jbedard:
My questions...

After https://github.com/aspect-build/rules_py/pull/907 does this still have an effect?

....

PR #907 is complementary but addresses a different artifact: it removes install_dir (the unpacked wheel directory) from whl_install's DefaultInfo.files. Our change removes raw .whl files from DefaultInfo by eliminating whl_lib.

  However, PR #907 alone doesn't fully solve either problem. install_dir was removed from whl_install's own DefaultInfo.files, but py_library still propagated it further — because py_library.DefaultInfo.files included transitive =     
  [transitive_srcs], where transitive_srcs collected PyInfo.transitive_sources from all deps (which includes install_dir). That's the separate change already on this branch that drops the transitive from py_library's DefaultInfo.
                                                                                                                                                                                                                                           
  So after PR #907 merges, our change still has positive effects:                                                                                                                                                                          
  
  1. Removes whl_lib — the only remaining source of raw .whl files in DefaultInfo.files, which PR #907 doesn't touch                                                                                                                       
  2. Removes the transition machinery — whl_mode_transition, lib_mode_transition, py_whl_library — none of which PR #907 touches                                                                                                         
  3. Fixes whl_requirements for transitive collection — with py_library no longer propagating transitive files through DefaultInfo, the old s.files-based whl_requirements would silently collect nothing from SCC py_library targets;     
  PyWheelInfo is what makes transitive collection correct                                                                                                                                                                                  
                                                                                                                                                                                                                                           
  The two changes together clean up both sides: PR #907 removes install dirs, our change removes wheel archives and the mode-switching machinery that existed to surface them.

jbedard:
This was instead deleted https://github.com/aspect-build/rules_py/pull/951

PR #921, opened 2026-04-06
runfiles: support Bazel 9 compact _repo_mapping format

gregmagolan:
## Problem

Bazel 9 introduces [`--incompatible_compact_repo_mapping_manifest`](https://github.com/bazelbuild/bazel/issues/26262) which rewrites `_repo_mapping` entries using a prefix-wildcard format to reduce file size from quadratic to linear in repo count. A source-repo field ending with `*` matches any canonical repo name that starts with that prefix.

The existing CSV parser only handled the old 3-column exact-match format. Any user on Bazel 9 with compact manifests enabled would get silently wrong runfile lookups for repos whose names match a compact prefix.

## Solution

- `RepoMapping` is promoted from a plain `HashMap` type alias to a struct holding separate `exact` and `prefixed` tables.
- `parse_repo_mapping` detects compact entries via the trailing `*` (unambiguous: `*` is illegal in canonical repo names) and routes them to the prefix table.
- `RepoMapping::get` tries exact match first, then iterates prefix entries in file order (first match wins).
- New test `test_parse_repo_mapping_compact_format` covers mixed exact+prefix files, prefix matching, exact-over-prefix priority, and no-match cases.

## References

- [bazelbuild/bazel#26262](https://github.com/bazelbuild/bazel/issues/26262) — `--incompatible_compact_repo_mapping_manifest` tracking issue; contains the full format spec and the list of runfiles libraries that need updating
- [Bazel 9.0.0 release notes](https://github.com/bazelbuild/bazel/releases/tag/9.0.0) — compact format is enabled by default in Bazel 9

## Test plan
- [ ] `bazel test //py/tools/runfiles/...` passes on CI
- [ ] Existing `test_parse_repo_mapping` and `test_parse_repo_mapping_invalid_file` still pass
- [ ] New `test_parse_repo_mapping_compact_format` passes

### DIFF
--- a/py/tools/runfiles/src/lib.rs
+++ b/py/tools/runfiles/src/lib.rs
@@ -140,8 +140,54 @@ enum Mode {
     ManifestBased(HashMap<PathBuf, PathBuf>),
 }
 
-type RepoMappingKey = (String, String);
-type RepoMapping = HashMap<RepoMappingKey, String>;
+/// Repo mapping supporting both the legacy 3-column CSV format and the compact
+/// prefix-wildcard format introduced in Bazel 9
+/// (`--incompatible_compact_repo_mapping_manifest`).
+///
+/// In the compact format, a source repo field ending with `*` is a prefix that
+/// matches any canonical repo name starting with that prefix. Exact entries take
+/// priority over prefix entries; among prefix entries the first match wins.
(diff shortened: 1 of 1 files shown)


jbedard:
py/tools/runfiles/src/lib.rs no longer exists

PR #930, opened 2026-04-08
fix(py_pex_binary): exclude hermetic python toolchain files to prevent AmbiguousDistributionError

xangcastle:
## Problem

When building a `py_pex_binary` from a `py_binary` that uses a hermetic Python toolchain (e.g., from rules_python), the build fails with:
```
pex.dist_metadata.AmbiguousDistributionError: Found more than one distribution inside external/python3_9_x86_64-unknown-linux-gnu/lib/python3.9/site-packages/:
pip-23.2.1.dist-info/METADATA
setuptools-68.2.2.dist-info/METADATA
```

This happens because the hermetic Python interpreter includes pip and setuptools in its `site-packages`, and `py_pex_binary` was incorrectly including these as dependencies. When `Distribution.load()` is called on the `site-packages` directory, it finds multiple `.dist-info` directories and raises an error.

## Solution

Exclude hermetic Python toolchain files from being processed as PEX dependencies:

1. **Added exclusion pattern** for the hermetic interpreter repository:
   - Pattern: `python_interpreters+` (matches repos like `aspect_rules_py++python_interpreters+python_3_11_aarch64_apple_darwin`)

2. **Added specific file exclusions** for internal toolchain files:
   - `_distutils_hack` - Part of the interpreter's distutils shim
   - `distutils-precedence.pth` - Interpreter configuration file

## Changes

- `py/private/py_pex_binary.bzl`: Added exclusion patterns and specific file filters
- `py/tests/py-pex-binary/BUILD.bazel`: Updated test to use hermetic Python 3.11 toolchain to validate the fix

## Testing

bazel test //py/tests/py-pex-binary:test__print_modules_pex
//py/tests/py-pex-binary:test__print_modules_pex  PASSED

### DIFF
--- a/py/private/py_pex_binary.bzl
+++ b/py/private/py_pex_binary.bzl
@@ -38,6 +38,8 @@ exclude_paths = [
     # these will match in bzlmod setup with --incompatible_use_plus_in_repo_names flag flipped.
     "rules_python++python+",
     "aspect_rules_py+/py/tools/",
+    # these will match the hermetic python interpreter from rules_py extension
+    "python_interpreters~",
     "python_interpreters+",
 ]
 
--- a/py/private/py_pex_binary.bzl
+++ b/py/private/py_pex_binary.bzl
@@ -38,6 +38,7 @@ exclude_paths = [
     # these will match in bzlmod setup with --incompatible_use_plus_in_repo_names flag flipped.
     "rules_python++python+",
     "aspect_rules_py+/py/tools/",
+    "python_interpreters+",
 ]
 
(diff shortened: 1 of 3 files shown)


py/private/py_pex_binary.bzl
```diff
@@ -54,6 +55,11 @@ def _map_srcs(f, workspace):
         if dest_path.find(exclude) != -1:
             return []
 
+    if "_distutils_hack" in f.path:
```
jbedard: What is this?

PR #935, opened 2026-04-14
deterministically resolve unconstrained dependencies when lockfile contains multiple versions

xangcastle:
When a `pyproject.toml` declares a dependency without a version specifier (e.g. `build`) and the `uv.lock` contains multiple versions of that package—commonly caused by `tool.uv.conflicts` or multiple dependency groups—Starlark analysis fails with:

```
Unable to resolve a default version for requirement ...
```

This happens because `extract_requirement_marker_pairs` in `uv/private/extension/projectfile.bzl` only falls back to `find_matching_version` when an explicit version specifier is present. If the specifier is empty, the fallback is skipped and the build aborts.

### Fix
Modified the *Look up version* block in `extract_requirement_marker_pairs` so that, when no default version exists and the requirement has no version constraint, we pass `">=0"` to `find_matching_version`. This deterministically selects the highest available version from the lockfile candidates instead of failing.

### Changes
- **`uv/private/extension/projectfile.bzl`**  
  - Relaxed the fallback condition from `if specifier and pkg_vers:` to `if pkg_vers:`.  
  - Default empty specifiers to `match_spec = ">=0"` before calling `find_matching_version`.  
  - Preserved the final `fail()` for packages that truly do not exist in the lockfile.

- **`uv/private/extension/test_projectfile.bzl`** *(new)*  
  Added unit tests covering:
  - Multi-version package without a specifier (selects latest).
  - Multi-version package with a specifier (honors the constraint).
  - Single-version lookup via `version_map` (no regression).
  - Extras and PEP 508 markers propagation.

- **`uv/private/extension/BUILD.bazel`**  
  Registered the new `projectfile_test_suite()`.

### Validation
```
bazel test //uv/private/extension:all   # 12/12 passed
bazel test //uv/private/versions:all     # 45/45 passed
```

### DIFF
--- a/uv/private/extension/projectfile.bzl
+++ b/uv/private/extension/projectfile.bzl
@@ -86,12 +86,13 @@ def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_m
         # specifier against all known versions of this package in the lockfile.
         specifier = remainder.strip()
         pkg_vers = package_versions.get(pkg_name, {})
-        if specifier and pkg_vers:
+        if pkg_vers:
+            match_spec = specifier if specifier else ">=0"
             candidates = {
                 ver: (lock_id, pkg_name, ver, "__base__")
                 for ver in pkg_vers.keys()
             }
-            v = find_matching_version(specifier, candidates)
+            v = find_matching_version(match_spec, candidates)
     if v == None:
         fail("Unable to resolve a default version for requirement {} in {}".format(repr(req_string), projectfile))
     else:
--- a/uv/private/extension/BUILD.bazel
+++ b/uv/private/extension/BUILD.bazel
@@ -1,6 +1,7 @@
 # TOOLCHAIN_DEPS
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load(":test_graph_utils.bzl", "graph_utils_test_suite")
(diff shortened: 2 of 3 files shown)


uv/private/extension/projectfile.bzl
```diff
@@ -86,12 +86,13 @@ def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_m
         # specifier against all known versions of this package in the lockfile.
         specifier = remainder.strip()
         pkg_vers = package_versions.get(pkg_name, {})
-        if specifier and pkg_vers:
+        if pkg_vers:
+            match_spec = specifier if specifier else ">=0"
```
jbedard, line 90: Does `>=0` handle everything? Is there support for `*` or something more generic?

Is it possible for odd versions like `13fbc94cd8b8870db289d5d9a2381d1b43a5275f` or `foobar` or `-1`?

jbedard (approved):

uv/private/extension/projectfile.bzl
xangcastle: @jbedard the resolution is deterministic within the hermetic scope of the lockfile. Per our key users request to minimize 'warning spew' by default, I'll gate these logs behind a `--verbose` flag in a follow-up PR.

PR #937, opened 2026-04-15
fix(uv): resolve per-group dependency versions from lockfile to prevent configuration conflicts

xangcastle:
When a `pyproject.toml` declares dependency groups with overlapping packages at different versions (commonly via `tool.uv.conflicts`), and some requirements lack explicit version specifiers (e.g. `"build"`), Starlark analysis can fail with:

```
Configuration conflict! Package packaging specifies two or more default package states!
```

### Root Cause
`extract_requirement_marker_pairs` was falling back to `find_matching_version(">=0", ...)` for unconstrained requirements, which always picks the *highest* version across the entire lockfile. This meant both dependency groups could end up resolving the same high version (e.g. `build==1.4.3`), pulling in transitive dependencies (e.g. `packaging==24.0`) that conflicted with the group's explicit pins (e.g. `packaging==21.3`).

### Fix
Introduce `_extract_lockfile_group_versions` in `uv/private/extension/projectfile.bzl` to read the exact versions `uv` already selected for each dependency group from the lockfile's `[package.dev-dependencies]` section. `collect_activated_extras` now seeds `group_preferences` with these lockfile-resolved versions before parsing requirements, ensuring unconstrained dependencies resolve to the correct per-group version without guessing.

### Verification
- Added `collect_activated_extras_transitive_remap_test` in `test_projectfile.bzl` to assert correct version resolution across conflicting groups.
- Fixed the `e2e/cases/unconstrained-dependencies` test suite so tests actually execute (added missing `pytest` dependency, regenerated `uv.lock`, fixed test imports, and added missing `print_python_package_version.py`).
- **Unit tests:** `15/15 PASSED`
- **E2E tests:** `2/2 PASSED` — confirming `packaging==24.0` in `depctx_py_main_uv` and `packaging==21.3` in `depctx_py_humble_uv` with no conflicts.

### DIFF
--- a/uv/private/extension/projectfile.bzl
+++ b/uv/private/extension/projectfile.bzl
@@ -6,7 +6,7 @@ load("//uv/private:normalize_name.bzl", "normalize_name")
 load("//uv/private/versions:versions.bzl", "find_matching_version")
 load(":dep_groups.bzl", "resolve_dependency_group_specs")
 
-def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_map, package_versions = {}):
+def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_map, package_versions = {}, preferred_versions = {}):
     """Parses a requirement string into a list of dependency-marker pairs.
 
     This function parses a PEP 508 requirement string (e.g.,
@@ -80,7 +80,9 @@ def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_m
             remainder = remainder[close_idx + 1:]
 
     # 4. Look up version
-    v = version_map.get(pkg_name)
+    v = preferred_versions.get(pkg_name)
+    if v == None:
+        v = version_map.get(pkg_name)
     if v == None:
         # For multi-version packages (e.g. conflicts), match the version
         # specifier against all known versions of this package in the lockfile.
@@ -152,8 +154,15 @@ def collect_activated_extras(projectfile, lock_id, project_data, lock_data, defa
     for group_name in dep_groups.keys():
         normalized_dep_groups[group_name] = []
         resolved_specs = resolve_dependency_group_specs(dep_groups, group_name)
+
+        group_preferences = {}
         for spec in resolved_specs:
-            for dep, marker in extract_requirement_marker_pairs(projectfile, lock_id, spec, default_versions, package_versions):
+            for dep, _marker in extract_requirement_marker_pairs(projectfile, lock_id, spec, default_versions, package_versions):
+                base = (dep[0], dep[1], dep[2], "__base__")
+                group_preferences[dep[1]] = base
+
+        for spec in resolved_specs:
+            for dep, marker in extract_requirement_marker_pairs(projectfile, lock_id, spec, default_versions, package_versions, group_preferences):
                 normalized_dep_groups[group_name].append(dep)
 
                 # Note that this is the base case for the reach set walk below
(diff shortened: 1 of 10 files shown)


mteggia-bdai:
This change solved the problem I was having with my dependencies, thanks!

jbedard:
@codex

uv/private/extension/projectfile.bzl
```diff
@@ -174,15 +214,19 @@ def collect_activated_extras(projectfile, lock_id, project_data, lock_data, defa
             it = worklist[idx]
             visited[it] = 1
 
-            for next, markers in graph.get(it, {}).items():
-                # Convert `next`, being a dependency potentially with marker, to its base package
```
jbedard, line 178: The comments were deleted?

uv/private/extension/projectfile.bzl
```diff
@@ -113,6 +115,33 @@ def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_m
...
+        lock_id: The lockfile identifier used in dependency tuples.
+        lock_data: The parsed content of the `uv.lock` file.
+
+    Returns:
+        A dictionary mapping normalized group names to dictionaries of
+        {package_name: (lock_id, package_name, version, "__base__")}.
+    """
+    result = {}
+    for pkg in lock_data.get("package", []):
+        if "virtual" not in pkg.get("source", {}):
```
jbedard, line 135: What does this mean? Why would packages with `"virtual"` sources not have group versions?

uv/private/extension/projectfile.bzl
```diff
@@ -113,6 +115,33 @@ def extract_requirement_marker_pairs(projectfile, lock_id, req_string, version_m
...
+            continue
+        for raw_group_name, deps in pkg.get("dev-dependencies", {}).items():
+            group_name = normalize_name(raw_group_name)
+            for dep in deps:
+                pkg_name = normalize_name(dep["name"])
+                if "version" in dep:
+                    result.setdefault(group_name, {})[pkg_name] = (lock_id, pkg_name, dep["version"], "__base__")
```
jbedard, line 142: Is it possible we're overwriting `[pkg_name]` here?

uv/private/extension/projectfile.bzl
xangcastle: the comment is pretty obvious

jbedard: No it's not lol

uv/private/extension/projectfile.bzl
gregmagolan: Ohh. I did not know about setdefault. Neat.

gregmagolan (approved):
Looks good to me

uv/private/extension/projectfile.bzl
xangcastle: In this case we are talking about `source = { virtual = "." }` in the `uv.lock`, **not virtual dependencies**. Please see [this piece](https://github.com/aspect-build/rules_py/blob/c9ad57d7349a191021c89bbad555ec7a128d3a3f/uv/private/extension/defs.bzl#L319-L322)
so in this case is the source code of the project, not an external dependency, that why doesn't have wheel, sdist or version. So we are filtering this bc it's the root project, uv just includes it in the lock so it knows which groups belong to it, but there's no actual package artifact to download.

uv/private/extension/projectfile.bzl
xangcastle: No, bc uv don't generate dupliates in dev-dependencies for the same group.

PR #940, opened 2026-04-22
:bug: fix(venv): insidious creation/cleanup race condition

### LINKED ISSUE #858: [Bug]: venv collision on build_tool
### What happened?

```
@@aspect_rules_py++uv+sdist_build__bazel_pypi_lock__timezonefinder__8_2_0//:build_tool
  local spawn failed for Action
  stderr:
  Collision detected at destination: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/bazel-out/linux_host_platform-opt-exec-ST-f6bf7583020e/bin/external/aspect_rules_py+
  +uv+sdist_build__bazel_pypi_lock__timezonefinder__8_2_0/.build_tool/lib/python3.12/site-packages/BUILD.bazel
    - Source: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/external/aspect_rules_py++uv+project__bazel_pypi_lock/BUILD.bazel (Symlink)
    - Source: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/external/aspect_rules_py++uv+sdist_build__bazel_pypi_lock__timezonefinder__8_2_0/BUILD.bazel (Symlink)
  Error:   × Unable to run command:
    ╰─▶ Multiple collisions detected. Aborting.
  @@aspect_rules_py++uv+sdist_build__bazel_pypi_lock__crontab__1_0_1//:build_tool
  local spawn failed for Action
  stderr:
  Collision detected at destination: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/bazel-out/linux_host_platform-opt-exec-ST-f6bf7583020e/bin/external/aspect_rules_py++uv+sdist_build__bazel_pypi_lock__crontab__1_0_1/.build_tool/lib/
  python3.12/site-packages/BUILD.bazel
    - Source: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/external/aspect_rules_py++uv+project__bazel_pypi_lock/BUILD.bazel (Symlink)
    - Source: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/external/aspect_rules_py++uv+sdist_build__bazel_pypi_lock__crontab__1_0_1/BUILD.bazel (Symlink)
  Error:   × Unable to run command:
    ╰─▶ Multiple collisions detected. Aborting.
  @@aspect_rules_py++uv+sdist_build__bazel_pypi_lock__lz4__4_4_3//:build_tool
  local spawn failed for Action
  stderr:
  Collision detected at destination: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/bazel-out/linux_host_platform-opt-exec-ST-f6bf7583020e/bin/external/aspect_rules_py++uv+sdist_build__bazel_pypi_lock__lz4__4_4_3/.build_tool/lib/
  python3.12/site-packages/BUILD.bazel
    - Source: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/external/aspect_rules_py++uv+project__bazel_pypi_lock/BUILD.bazel (Symlink)
    - Source: /var/lib/applied-bk-bazel-runner/bazel/execroot/_main/external/aspect_rules_py++uv+sdist_build__bazel_pypi_lock__lz4__4_4_3/BUILD.bazel (Symlink)
  Error:   × Unable to run command:
    ╰─▶ Multiple collisions detected. Aborting.
```

### Version

Output of `bazel --version`:

### How to reproduce

```shell

```

### Any other information?

_No response_

13steinj:
Suffixes the venv created with the running shell's PID, ensuring that each call generates a venv at a unique location.

Also, explicitly cleans the venv up at the end to avoid too many piling up.

Fixes #858, #899.

---

I am near certain the issue I've been wrestling with is the same as #858 and that the root cause that this PR fixes is the same as the root cause for #899. I don't know if there is any code path in which the venv can be created outside the bazel output base, which is why I added the `trap` cleanup. If there isn't I am happy to remove that line.

---

### Test plan

- Manual testing; On a large enough codebase that uses `py_binary` and specific `venv`s (say, via the unstable `uv` based rules and `dependency_groups` and a `uv.lock` lockfile, and more specifically, one that has build actions that use a `py_binary` as a code generator), attempt to build on a beefy 96 core machine with just the right actions cached such that the venvs collide.
  Sadly as this is a race condition your machine needs to be just beefy enough that the actions run in the right order to create the perfect storm of pain.
  Then, do it again with this branch, and see that the venvs no longer collide.
  If one is using remote execution, you will not (or at least it will be very difficult) to be able to reproduce the original bug, because it appears that under remote execution the venvs are (a) potentially sharded across different nodes, and (b) I imagine the remote execution implementation defines its own prefix path per action, such that I am sure that there exist implementations that have a more unique path based on the action information.

### DIFF
--- a/py/private/run.tmpl.sh
+++ b/py/private/run.tmpl.sh
@@ -35,7 +35,8 @@ function python_location {
 }
 
 VENV_TOOL="$(rlocation {{VENV_TOOL}})"
-VIRTUAL_ENV="$(alocation "${RUNFILES_DIR}/{{ARG_VENV_NAME}}")"
+VIRTUAL_ENV="$(alocation "${RUNFILES_DIR}/{{ARG_VENV_NAME}}.$$")"
+trap 'rm -rf "${VIRTUAL_ENV}"' EXIT
 export VIRTUAL_ENV
 
 "${VENV_TOOL}" \


dzbarsky:
Isn't this moot with static venv? Is the plan to deprecate/delete all this code? :)

jbedard:
Yes I think this will all go away, but this might still be worth merging in 1.x for now?

13steinj:
> Isn't this moot with static venv? Is the plan to deprecate/delete all this code? :)

I don't have context on this at all, but would generally appreciate a merge to avoid local patches getting stale :)

13steinj:
Assuming you're referring to #967 now that I am seeing this at a reasonable hour.
I don't care how this gets fixed so long as it does, but I would kindly ask / suggest it's worth merging to fix in 1.x for now, to avoid the scenario that the new behavior has a change or breakage that people can't jump forward to.

jbedard:
@13steinj I am quite hesitant to be creating random directories like this and then depending on `trap` to delete them which is why I haven't merged this. I'm not certain how reliable the `trap` is, or if this might start leaving a collection of infinitely-growing files behind?

13steinj:
@jbedard
`trap ... EXIT` is reliable in bash/zsh: it fires on normal exit, `set -e` aborts, and most signals. The main miss case is `SIGKILL`, but that's true of any cleanup mechanism.

As for accumulation: these directories (normally) live inside the output base, so bazel clean sweeps them regardless. They won't grow unboundedly like a `/tmp` leak would (but if someone does something funky and manages to force these be made in `/tmp`, most modern linux systems use tmpfs / shm.

Happy to add some additional belt-and-suspenders of your choosing, but I think this is sufficient as-is. We've been running with this patch, with remote execution and not, for the past 3-4 weeks on a fairly large monorepo with a lot of Python and nobody has reported a bug, and none of the internal monitoring has sounded alarms on disk space related to python venvs.

13steinj:
But now that there's a conflict, I assume you'd like me to verify that the latest release (or latest main?) still has the issue?

jbedard:
main is now moving toward an alpha release of v2 which will not have this bug anymore. If you'd still like to debate adding this to a v1 patch I can create a v2 branch that this PR will target?

13steinj:
I am fine using v2 assuming it "Just Works" :tm:, once out I can test and report back. When do you expect v2 to be released?

jbedard:
I hope to have an alpha release maybe early next week 🙏

13steinj:
please let me know if you have any updates on this alpha.

As an aside, I tried updating to the latest commit several times last week (most recently a1b9e0716ae4c47696eca58b13d048fe4eae4fdd) ; am I correct in understanding that `rules_rs` and a full rust toolchain is now a hard dependency of `aspect_rules_py` ?

This would make the upgrade (and even testing it) significantly more difficult in our environment.

jbedard:
If you build from source then rules_rs is required to compile the rust code. When a release is tagged the rust code is pre-compiled and rules_rs is no longer required

13steinj:
I have tested the alpha tag and things _look_ good? Going to leave this PR open until I can fully validate against a full release though, if that's alright?

I plan to upstream another coworker's small feature patch later today/tomorrow that I hope can make it in to 2.0; as to avoid having the patch internally.

jbedard:
Close?

13steinj:
Happy to close , I'll leave the branch around in case someone comes knocking for 1.x though

PR #942, opened 2026-04-23
feat: new py_image_layer

thesayyn:

TODO

### Test plan

TODO
- New test cases added
- Manual testing; please provide instructions so we can reproduce: TODO


py/private/py_image_layer.bzl
```diff
@@ -1,215 +1,891 @@
...
+    bsdtar, bsdtar_files = _tar_toolchain(ctx)
+
+    seen_labels = {}
+    all_pkgs = []
+    for pkg in info.pip_packages.to_list():
+        if pkg.label not in seen_labels:
+            seen_labels[pkg.label] = True
+            all_pkgs.append(pkg)
+
+    pkg_by_label = {pkg.label: pkg for pkg in all_pkgs}
```
akafael, line 652: This code is doing a bit of double work. It manually filters for unique labels using a dictionary and a list, then immediately builds another dictionary at the end.

Since the final goal seems to be creating a dictionary mapped by labels (pkg_by_label), we can skip the intermediate list and the "seen" check entirely to reduce the memory usage.

Suggestion:

Option 1; (Shorter version, that keeps the last occurrence of each package)
```suggestion
    pkg_by_label = {pkg.label: pkg for pkg in info.pip_packages.to_list()}
    all_pkgs = pkg_by_label.values()
```

Option 2: (Slightly modified version that keeps the first occurrence of each package)
```suggestion
pkg_by_label = {}
for pkg in info.pip_packages.to_list():
    if pkg.label not in pkg_by_label:
        pkg_by_label[pkg.label] = pkg

all_pkgs = pkg_by_label.values()
``

File: py/private/py_image_layer.bzl [hunk 2]
```diff
@@ -1,215 +1,891 @@
...
+
+    source_tar = _declare_group_tar(
+        ctx, bsdtar, bsdtar_files,
+        "{}_default.tar.gz".format(ctx.attr.name),
+        "default",
+        info.source_files,
+        _source_map,
+        "Creating source layer for %s" % ctx.label,
+    )
+    all_tars.append(source_tar)
```
akafael, line 747: **Suggestion**: Eliminate Redundant Array Appends in `_py_image_layer_impl`

 In `_py_image_layer_impl`, `tar_out` items are manually appended to both `all_tars` and `dep_tars` quite often. 

Since `dep_tars` is essentially identical to `all_tars` *except* for the `source_tar` appended at the very end, you can entirely drop the `dep_tars` variable during iteration and simply clone the list at the end.

**Proposed solution:**
1.  Just append to all_tars normally inside your loops
2. Before appending the source layer, clone `all_tars` to act as `dep_tars`

```suggestion
    dep_tars = list(all_tars)
    
    source_tar = _declare_group_tar(
        ctx, bsdtar, bsdtar_files,
        "{}_default.tar.gz".format(ctx.attr.name),
        "default",
        info.source_files,
        _source_map,
        "Creating source layer for %s" % ctx.label,
    )
    all_tars.append(source_tar)
```

py/private/py_image_layer.bzl
thesayyn: oh nice, i actually did not write this code, so probably there is a lot of optimizations here.

akafael: This new version is really nice! It provides some great optimizations considering the previews version. These are just suggestions to improve even it further.

py/private/BUILD.bazel
```diff
@@ -20,14 +22,21 @@ py_library(
     deps = ["//py/private/pytest_shard"],
 )
 
+_rules_python_py_binary(
```
jbedard, line 25: We need to move toward removing the `rules_python` dep. What is the reason for this?

py/tests/internal-deps/adder/BUILD.bazel
```diff
@@ -9,9 +9,5 @@ py_library(
     imports = [".."],
     # This library contributes to the container test, testing we can pull in and use a library from another
     # package in the repo.
-    visibility = [
```
jbedard, line 12: revert?

py/private/py_image_layer.bzl
```diff
@@ -1,215 +1,891 @@
...
+    matched py_library targets in the binary's dep closure.
+  - Ungrouped pip packages: squashed by the rule into one per-rule tar.
+"""
+
+load("@rules_python//python:defs.bzl", "PyInfo")
+load("//py/private/toolchain:types.bzl", "PY_TOOLCHAIN")
+
+_TAR_TOOLCHAIN = "@tar.bzl//tar/toolchain:type"
+
+_WHL_INSTALL_PREFIX = "aspect_rules_py++uv+whl_install__"
```
jbedard, line 40: bazel 8 vs 9 have different separators (`+` vs `~`) so this will fail

PR #944, opened 2026-04-23
py_binary venvs: analysis-time assembly + shared `external_venv` (v2.0.0)

gregmagolan:
## What

Overhauls how `py_binary` / `py_test` acquire virtualenvs and consolidates the API surface. Ships as **v2.0.0**.

**Legacy `py_binary`** staged its site-packages tree at launcher startup: every invocation ran a Rust `venv_bin` tool that walked the runfiles, created symlinks under a temp venv dir, and then exec'd Python against it. On small graphs that's a few hundred ms of overhead; on large monorepos with thousands of wheels we measured 10s of seconds per `bazel run` / `bazel test` — on every invocation, not amortised by caching. **Legacy `py_venv_binary`** avoided that cost by building a real on-disk venv as a Bazel `declare_directory` tree artifact, but paid for it with tree-artifact remote-execution / remote-cache penalties and a different launcher shape that didn't integrate with `py_image_layer` or the rest of the `py_binary` surface.

This PR unifies both onto a single model: the venv is a tree of individual Bazel outputs (`ctx.actions.symlink` + `ctx.actions.write` outputs, not a tree artifact) produced at analysis time. The launcher is a no-op by comparison — Bazel has already placed the venv files in runfiles — so startup is milliseconds, not seconds, regardless of graph size. On top of that, `py_binary(external_venv = :shared_venv)` lets many binaries share a single `py_venv` target, and `py_binary(expose_venv = True)` auto-emits a first-class sibling `:<name>.venv` py_venv per target so the shared-venv pattern is available per-callsite without a separate declaration.

## Why

Two user-visible wins:

1. **Faster startup.** The launcher no longer runs a Rust binary to materialise the venv before exec'ing Python. Improvement scales with wheel count — biggest impact on large monorepos, where launcher cost was dominant.

2. **Shared virtualenvs.** The new `external_venv` attribute lets many `py_binary` / `py_test` targets share a single `py_venv`. Point your IDE at one venv and every entrypoint in the repo resolves to the same interpreter + dep closure. An analysis-time coverage check rejects binaries whose dep closure isn't a subset of the venv's, with a clear error naming the missing wheels.

A few smaller ergonomic improvements fell out:
- `env = {…}` and `env_inherit = […]` declared on a `py_venv` flow through to `py_binary(external_venv = …)` consumers; binary-level `env` wins on key conflicts.
- `isolated = False` on `py_binary` opts out of Python's `-I` flag for code that needs `PYTHONPATH` or script-dir-on-sys.path semantics (matches the legacy `py_venv_binary` shape).
- `py_image_layer` produces valid container tars in all configurations we test (container test coverage turned out to have been latently-broken before this work).
- Stable load paths: `@aspect_rules_py//py:defs.bzl`, `//py:extensions.bzl`, `//uv:defs.bzl`, `//uv:extensions.bzl`. The `/unstable/` paths graduated.

## Shape of the change

**`py_binary` / `py_test` gain `expose_venv`.** When `expose_venv = True` is passed, the macro splits into a first-class sibling `:<name>.venv` `py_venv` target carrying all venv-shaping attrs (`deps`, `imports`, `package_collisions`, `include_*_site_packages`, `interpreter_options`) plus the binary/test rule with `external_venv = ":<name>.venv"`. The `.venv` sibling is consumable (other targets can `external_venv = "//:<name>.venv"`) and runnable (`bazel run :<name>.venv` drops into the hermetic interpreter). `expose_venv` defaults to `False` so default `py_binary` callers emit exactly one target — no graph bloat.

**Auto-emitted `<name>.venv` sibling removed.** In v1.x the `py_binary` macro unconditionally emitted a sibling `:<name>.venv` link target. That doubled the node count in `bazel query :*` for every caller and forced the internal venv to awkwardly avoid the `<name>.venv` name. v2.0.0 drops the auto-emit; users who want the IDE-materialise-to-workspace behaviour declare `py_venv_link(name = ..., venv = ":<name>.venv")` explicitly (and must also pass `expose_venv = True` on the binary to get a sibling venv to point at).

**`py_venv_link` is now an explicit opt-in rule.** Its signature changed: takes `venv = <label>` pointing at an existing `py_venv` (typically the one from `expose_venv = True`), produces a runnable target whose `bazel run` materialises a workspace-local symlink to the venv tree.

**`py_venv_binary` / `py_venv_test` were removed.** Calling them now `fail()`s at analysis with a clear migration message. Replacement is `py_binary` / `py_test` with `expose_venv = True, isolated = False`. The legacy deprecation-warning path (print-and-delegate) is gone; the failure is the one-and-only migration signal.

**`/unstable/` load paths were removed.** Everything previously at `//py/unstable:{defs,extension}.bzl` and `//uv/unstable:{defs,extension}.bzl` moved to the stable `//py:defs.bzl`, `//py:extensions.bzl`, `//uv:defs.bzl`, `//uv:extensions.bzl` paths. Loading the old paths `fail()`s with a message naming the replacement.

**Dropped Rust tooling.** The `venv_bin` and `venv_shim` crates and their toolchain types (`VENV_TOOLCHAIN`, `VENV_EXEC_TOOLCHAIN`, `SHIM_TOOLCHAIN`) are gone — their jobs (runtime venv assembly, interpreter-indirection shim) are now done in Starlark. Only `unpack` remains, collapsed into a single crate under `py/tools/unpack/`. A future shell-out to the `uv` CLI could remove that too once the uv toolchain lands.

**Two-hop symlinks for containers.** The venv's per-top-level `site-packages/<name>` symlinks were previously calibrated for the runfiles layout only, which broke `py_image_layer`'s bazel-bin tar walk. They now route through a `<venv>/_wheels/<i>/` directory alias whose relative targets resolve identically from bazel-bin, runfiles, and inside an OCI container.

**New test coverage.** Four regression tests added for pytest-integration machinery that previously relied only on downstream usage to validate: pytest-xdist parallel-worker execution, `py_pytest_main(chdir = ...)` template substitution, `COVERAGE_MANIFEST` / LCOV post-processing in `pytest_main.py`, and multiple `py_binary` + `py_test` targets sharing a single `external_venv` py_venv.

## tar.bzl upstream fixes

Building `py_image_layer` on real-world inputs surfaced two bugs in [bazel-contrib/tar.bzl](https://github.com/bazel-contrib/tar.bzl)'s `preserve_symlinks.awk`:

- [#106 — `make_relative_link` off-by-one](https://github.com/bazel-contrib/tar.bzl/pull/106). Emits N `../`s for a path of N segments; should be N-1. Produces dangling symlinks whose traversal escapes the archive.
- [#107 — symlink detection under any transitioned config](https://github.com/bazel-contrib/tar.bzl/pull/107). Classifier was scoped to the mtree_mutate action's own `bin_dir` and missed targets under transitioned configs (`-ST-<hash>` suffix) and under `external/<repo>/`. Also flips `readlink` order so authored `declare_symlink + target_path` strings land in the archive verbatim.

Both fixes are carried locally in `py/private/modify_mtree.awk` via a small custom `mtree_preserve_symlinks` rule. Once the upstream PRs merge and we bump `tar.bzl`, the local fork retires and `py_image_layer` reverts to `mtree_mutate(preserve_symlinks = True)`.

## Migration guide

For a walkthrough with before/after examples and a troubleshooting section: [`docs/migrating_v1_v2.md`](https://github.com/aspect-build/rules_py/blob/optimized_py_binary/docs/migrating_v1_v2.md).

## Breaking changes (v2.0.0)

All breaking changes surface with a clear error at analysis time; `bazel test //...` after the bump names every callsite that needs updating.

### Removed macros — `py_venv_binary` / `py_venv_test`

Calling either macro now `fail()`s. Replacement is plain `py_binary` / `py_test` with `expose_venv = True, isolated = False`. Plain `py_binary(...)` with no extra attrs is the common case (analysis-time venv assembly is the default); the two extra attrs exist for callers who need the old split-target shape plus PYTHONPATH-honoring launcher.

```starlark
# v1.x
py_venv_binary(
    name = "my_tool",
    srcs = ["my_tool.py"],
    main = "my_tool.py",
    deps = [...],
)

# v2.0.0 — common case
py_binary(
    name = "my_tool",
    srcs = ["my_tool.py"],
    main = "my_tool.py",
    deps = [...],
)

# v2.0.0 — preserve the old split + launcher semantics
py_binary(
    name = "my_tool",
    srcs = ["my_tool.py"],
    main = "my_tool.py",
    deps = [...],
    expose_venv = True,
    isolated = False,
)
```

### Removed load paths — `/unstable/`

`//py/unstable:{defs,extension}.bzl` and `//uv/unstable:{defs,extension}.bzl` are gone. Each now `fail()`s at load time with a message pointing at the stable replacement.

```starlark
# v1.x
load("@aspect_rules_py//py/unstable:defs.bzl", "py_venv")
load("@aspect_rules_py//uv/unstable:defs.bzl", "gazelle_python_manifest")
interpreters = use_extension("@aspect_rules_py//py/unstable:extension.bzl", "python_interpreters")
uv = use_extension("@aspect_rules_py//uv/unstable:extension.bzl", "uv")

# v2.0.0
load("@aspect_rules_py//py:defs.bzl", "py_venv")
interpreters = use_extension("@aspect_rules_py//py:extensions.bzl", "python_interpreters")
```

### Removed auto-emitted sibling — `:<name>.venv` no longer shows up for free

In v1.x, every `py_binary(name = "foo")` call auto-generated a `:foo.venv` link target. v2.0.0 drops the auto-emit. Callers who ran `bazel run :foo.venv` for IDE integration must either:

- Opt into `expose_venv = True` on the binary to get a consumable + runnable `:foo.venv` py_venv, then declare `py_venv_link` explicitly if they also need the workspace-materialise behaviour, or
- Stop relying on the sibling entirely (point the IDE at a `py_venv` declared separately, or at bazel-bin directly).

```starlark
# v1.x — :my_app.venv auto-generated
py_binary(name = "my_app", ...)
# $ bazel run :my_app.venv          # worked

# v2.0.0 — opt in explicitly
py_binary(name = "my_app", expose_venv = True, ...)
py_venv_link(name = "my_app_ide", venv = ":my_app.venv")
# $ bazel run :my_app.venv          # drops into interpreter
# $ bazel run :my_app_ide           # materialises workspace symlink
```

### Changed signature — `py_venv_link`

Was: takes `deps` / `srcs` / `imports` and builds its own internal venv.
Now: takes `venv = <label>` pointing at an existing `py_venv` target. Pair with `py_binary(expose_venv = True, ...)` or a standalone `py_venv` declaration.

```starlark
# v1.x
py_venv_link(
    name = "my_venv",
    srcs = ["main.py"],
    deps = ["@pypi//fastapi"],
)

# v2.0.0
py_venv(
    name = "my_venv_target",
    deps = ["@pypi//fastapi"],
)
py_venv_link(
    name = "my_venv",
    venv = ":my_venv_target",
)
```

### Removed Rust toolchain types

`VENV_TOOLCHAIN`, `VENV_EXEC_TOOLCHAIN`, and `SHIM_TOOLCHAIN` are gone — their work (runtime venv staging, interpreter-indirection shim) is now done in Starlark at analysis time. Users who registered overrides for these toolchain types must delete those registrations; leaving them in place produces toolchain-resolution errors.

### Changed venv internal layout — two-hop site-packages symlinks

The per-top-level `<venv>/lib/python<ver>/site-packages/<top>` symlinks now route through an intermediate `<venv>/_wheels/<i>/` directory alias instead of pointing directly at the wheel's materialised tree. Python-level consumers (site.py, importlib, every normal user) see no difference. Filesystem-walking tools that iterate the venv tree and expect one-hop symlinks — custom tar rules, PEX-style packagers, docker build scripts that walk `site-packages/` — may need updating. The relative-link depth is the same in bazel-bin, runfiles, and inside an OCI image, so the new shape is strictly more portable than the old one.

### Unchanged (explicitly, for the skim-reader)

- The string `venv = "..."` attribute (uv pip-extension wheel selection) is unchanged. The new `external_venv = :label` is a separately-typed attribute that coexists with it.
- `py_binary` callers who don't opt into `external_venv`, `expose_venv`, or `isolated = False` see no behavioural changes beyond the startup-performance win.
- The default on-disk venv basename stays `.<name>.venv/` — same as v1.x. IDEs auto-detect the path unchanged.


uv/private/whl_install/rule.bzl
```diff
@@ -85,18 +91,26 @@ def _whl_install(ctx):
...
+            imports = depset([site_packages_rfpath]),
             has_py2_only_sources = False,
             has_py3_only_sources = True,
             uses_shared_libraries = False,
         ),
     ]
 
+    if ctx.attr.top_levels or ctx.attr.console_scripts:
+        providers.append(PyWheelInfo(
+            wheels = depset(direct = [struct(
```
jbedard, line 103: A `depset` of `struct`s. IDK if I've seen that anywhere before, have you seen or done that before?

py/private/providers.bzl
```diff
@@ -1,10 +1,23 @@
...
+`ctx.actions.symlink` per top-level name, merging wheels into a single
+`site-packages/` tree without invoking a runtime tool.
+""",
     fields = {
-        "files": "Depset of all files including deps for this wheel",
-        "default_runfiles": "Runfiles of all files including deps for this wheel",
+        "wheels": "Depset of struct(top_levels = [str], site_packages_rfpath = str) — one per wheel in the transitive closure.",
+        # Legacy fields, not populated by rules_py itself. Retained so third-party
+        # consumers introspecting the provider's field list don't break.
+        "files": "DEPRECATED. Unused.",
```
jbedard: This seems breaking enough that I wonder if we should just create a new provider. Support both for now, delete this one in the next semver...

py/private/py_library.bzl
```diff
@@ -240,6 +262,7 @@ py_library_utils = struct(
     make_instrumented_files_info = _make_instrumented_files_info,
     make_merged_runfiles = _make_merged_runfiles,
     make_srcs_depset = _make_srcs_depset,
+    make_wheels_depset = _make_wheels_depset,
```
jbedard, line 217: I don't think this is used and can be dropped?

py/private/py_library.bzl
```diff
@@ -155,6 +155,24 @@ def _make_imports_depset(ctx, imports = [], extra_imports_depsets = []):
...
+def _make_wheels_depset(ctx, extra_transitive = []):
+    """Collect PyWheelInfo.wheels across deps + resolutions.
+
+    Used by downstream rules that want a flat list of wheels (with their
+    declared top-level names) in the transitive closure, for building a
+    merged site-packages via symlinks.
+    """
+    transitive = [
+        target[PyWheelInfo].wheels
+        for target in getattr(ctx.attr, "deps", [])
```
jbedard, line 119: Why `getattr`? Is this util method used across multiple rules that have different attributes?

uv/private/whl_install/rule.bzl
gregmagolan: Claude defends his choice:

Fair question — it's an unusual shape. Let me give the honest tradeoffs.

Why I reached for it
py_library needs to accumulate wheel metadata transitively through deps. The depset idiom depset(transitive=[dep[PyWheelInfo].wheels for dep in ...]) gives O(1) merge at each node — which matters for graphs with hundreds of wheels (like examples/pytest:pytest_test with ~50). A list-based equivalent is O(N) per merge → O(N²) over the whole graph.

Bazel's depset does support arbitrary hashable Starlark values, not just File/Label/string. Structs are hashable if their fields are, so depset(direct=[struct(...)]) works and deduplicates by structural equality at flatten time.

Why it's still unusual
You rarely see this in practice because:

depset is strongly associated with File artifact sets — runfiles, transitive_sources, etc. Readers expect "this is a set of files." A struct snaps them out of that mental model and they have to re-parse what's going on.
Dedup cost is higher for structs — File equality is O(1) identity-ish; struct equality walks every field. Negligible at our scale (hundreds) but a code smell to buildfarm-scale maintainers.
.to_list() is always the first thing py_binary does anyway — we flatten immediately in _resolve_wheel_collisions. If we flatten at the first use, did we save anything by not flattening at the py_library level? Only if intermediate py_library nodes see the same depset repeatedly, which in Bazel's action graph they effectively do (depset is immutable and reference-shared).
The alternatives
(a) List of structs. Simple and readable. Merge via [] + ... in py_library. Acceptable for small graphs, potentially O(N²) for large. The dedup behavior becomes manual — same wheel reachable via two deps appears twice in the list.

(b) Dict of site_packages_rfpath → struct. Most semantically honest since each wheel is uniquely identified by its site-packages path. Natural dedup. Downside: merging dicts transitively through py_library requires manual loops (vs depset's one-liner) and silently tolerates conflicts if two wheels claim the same path.

(c) Depset of strings (our current struct fields serialized). Awkward — packs/unpacks three fields per row into a string, brittle against field changes.

My take
The argument for it: py_library nodes compose cheaply, dedup is structural and correct, and the performance wins on large graphs where this matters most.

py/private/py_library.bzl
gregmagolan: Used by py_binary: `wheels_depset = _py_library.make_wheels_depset(ctx)`

py/private/py_library.bzl
gregmagolan: yes, shared across py_library, py_binary/py_test, py_venv*, and py_unpacked_wheel — the last of which doesn't have deps at all. The getattr is consistent with every sibling helper in the file and matches that contract.

py/private/providers.bzl
gregmagolan: That's a good call. We can just leave the old name dormant and schedule it for removal

py/private/py_binary.bzl
```diff
@@ -168,21 +445,29 @@ Part of the aspect_rules_py//uv system, has no effect in rules_python's pip.
...
+* "ignore": Use the first wheel silently; the second is skipped.
+
+Wheels whose contents aren't visible to analysis (no `PyWheelsInfo`) can't
+collide here — they fall through to `.pth`-based resolution where Python's
+import order (first `sys.path` hit wins) decides.
 """,
         default = "error",
         values = ["error", "warning", "ignore"],
     ),
-    "interpreter_options": attr.string_list(
```
jbedard, line 168: This just moved?

py/private/py_binary.bzl
```diff
@@ -25,36 +183,160 @@ def _py_binary_rule_impl(ctx):
...
-    pth_lines.add(escape)
-
-    pth_lines.add_all(imports_depset, format_each = "{}/%s".format(escape))
-
-    site_packages_pth_file = ctx.actions.declare_file("{}.pth".format(ctx.attr.name))
+    # Collect wheels contributing metadata-rich PyWheelsInfo. Anything without
+    # PyWheelsInfo (hand-written py_unpacked_wheel without `top_levels`, or
+    # non-wheel py_library deps) falls through to the .pth path below.
+    wheels_depset = _py_library.make_wheels_depset(ctx)
+    wheels = wheels_depset.to_list()
```
jbedard: Did you think about avoiding the `to_list()` calls?

py/private/py_binary.bzl
```diff
@@ -221,7 +506,6 @@ py_base = struct(
     test_attrs = _test_attrs,
     toolchains = [
         PY_TOOLCHAIN,
-        VENV_TOOLCHAIN,
```
jbedard, line 210: Does this mean the entire toolchain can be deleted?

gregmagolan:
Put up https://github.com/bazel-contrib/tar.bzl/pull/106

py/private/py_binary.bzl
gregmagolan: Yes

py/private/py_binary.bzl
gregmagolan: This is intentional. We use depset throughout py_library and finally we de-dup in py_binary with to_list. More efficient the de-duping along the way.

gregmagolan:
Put up https://github.com/bazel-contrib/tar.bzl/pull/107

py/private/py_binary.bzl
gregmagolan: Yes. It moved to the py_venv rule now.

e2e/cases/uv-deps-650/crossbuild/toolchain_test.bzl
```diff
@@ -7,9 +7,9 @@ into a file, so sh_test scripts can inspect which binary was selected.
 UNPACK_TOOLCHAIN = "@aspect_rules_py//py/private/toolchain:unpack_exec_toolchain_type"
 
 def _unpack_toolchain_path_impl(ctx):
-    unpack_bin = ctx.toolchains[UNPACK_TOOLCHAIN].bin.bin
+    unpack = ctx.toolchains[UNPACK_TOOLCHAIN].bin.bin
```
jbedard: revert?

e2e/cases/root-dir-paths-538/BUILD.bazel
```diff
@@ -2,20 +2,34 @@ load("@aspect_rules_py//py:defs.bzl", "py_binary")
 load("@aspect_rules_py//py/unstable:defs.bzl", "py_venv_binary")
```
jbedard: Half of these have changed (to drop `/unstable`) and half have not. If we are keeping `py_venv_binary` we should decide where it lives and only load from one place...

e2e/cases/root-dir-paths-538/BUILD.bazel
```diff
@@ -2,20 +2,34 @@ load("@aspect_rules_py//py:defs.bzl", "py_binary")
...
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
 
-# py_binary uses run.tmpl.sh which has the alocation bug
```
jbedard, line 5: Is this no longer true? Or is the robot deleting random comments?

e2e/cases/uv-deps-650/crossbuild/toolchain_test.bzl
gregmagolan: intentional name change

py/private/py_semantics.bzl
```diff
@@ -104,8 +114,11 @@ def _determine_main(ctx):
         Artifact; the main file. If one can't be found, an error is raised.
     """
     if ctx.attr.main:
-        if not ctx.attr.main.label.name.endswith(".py"):
-            fail("main must end in '.py'")
+        # Check the RESOLVED file name, not the label name — the label may
```
jbedard: Is this (+ a test) something that can be merged now? Ask claude to cherry-pick this change + a test...

e2e/cases/root-dir-paths-538/BUILD.bazel
gregmagolan: No longer true. The new approaches fixed the root cause.

e2e/cases/root-dir-paths-538/BUILD.bazel
gregmagolan: We decided to drop unstable :)

py/private/py_venv/entrypoint.tmpl.sh
```diff
@@ -1,15 +1,29 @@
 #!/usr/bin/env bash
+# Launcher for py_venv targets — `bazel run :name` activates the venv
+# and exec's its bin/python interactively.
+#
+# (py_venv_binary / py_venv_test don't use this template — they expand
```
jbedard: This sentence seems irrelevant here.

py/private/py_venv/entrypoint.tmpl.sh
```diff
@@ -1,15 +1,29 @@
...
+# pyvenv.cfg, bin/python (+ versioned aliases), activate, site-packages
+# symlinks, etc. All this script does is rlocation-resolve bin/python,
+# export $VIRTUAL_ENV (for any child process that checks), and exec.
 
 if {{DEBUG}}; then
     set -x
 fi
 
 {{BASH_RLOCATION_FN}}
-
```
jbedard, line 8: revert so you can still read the code that gets generated 🙏

py/private/py_binary.bzl
```diff
@@ -14,79 +16,150 @@ def _dict_to_exports(env):
         for (k, v) in env.items()
     ]
 
+def _check_venv_coverage(ctx, imports_depset, wheels_depset, vinfo):
+    """Analysis-time check: everything the binary needs must live in the venv.
```
jbedard, line 21: Did you think about making this a validation action? Or making it optional?

IDK if it's worth changing it from validation-in-starlark to validation-action, but it would allow avoiding the `.to_list()` calls I think?

py/private/py_binary.bzl
```diff
@@ -99,27 +172,22 @@ def _py_binary_rule_impl(ctx):
...
     )
 
     instrumented_files_info = _py_library.make_instrumented_files_info(
         ctx,
         extra_source_attributes = ["main"],
     )
 
+    default_info_files = [executable_launcher, main]
+    if not external_venv:
+        default_info_files = default_info_files + extra_runfiles
```
jbedard, line 193: `extra_runfiles` got added to `DefaultInfo` in addition to `runfiles`? Normally I think we'd want to (a) keep them separate (b) not duplicate them?

jbedard: https://github.com/aspect-build/rules_py/pull/949 ?

py/private/py_semantics.bzl
jbedard: https://github.com/aspect-build/rules_py/pull/950

py/defs.bzl
```diff
@@ -42,15 +42,30 @@ load("//py/private:py_pex_binary.bzl", _py_pex_binary = "py_pex_binary")
...
+py_venv = _py_venv
 py_venv_link = _py_venv_link
 
+# Removed in v2.0.0: calling these `fail()`s at analysis with a
+# migration message pointing at `py_binary` / `py_test` with
+# `expose_venv = True, isolated = False`. Kept exported here so that
+# `load("@aspect_rules_py//py:defs.bzl", "py_venv_binary")` resolves
+# and the call-site failure surfaces the friendly message (instead of
+# Bazel's generic "symbol not exported" error).
+py_venv_binary = _py_venv_binary
```
jbedard: I think the deprecated ones should remain where they were. There's no reason to tell people to migrate from `py/unsafe/defs.bzl` to `py/defs.bzl` just to be told it's unsupported

py/private/py_binary.bzl
```diff
@@ -149,33 +238,113 @@ be deprecated in the future.
         doc = """The name of the Python virtual environment within which deps should be resolved.
 
 Part of the aspect_rules_py//uv system, has no effect in rules_python's pip.
+""",
+    ),
+    "external_venv": attr.label(
```
thesayyn, line 243: I find this attribute confusing albeit nice to have, unless its required by our customers i'd say we drop it and reconsider later if we still think its a good idea

py/private/py_image_layer.bzl
```diff
@@ -155,27 +156,29 @@ def py_image_layer(
...
-            srcs = [binary],
-            **kwargs
-        )
+    mtree_spec(
+        name = manifest_name + ".preprocessed",
+        srcs = [binary],
+        **kwargs
+    )
+
+    mtree_preserve_symlinks(
```
thesayyn, line 175: this break my new impl of py_image_layer i'll fix it there, lets drop it here.

xangcastle: This also break the pth function helper, but I think we can start from here, and then fix everything

py/private/py_venv/py_venv.bzl
```diff
@@ -295,37 +170,63 @@ Only works with the experimental Aspect pip machinery.
...
-  firstparty code in external repositories) and symlinks for 3rdparty wheels
-  (detected by the presence of site-packages or dist-packages in the path).
-  Copies and patches binaries.
-
-This mode is designed to minimize inode usage while still providing a fully
-functional virtualenv. First-party code is referenced via .pth files (one entry
-per import root), while third-party wheels are symlinked into the venv to ensure
-proper package metadata and console scripts work correctly.
-        """,
+        doc = """Legacy attr, no longer honored.
```
jbedard, line 186: If we're doing breaking changes this shoudl just be dropped

PR #946, opened 2026-04-24
perf: set repo and module reproducible = True where possible

jbedard:
### Test plan

### DIFF
--- a/py/BUILD.bazel
+++ b/py/BUILD.bazel
@@ -33,6 +33,7 @@ bzl_library(
         ":toolchains",
         "//py/private/release:version",
         "@aspect_tools_telemetry_report//:defs.bzl",
+        "@bazel_features//:features",
     ],
 )
 
@@ -45,6 +46,7 @@ bzl_library(
         "//py/private/toolchain:autodetecting",
         "//py/private/toolchain:repo",
         "//py/private/toolchain:tools",
+        "@bazel_features//:features",
         "@bazel_lib//lib:repositories",
         "@bazel_tools//tools/build_defs/repo:http.bzl",
     ],
--- a/py/extensions.bzl
+++ b/py/extensions.bzl
@@ -1,6 +1,7 @@
 "Module Extensions used from MODULE.bazel"
 
 load("@aspect_tools_telemetry_report//:defs.bzl", "TELEMETRY")  # buildifier: disable=load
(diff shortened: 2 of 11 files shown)


jbedard:
FAQ: 

Still not marked reproducible, with the reason:                                                                                                                                                                                       
                                                                                                                                                                                                                                        
  Repository rules                                                                                                                                                                                                                      
  - local_python_interpreter (py/private/interpreter/repository.bzl:534) — probes a local interpreter via rctx.execute                                                                                                                  
  - _autodetecting_py_toolchain (py/private/toolchain/autodetecting.bzl:52) — auto-detects python3 on PATH                                                                                                                              
  - host_platform_repo (uv/private/host/repository.bzl:83) — local = True, probes host OS/libc                                                                                                                                          
  - system_diffutils (uv/private/diffutils/repository.bzl:17) — local = True, probes host patch                                                                                                                                         
  - sdist_build (uv/private/sdist_build/repository.bzl:241) — runs a user-supplied configure tool via rctx.execute; output depends on an external command                                                                               
  - git_archive (uv/private/git_archive/repository.bzl:83) — correctly conditional: reproducible when commit is pinned, not when ref is resolved via git ls-remote. Already uses repo_metadata to signal this per-invocation — leave    
  as-is.                                                                                                                                                                                                                                
                                                                                                                                                                                                                                        
  Module extensions                                                                                                                                                                                                                     
  - uv (uv/private/extension/defs.bzl:726) — declares reproducible, but internally calls sdist_build, which is not. The extension's reproducibility claim is only as strong as its children; worth revisiting if sdist_build can't be   
  cleaned up.                                                                                                                                                                                                                           
  - python_interpreters (py/private/interpreter/extension.bzl:580) — already dynamically reproducible: sets reproducible = False only when a user passes "latest" (which hits the GitHub API). Leave as-is.
                                                                                                                                                                                                                                        
  The first four (local/autodetect) genuinely cannot be reproducible. sdist_build is the interesting one — it's the only non-local, non-network rule on the "no" list, and it's what keeps the uv extension's reproducibility claim     
  leaky. Fixing it would require the configure tool to be deterministic (pure function of the sdist archive), which may already be true in practice — worth a look if you want to tighten this further.

gregmagolan (approved):
🌮

PR #955, opened 2026-04-28
refactor: upgrade toml2json to 0.0.26

jbedard:
### Test plan


dzbarsky (approved):
approved, but i think this can be simplified to much less code, here's how rules_rs handles it:

https://github.com/hermeticbuild/rules_rs/blob/597eefe2a71e192799863d739d0ecef29d866e22/MODULE.bazel#L35-L50 and https://github.com/hermeticbuild/rules_rs/blob/597eefe2a71e192799863d739d0ecef29d866e22/rs/private/toml2json.bzl#L3-L4

uv/private/tomltool/toml.bzl
```diff
@@ -1,79 +1,18 @@
-"""
+""""""
```
dzbarsky: is this appeasing a linter or something? so silly :)

dzbarsky (approved):

uv/private/tomltool/toml.bzl
jbedard: I updated to something slightly more descriptive lol

PR #957, opened 2026-04-28
refactor: cleanup unused bazel/platforms/* logic

jbedard:
### Test plan


bazel/platforms/BUILD.bazel
```diff
@@ -1,36 +1,28 @@
...
-        p.os,
-        p.cpu,
-        linker_suffix(
-            "_",
-            p.linker,
-        ),
-    ),
+    name = "linux_{}".format(cpu),
     constraint_values = [
-        "@platforms//os:{}".format(p.os),
```
dzbarsky, line 13: you can just do `parents = ["@rules_rs//rs/platforms:linux-x86-gnu"]` or whatever it's called, ditto for arm64 one. Just unroll the comprehension. Or honestly just use the upstream ones directly...

bazel/platforms/BUILD.bazel
```diff
@@ -1,36 +1,28 @@
...
+) for cpu in [
+    "aarch64",
+    "x86_64",
+]]
 
-alias(
-    name = "linux_arm64",
-    actual = ":linux_aarch64",
+[platform(
+    name = "macos_{}".format(cpu),
```
dzbarsky: same here. The advantage will be that in future we will set custmo build flags on those platforms as well to make the over build even closer to 0-config out of hte box

dzbarsky (approved):

xangcastle (approved):

PR #962, opened 2026-05-04
fix: ensure wheels in lockfiles have hashes

tsudol-plaid:
It is possible for a `package.wheels` entry to not have a `hash` field, for unclear reasons. This causes various parts of rules_py to break because we assume all wheels have a hash. To fix this, `normalize_deps` calculates a hash for any wheel that is missing one.

### DIFF
--- a/uv/private/extension/lockfile.bzl
+++ b/uv/private/extension/lockfile.bzl
@@ -52,6 +52,12 @@ def normalize_deps(lock_id, lock_data):
             # So we need to extract the version component here
             dep["version"] = default_versions.get(dep["name"])[2]
 
+    def _ensure_wheel_hash(whl):
+        if "hash" not in whl:
+            # Hashes in uv.lock start with "sha256:". For compatibility, we prefix these
+            # hashes with "sha1:". (The ':' is what matters for our purposes.)
+            whl["hash"] = "sha1:" + sha1(whl["url"])
+
     for spec in lock_data.get("package", []):
         # Backfill the sdist URL if the source is a URL file
         if "sdist" in spec and not "url" in spec["sdist"]:
@@ -63,6 +69,9 @@ def normalize_deps(lock_id, lock_data):
         for extra_deps in spec.get("optional-dependencies", {}).values():
             for dep in extra_deps:
                 _fix_version(dep)
+
+        for whl in spec.get("wheels", []):
+            _ensure_wheel_hash(whl)
 
     return default_versions, package_versions, lock_data
 
@@ -196,17 +205,12 @@ def collect_bdists(lock_data):
     bdist_table = {}
     for package in lock_data.get("package", []):
         for bdist in package.get("wheels", []):
-            identifier = None
-            if "hash" in bdist:
-                hash = bdist["hash"]
-                identifier = hash.split(":")[1][:16]
-            else:
-                hash = sha1(bdist["url"])
-                identifier = hash[:16]
-
+            # normalize_deps ensures what every wheel has a "hash" field.
+            bdist_hash = bdist["hash"]
+            identifier = bdist_hash.split(":")[1][:16]
             bdist_repo_name = "whl__{}__{}".format(package["name"], identifier)
             bdist_specs[bdist_repo_name] = bdist
-            bdist_table[hash] = "@{}//file".format(bdist_repo_name)
+            bdist_table[bdist_hash] = "@{}//file".format(bdist_repo_name)
 
     return bdist_specs, bdist_table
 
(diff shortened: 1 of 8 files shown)


tsudol-plaid:
I ran the tests locally, but I suspect there isn't a specific test for this case. What's the right way to add a test for this?

jbedard (approved):

jbedard:
Do you have a lockfile that reproduces this? The best way might would be trimming that lockfile down to the minimum which reproduces it and then loading that file as a test. Maybe checkout `e2e/cases/*/{BUILD,uv.lock}` files and do something similar to those cases?

tsudol-plaid:
> Do you have a lockfile that reproduces this? The best way might would be trimming that lockfile down to the minimum which reproduces it and then loading that file as a test. Maybe checkout `e2e/cases/*/{BUILD,uv.lock}` files and do something similar to those cases?

Right, that makes sense. I added a test case for it, based on the package (`torch`) where I originally found this issue.

Unfortunately I'm running on a mac, so I'm not sure if the test case I wrote is broken or if it just doesn't run on mac.

jbedard:
That seems to reproduce it for me locally, and your fix then works, and CI seems to agree 👍 

I might just trim the uv.lock down to a minimum but otherwise LGTM!

uv/private/extension/lockfile.bzl
```diff
@@ -52,6 +52,12 @@ def normalize_deps(lock_id, lock_data):
             # So we need to extract the version component here
             dep["version"] = default_versions.get(dep["name"])[2]
 
+    def _ensure_wheel_hash(whl):
+        if "hash" not in whl:
+            # Hashes in uv.lock start with "sha256:". For compatibility, we prefix these
+            # hashes with "sha1:". (The ':' is what matters for our purposes.)
+            whl["hash"] = "sha1:" + sha1(whl["url"])
```
jbedard: Does the `sha1` actually get used anywhere? Notice the commit I pushed where this actually broke things because this hash was being used as a sha256 hash in uv/private/extension/defs.bzl...

jbedard: I guess it is used as the `identifier` down below in this same file so it still serves a purpose even if not passed as the `http_file(sha256)`?

jbedard:
@tsudol-plaid can you sign the CLA? I think navigating to https://cla-assistant.io/aspect-build/rules_py?pullRequest=962 should work

Also please review the commit I added

uv/private/extension/lockfile.bzl
jbedard: Also... shouldn't this be the hash of the content of the URL? So I think we are calculating the wrong hash here and this fix should maybe be changed back to more like what it was doing previously where it never assigns to `whl["hash"]`...

jbedard: See the second commit of mine for that

tsudol-plaid: My thinking was that instead of updating every place that accesses the `hash` field so that it's aware that `hash` may not exist, it makes more sense to extend `normalize_deps` to add missing hashes. Hence why it writes to `whl["hash"]` with the hash of the wheel's URL, since that's what `collect_bdists` calculates for the hash.

You have a deeper understanding of what we're trying to achieve here, so I'll leave it up to you!

tsudol-plaid:
I've thought about it further, and using the `url` as the unique ID instead of `hash` makes more sense. The point of hashes is to verify the downloaded object, not necessarily to uniquely identify it. (At least, that's what I get from the [uv docs](https://docs.astral.sh/uv/reference/cli/#uv-pip-sync--require-hashes).) URLs, on the other hand, are unique by definition.

Your changes look good to me!

uv/private/extension/lockfile.bzl
jbedard: The problem was it was later assumed that `whl["hash"]` can be used as the `http_file(sha256)` (see second change in uv/private/extension/defs.bzl that protects against that now).

That change in uv/private/extension/defs.bzl fixed the issue, but made me too scared to put a sha1 into the `"hash"` key, who knows who else assumes its' a sha256... 😨

jbedard:
Note this was merged to `main` which is currently a 2.x WIP.

Would you prefer this goes into a `1.x` branch+release? I haven't created that branch yet but will if necessary...

tsudol-plaid:
> Note this was merged to `main` which is currently a 2.x WIP.
> 
> Would you prefer this goes into a `1.x` branch+release? I haven't created that branch yet but will if necessary...

That would be helpful if it's not too much trouble, but I can also get by with using my local copy for now.

Is there a timeline for 2.x?

jbedard:
I cherry-picked it into a 1.x branch and will tag a release soon.

I hope to have an alpha for 2.x within a week but no guarantee about how stable that will be yet.

PR #966, opened 2026-05-05
chore: define linux_gnu platform in e2e instead of via @aspect_rules_py

jbedard:
### Test plan


e2e/.bazelrc
```diff
@@ -10,8 +10,8 @@ common --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
 # platform for gnu Linux targets. The auto-detected host platform doesn't include it, so we
 # use a custom platform that inherits host cpu/os constraints and adds gnu.2.28.
 common --enable_platform_specific_config
-common:linux --platforms=@aspect_rules_py//bazel/platforms:linux_gnu
-common:linux --host_platform=@aspect_rules_py//bazel/platforms:linux_gnu
```
jbedard, line 14: Once published to BCR the `@aspect_rules_py//bazel/platforms:linux_gnu` target which uses `llvm` [becomes private](https://github.com/aspect-build/rules_py/blob/5dda303a619d485ce2b8c9681ba71c430ee4f080/MODULE.bazel#L106) so it must be duplicated in the e2e test that BCR runs

xangcastle (approved):

PR #965, opened 2026-05-05
refactor: starlark venv assembly engine

xangcastle:
This PR introduces a pure-Starlark venv assembly engine (assemble_venv) that constructs Python virtualenvs entirely at analysis time using ctx.actions.symlink and ctx.actions.write. This replaces the previous runtime Rust-based venv construction with a declarative, Bazel-native approach.

## **What's new**
- py/private/venv.bzl — Core assemble_venv() helper. Resolves wheel collisions, emits two-hop symlinks for wheels with PyWheelsInfo, generates .pth files for first-party code, produces pyvenv.cfg, activation script, and console-script wrappers. bin/python is emted as an unresolved symlink with an explicit relative target_path computed from the venv up to the runfiles root and down through the interpreter's rlocation path. This avoids Bazel-version-dependent absolute-path symlinks that break sys.base_prefix inside containers.
- py/private/_virtualenv.py & py/private/venv_activate.tmpl.sh — Supporting files for venv activation and sitecustomize behaviour.
- py/private/mtree_preserve_symlinks.bzl & py/private/modify_mtree.awk — Fork of mtree_spec logic that rewrites symlink-shaped rows to type=link so py_image_layer preserves unresolved symlinks in tar layers instead of following them.
- PyWheelsInfo provider — Added to py/private/providers.bzl and propagated through py_library, py_unpacked_wheel to carry wheel metadata (top-levels, console scripts, namespace packages) for collision detection and console-script generation.

## **What's changed**
- py/private/py_venv/py_venv.bzl — Rewritten to use assemble_venv. The py_venv rule now builds venvs at analysis time. Legacy py_venv_binary / py_venv_test are preserved as backward-compatible wrappers (to be removed in PR5).
- py/private/py_venv/entrypoint.tmpl.sh — Updated to use ARG_VENV_PYTHON instead of VENV.
- py/private/py_venv/types.bzl — VirtualenvInfo now carries bin_python, venv_name, imports, wheels.
- py/private/py_semantics.bzl — _resolve_toolchain now returns freethreaded flag (needed for freethreaded interpreter layout).
- py/private/py_unpacked_wheel.bzl — Added top_levels, console_scripts, namespace_top_levels attrs; emits PyWheelsInfo.
- py/private/py_library.bzl — Added make_wheels_depset, propagates PyWheelsInfo.
- py/private/py_image_layer.bzl — Uses mtree_preserve_symlinks to handle unresolved symlinks in container layers.
[...]

PR #972, opened 2026-05-06
refactor: py_binary / py_test venv integration

xangcastle:
This PR connects the Starlark venv assembly engine (introduced in PR #965) to py_binary and py_test. Every Python binary now gets its own internal venv built entirely at analysis time via assemble_venv, replacing the previous runtime Rust-based venv construction. It also introduces external_venv for sharing venvs across targets and expose_venv for IDE-friendly .venv symlinks.

## Breaking changes                                                                                                                                                                                                                   
                                                                      
  1. **`py_binary_rule` and `py_test_rule` removed from `//py:defs.bzl`.** The rule-level escape hatches are gone; callers must use the `py_binary` / `py_test` macros. The one in-tree user                                            
  (`py/tests/external-deps/custom-macro/macro.bzl::click_cli_binary`) was migrated.
                                                                                                                                                                                                                                        
  2. **`py_venv` semantics changed.** On main, `py_venv = py_venv_link` (write a workspace-local symlink to a venv tree). On this branch, `py_venv` is a separate rule that emits `VirtualenvInfo`. Existing `py_venv(name = "...", srcs
   = [...], ...)` callsites that expected the symlink-writing behaviour need to switch to `py_venv_link`.
                                                                                                                                                                                                                                        
  3. **The `external_venv` attribute is not part of the public API.** It exists on the underlying rule (`py_venv_exec`) for the macro to wire the sibling venv in, but isn't documented for direct use. Use `expose_venv = True` if you 
  want to control or share the venv.                                                                                                                                                                                                    
  
  4. **`venv_dir_basename` removed from `py_binary` / `py_test` macros.** The on-disk venv tree is always `.{name}.venv/`. Users who pinned a custom basename must declare an explicit `py_venv` and pair via `expose_venv` /           
  `py_venv_link`.                                                                                                                                                                                                                       
  
  5. **Public macros always emit a sibling target.** Even with `expose_venv = False`, a private `:_{name}_venv` exists in the package (visibility-restricted to the package, `tags = ["manual"]`). Wildcard target patterns (`:all`,    
  `...`) still skip it because of `manual`, but it shows up in `bazel query 'siblings(...)'`. With `expose_venv = True` the sibling becomes the public `:{name}.venv`.                                                                  
  
  6. **`package_collisions`, `include_system_site_packages`, `include_user_site_packages` no longer exist on `py_venv_exec`.** They're now venv-only attrs (the macro routes them to the sibling `py_venv`). Anyone who would have set  
  these directly on the rule is also affected by (1) — the path is closed off.                                                                                                                                                          
   
  7. **`py_test(pytest_main = True)` now appends `//py/private:default_pytest_main` to deps and writes a `:{name}.pytest_paths` args file** as part of the macro. Behaviourally the same for users; mechanically a new sibling target is
   created.                                                                                                                                                                                                                             
   
  8. **Internal rule rename** (low-impact, but worth noting): `//py/private:py_binary.bzl` exports `py_venv_exec` and `py_venv_exec_test`, not `py_binary` / `py_test`. The file itself still lives at `py_binary.bzl` (rename deferred 
  via TODO to avoid in-flight branch conflicts).                                                                                                                                                                                        
   
  ## Macro-time main resolution caveat                                                                                                                                                                                                  
                                                                      
  `py_binary` / `py_test` keep the rules_python-style fallback for an absent `main =` — if `srcs` has one entry, that's the main; otherwise look for a suffix-match against `<name>.py`. This used to run at rule-impl time on resolved 
  files; it now runs at macro-eval time on label strings. If `main` would resolve to a file produced by another rule (e.g. a `genrule` whose output is `<name>.py`), the macro can't see that output basename — pass `main =`           
  explicitly. Documented in the `py_binary` macro docstring.

  
## **What's NOT included (future PRs)**

- No Rust tooling removal yet (PR3)
- No container-specific fixes beyond what PR1 already provided (PR4)
- No breaking changes to py_venv_binary / py_venv_test legacy wrappers which still work (PR5)
- No stable API path graduation (PR6)


jbedard:
@codex

py/private/py_binary.bzl
```diff
@@ -15,111 +23,110 @@ def _dict_to_exports(env):
...
         PyInfo(
-            imports = imports_depset,
-            transitive_sources = srcs_depset,
+            # Surface the venv's imports through PyInfo so downstream
+            # consumers (e.g. py_pex_binary's `--sys-path=`) see the
+            # same sys.path the launcher will run with.
+            imports = vinfo.imports,
+            # No `srcs` / `deps` on this rule — first-party sources
+            # live on the sibling venv. Keep the depset empty here.
+            transitive_sources = depset(),
```
jbedard: Is this ok?

py/private/py_pex_binary.bzl
```diff
@@ -38,6 +38,16 @@ exclude_paths = [
...
+    # py_binary assembles a synthetic venv at <pkg>/.<name>.venv/ with a
+    # lib/python<MAJ>.<MIN>/site-packages/ subtree holding the .pth file
+    # that tells the interpreter where wheels live. These files are
+    # launcher-side plumbing and should not be packed into the PEX — pex
+    # discovers wheels itself from PyInfo.imports / --dep. Matching the
+    # inner "/site-packages/" path component excluding actual wheels is
+    # tricky, so we match on the outer "/.<name>.venv/" hidden-dir
+    # segment that only our synthetic venv tree produces. Wheel payload
+    # paths never contain a "/.<something>.venv/" segment.
+    ".venv/",
```
jbedard, line 44: What is this?

py/private/run.tmpl.sh
```diff
@@ -1,4 +1,15 @@
 #!/usr/bin/env bash
+# Launcher for py_binary targets.
+#
+# A real venv tree (pyvenv.cfg + bin/python symlink + lib/.../site-packages/<name>.pth)
```
jbedard: Do we need this comment repeated in every single .sh?

py/private/py_venv/py_venv.bzl
```diff
@@ -387,42 +400,71 @@ def _split_kwargs_for_venv(kwargs):
...
-    emitted `:{name}.venv` is a first-class public target: shareable
-    via `external_venv` from other `py_binary` / `py_test` callsites,
-    and `bazel run :{name}.venv`-able to drop into the hermetic
-    interpreter for an interactive session.
+    * `expose_venv = True` — emits a public `:{name}.venv` py_venv
+      target. Runnable (`bazel run :{name}.venv` drops into the
+      hermetic interpreter) and pairable with `py_venv_link` for IDE
+      integration. The venv inherits the binary's visibility.
+
+    * `expose_venv = False` (default) — emits a private
```
jbedard: No need to document the "private"

py/private/py_binary.bzl
jbedard: Was an empty `depset()`; now reads `vinfo.transitive_sources` from the sibling `py_venv` (added a `transitive_sources` field on `VirtualenvInfo`) so downstream consumers see the same source closure they would if `srcs`/`deps` lived on the binary.

py/private/py_pex_binary.bzl
jbedard: The sibling `py_venv` writes a tree at `<pkg>/.<name>.venv/` that ends up in the binary's runfiles. That tree is launcher plumbing (`pyvenv.cfg`, `<name>.pth`, `bin/python` symlink, console-script wrappers) — pex builds its own dep tree from `--dep`, so we skip these paths. Trimmed the comment.

py/private/run.tmpl.sh
jbedard: Removed the new top-of-file block; only the existing `NB:` about `sh:toolchain_type` remains.

py/private/py_venv/py_venv.bzl
jbedard: Trimmed: the `expose_venv = False` bullet and the inline "Private sibling: …" comment are gone. Docstring now only describes `expose_venv = True`.

py/tests/py-venv-enable-site/BUILD.bazel
```diff
@@ -10,6 +13,25 @@ py_venv_test(
...
+py_test(
+    name = "venv_test",
+    srcs = [
+        "test.py",
+    ],
+    expose_venv = True,
+    imports = [
+        ".",
+    ],
+    include_system_site_packages = True,
```
thesayyn: maybe drop the second test here.

e2e/cases/venv-bin-scripts-423/BUILD.bazel
```diff
@@ -10,3 +10,16 @@ py_venv_test(
         "@pypi//dice",
     ],
 )
+
+py_test(
+    name = "venv_test",
```
thesayyn: naming here is confusing, it would be nice if this was named `test_isolated`

py/tests/py-venv-enable-site/BUILD.bazel
jbedard: Dropped — already removed from `py/tests/py-venv-enable-site/BUILD.bazel`.

e2e/cases/venv-bin-scripts-423/BUILD.bazel
jbedard: Renamed `venv_test` → `test_isolated`.

jbedard (approved):

PR #977, opened 2026-05-06
feat: shared-hub namespacing + ambiguous-label disambiguation

gregmagolan:
## Summary

Adds shared-hub support to `aspect_rules_py//uv` — multiple `uv.project()` declarations can target the same hub, including with same-named `[dependency-groups]` (both projects can define `prod`, etc.) — without the analysis-time `Conflict on configuration name` hard-fail.

The hub publishes packages under two label shapes:

- **Unqualified `@<hub>//<package>`** — resolves at `select()` time on the active `dep_group`. The hub deduplicates providers by `(group, version)`: when multiple projects provide the package in the same group at the same version, the alias collapses to a single canonical arm and resolves cleanly. Only a true version conflict — same group, different versions — surfaces Bazel's "multiple keys match", at which point reach for the qualified shape.
- **Project-qualified `@<hub>//project/<name>:<package>`** — always available, scoped to one project's groups. Used to disambiguate a real version conflict, or when overrides need to apply per-project. The `project/` prefix is reserved on **labels** because the package side comes from PyPI, which the user doesn't control — without the prefix, picking a project name that matched a PyPI package would silently shadow it.

The `dep_group` flag has four value shapes (no `project/` prefix on the flag-value side, since both axes there are user-controlled):

- `""` (empty) — the flag's `build_setting_default`. Matches the synthesized empty-keyed group for projects without `[dependency-groups]`. Single-project hubs resolve zero-config; PEP 735 forbids empty group names so this can never collide with a user-declared group.
- `<group>` — broad. Activates the named group across **every** project in the hub that declares it. `dep_group=prod` activates both `web`'s and `worker`'s `prod` groups simultaneously.
- `<name>` — synthesized for projects without `[dependency-groups]`, alongside `""`. Setting `dep_group="<name>"` activates exactly one project's resolution in a multi-project hub, even when the synthesized empty group would otherwise collapse multiple projects into one shared group.
- `<name>/<group>` — narrow. Activates only the named project's `<group>`, leaving every other project's same-named group untouched. The flag-value layer equivalent of the qualified hub label, useful when targets in a multi-project hub need to disambiguate without rewriting every dep.

Project names and dependency-group names share one user-controlled namespace at the flag-value layer. The extension fails at hub construction time if a project's stamp matches a group name declared by any other project in the same hub — both names are locally adjustable, so collisions are user error and surfaced loudly.

### What changed

- `extension/projectfile.bzl`: when `[dependency-groups]` is absent, synthesize `{"": ..., <name>: ...}` both mapping to the lock manifest members.
- `extension/defs.bzl`:
  - hub_cfgs reshaped from flat `{configurations, packages}` to project-keyed `{projects, package_owners}`. Per-project package map tracks sorted unique versions per `(group, package)` to drive the version-equality dedup.
  - Removed the `Conflict on configuration name` hard-fail.
  - Added a stamp-vs-group-name collision check that runs after parsing all projects in a hub.
  - `default_build_dependencies` BFS activation extracted into `_activate_default_build_deps`.
- `uv_hub/repository.bzl`: full rewrite of BUILD generation —
  - `//dep_group:<stamp>__<group>` config_settings replace flat per-cfg settings; both projects' `prod` coexist as e.g. `web__prod` / `worker__prod`. A narrow `<stamp>__q__<group>` config_setting also fires for the qualified flag value `<stamp>/<group>` (skipped for `""` and the synthesized `<stamp>` alias, which the bare flag values already handle).
  - `//project/<stamp>/BUILD.bazel` emits per-project package aliases scoped to that project's groups; cross-project ambiguity is impossible by construction.
  - `//<package>/BUILD.bazel` groups providers by `(group, versions_tuple)` and emits one canonical arm per cluster. Same-version providers in the same group collapse to one arm; version conflicts produce multiple arms that multi-match (and Bazel surfaces the conflict) when the active `dep_group` activates them. Each provider also contributes a per-project narrow arm so `dep_group="<owner>/<group>"` resolves unambiguously regardless of overlap.
  - `gazelle_index_whls` selects per-project so only the active project's whls flow through, avoiding forced analysis of every project's sdist build tools at index time.
  - `compatible_with` / `incompatible_with` helpers fan out a bare group name to every project's namespaced config_setting.
- `uv_project/repository.bzl`: per-project internal `//private/dep_group:<cfg>` config_settings now include both broad and narrow forms, mirroring the hub. Empty `cfg` ("") maps to a `_default` Bazel target name (since target names can't be empty), while the flag_value stays `""`.
- `uv/private/constraints/dep_group/BUILD.bazel`: flag default `""`. Matches the synthesis-fallback empty-keyed group; eliminates the need for a `.bazelrc` entry in the common case.

### Caveat: dedup is on `(group, version)` only

Two projects pinning the same version but applying different `uv.override_package` overrides (`target=`, `post_install_patches`) produce differing wheel content. The canonical-arm choice is deterministic (lex-first project by stamp) but only one project's overrides apply through the unqualified label. Users with per-project overrides on shared packages should reach for the qualified label or the narrow `dep_group="<name>/<group>"` form.

### `default_build_dependencies` activation

Build-system-only deps declared via `uv.project(default_build_dependencies = [...])` aren't reachable from the project's runtime activation set, so `@<project_id>//:setuptools` / `@<project_id>//:build` aliases that the sdist build_tool's deps reference need explicit activation. After `collect_activated_extras`, `_activate_default_build_deps` BFSes from each entry across the marker_graph and activates every reachable node into every cfg of the project. This puts build deps and their transitive runtime closure into `dep_to_scc`, so the project repo emits the expected aliases and the sdist build_tool can be built end-to-end.

The augmentation is gated to projects WITHOUT explicit `[dependency-groups]`. Projects with explicit groups own their build-tool version pinning, and forcing an additional default-versions activation can conflict with a per-group pin (e.g. `[tool.uv.conflicts]` declaring two mutually-exclusive groups at different `build` versions, as in `cases/unconstrained-dependencies`). Those projects are expected to declare build tools inside their groups.

### E2E updates

- 25 e2e `pyproject.toml`s left unchanged (synthesis fallback handles them via `""` + `<name>`).
- ~28 e2e BUILDs updated from `dep_group = "<project_name>"` (the pre-PR2 implicit naming) to `dep_group = "<normalized_project_name>"` (the PEP 503-normalized synthesized alias). For projects with hyphens in their `[project].name`, this means hyphen → underscore in the `dep_group` value.
- 2 BUILDs that consume hub deps from a non-transitioning context use a project-qualified label: `cases/uv-whl-install-output-group/BUILD.bazel` (a standalone filegroup), `cases/uv-workspace-789/packages/pi/BUILD.bazel` (a py_library wildcard-built).
- `cases/uv-gazelle-778/BUILD.bazel` references the bare `<name>` synthesized aliases plus the `extras` explicit group.

### New tests

- **`e2e/cases/uv-shared-hub-848/`** (7 tests) — two projects (`proj-a-848`, `proj-b-848`) sharing a hub `pypi_shared_848`. Each defines a `shared` group (overlapping, same version) and a project-unique `<x>_only` group. Exercises:
  - Qualified labels (`@hub//project/proj_a_848:pytest`) under `dep_group=shared` regardless of overlap.
  - Unqualified label `@hub//pytest` under `dep_group=shared` resolving via the version-equality dedup canonical arm.
  - Unqualified labels under `dep_group=a_only` / `b_only`, where exactly one project provides the package in the active group.
  - Narrow flag form `dep_group=proj_a_848/shared` and `proj_b_848/shared` activating one project's `shared` group only.
- **`e2e/cases/uv-synthesis-default-848/`** (4 tests) — single project with NO `[dependency-groups]` in its `pyproject.toml`. Bound to a separate hub `pypi_synth_848` to keep the synthesis path isolated. Exercises:
  - Implicit `dep_group` resolution via the flag default `""` (no `dep_group=` attr, no `.bazelrc` entry).
  - Explicit `dep_group=""` per-target.
  - Explicit `dep_group="synth_default_848"` per-target (the bare project-name alias).
  - Project-qualified label `@hub//project/synth_default_848:pytest`.
- **`py/tests/uv-qualified-label/`** (2 tests) — main-repo regression on the `//project/<name>/BUILD.bazel` emitter against the `@pypi` hub.

### Documentation

`docs/uv.md` rewritten where it referenced the old "venv = configuration" terminology — now consistently uses "dep_group". Adds:

- A "Flag-value summary" table at the top of "Dependency groups" listing the four flag-value shapes (`""`, `<group>`, `<name>`, `<name>/<group>`).
- "Default behavior when `dep_group` is unset" subsection covering the empty-default and per-target override.
- "Comparison with uv outside Bazel" mapping `uv sync` flags to the Bazel `dep_group` equivalents.
- "Patterns for overlay behavior" showing self-reference and `include-group` patterns for users wanting `[project].dependencies` overlay semantics.
- "Synthesis fallback: empty and `<name>` aliases" explaining the synthesized aliases and the cross-project collision check.
- "Per-project narrow activation: `<name>/<group>`" covering the narrow flag form.
- "Authoring a library that's published *and* used in a Bazel build" — guidance for the common library shape.
- "Sharing a hub across multiple projects" covering both label shapes and dedup behavior, including the asymmetric reasoning for the `project/` prefix on labels (uncontrolled package namespace) versus its absence on flag values (fully user-controlled).

### Suggested release notes

- (`uv`): `uv.project()` may now share a `hub_name` with other projects, including ones that declare `[dependency-groups]` of the same name. The hub presents both unqualified `@<hub>//<package>` and project-qualified `@<hub>//project/<name>:<package>` labels — the unqualified shape resolves cleanly when providers in the active group all pin the same version, and Bazel surfaces "multiple keys match" only on a real version conflict.
- (`uv`): Projects without `[dependency-groups]` in their `pyproject.toml` now get a synthesized `<name>` alias alongside the empty-keyed group `""`. Setting `dep_group="<name>"` activates exactly one project's resolution in a multi-project hub regardless of cross-project version diversity.
- (`uv`): the `dep_group` flag accepts a narrow `<name>/<group>` form that activates exactly one project's `<group>`, leaving every other project's same-named group untouched. The flag-value layer equivalent of the qualified hub label.
- (`uv`): the `dep_group` flag now defaults to `""` (empty), matching the synthesis fallback's empty-keyed group. Single-project hubs without `[dependency-groups]` resolve hub labels with no `.bazelrc` configuration.
- BREAKING (`uv`): When a `pyproject.toml` lacks a `[dependency-groups]` table, the extension synthesizes the empty-keyed group `""` plus a `<name>` alias instead of a group named after the project. Migrate by either adding an explicit `[dependency-groups]` table to the project's `pyproject.toml`, switching `dep_group="<project_name>"` references to use the PEP 503-normalized form (hyphens → underscores), or letting the empty default take over.
- BREAKING (`uv`): the `Conflict on configuration name` hard-fail for two projects in one hub registering the same group name is removed. A new collision check fails if any project's stamp matches a group name declared by another project in the same hub.

### Test plan

- `bazel test //...` in main repo: 138 pass.
- `bazel test //...` in e2e: 84 pass / 22 skipped (skips are platform-gated).


e2e/cases/firebase-admin-import/pyproject.toml
```diff
@@ -15,3 +15,7 @@ dependencies = [
     "setuptools",
     "firebase-admin==6.6.0",
 ]
+
+
+[dependency-groups]
```
jbedard: Why is this added everywhere?

gregmagolan: Defensive change before an improvement. Reverting them now.

docs/uv.md
```diff
@@ -242,75 +376,94 @@ platform_transition_filegroup(
...
 ```
 
 ## A mental model
 
 ```
-@pypi                                     # Your UV built hub repository
-@pypi//requests:requests                  # The library for a requirement
+@pypi                                     # Your UV-built hub repository
+@pypi//requests:requests                  # Unqualified library label
+@pypi//project/web:requests                      # Project-qualified label (when needed)
```
jbedard: comment alignement

docs/uv.md
```diff
@@ -134,6 +141,133 @@ py_binary(
...
+```
+
+```starlark
+# apps/web/BUILD.bazel
+py_binary(
+    name = "bin",
+    srcs = ["main.py"],
+    dep_group = "prod",
+    deps = [
+        "@pypi//flask",                # unqualified — only `web` provides it
```
jbedard: alignment 🤷

docs/uv.md
```diff
@@ -134,6 +141,133 @@ py_binary(
...
+    ],
+)
+
+# apps/worker/BUILD.bazel
+py_binary(
+    name = "bin",
+    srcs = ["main.py"],
+    dep_group = "prod",
+    deps = [
+        "@pypi//celery",               # unqualified — only `worker` provides it
```
jbedard: alignment 🤷

uv/private/extension/defs.bzl
```diff
@@ -448,24 +448,47 @@ def _parse_projects(module_ctx, hub_specs):
...
+                # pin the same version; that's the dedup case. True version
+                # conflicts surface as Bazel "multiple keys match" only at
+                # the active dep_group, exactly when they're real.
+                package_owners = {},
             ))
 
-            for cfg in configuration_names.keys():
-                if cfg in hub_cfg.configurations:
-                    fail("Conflict on configuration name {} in hub {}".format(cfg, project.hub_name))
+            if project_id in hub_cfg.projects:
```
jbedard, line 717: Didn't we just set `projects = {}`?

uv/private/extension/defs.bzl
```diff
@@ -448,24 +448,47 @@ def _parse_projects(module_ctx, hub_specs):
...
+            # activated under multiple markers in a single group, hence
+            # `set()` to dedupe. The version list is what drives
+            # version-equality dedup of the unqualified hub label.
+            project_packages = {}
             for package, cfgs in version_activations.items():
-                for cfg in cfgs.keys():
-                    hub_cfg.packages.setdefault(package, {})[cfg] = "@{}//:{}".format(project_id, package)
+                # Starlark has no set comprehension; use dict-keys trick to dedupe.
+                project_packages[package] = {
+                    group: sorted({dep_id[2]: True for dep_id in dep_ids.keys()}.keys())
```
jbedard: that random `2` needs an explanation...

uv/private/extension/defs.bzl
```diff
@@ -448,24 +448,47 @@ def _parse_projects(module_ctx, hub_specs):
...
+                    for group, dep_ids in cfgs.items()
+                }
+                owners = hub_cfg.package_owners.setdefault(package, [])
+                if project_id not in owners:
+                    owners.append(project_id)
+
+            hub_cfg.projects[project_id] = struct(
+                # Used as the //.<stamp>:<pkg> qualified label component.
+                # Already normalize_name'd above (line ~160), so it's a
```
jbedard: Lets not reference line numbers in comments

jbedard: This full comment block is probably useless actually?

uv/private/extension/defs.bzl
```diff
@@ -604,10 +627,21 @@ def _uv_impl(module_ctx):
         )
 
     for hub_id, hub_cfg in cfg.hub_cfgs.items():
+        # Serialize the per-project view as a single JSON blob — repository
```
jbedard: Lets not explain the reasoning behind the `uv_hub` attributes here, just explain why we are converting a struct to a plain dict (the second sentence is all we need I think).

jbedard:
I added [5e673fb](https://github.com/aspect-build/rules_py/pull/977/commits/5e673fb5e3c7cf28bd5aca0669175e60ad27bdd9) reverting some random things that seemed unnecessary/impossible. Test failures seem the same as before that so I think something is wrong, claude claims the failures are due to:

>   uv-sdist-native-build's pyproject.toml has no [dependency-groups], so it hits the synthesis fallback. The synthesized dep_groups (default, project/uv-sdist-native-build) only contain the lock manifest members — python-geohash and the project itself. They don't include the build-system tools (setuptools, build) needed to build python-geohash's sdist.

and we should:

>   In projectfile.bzl, the synthesis fallback's fallback_members needs to include sdist-build transitive deps too — or alternatively the project-hub generator should always emit aliases for every package in the lock graph, not just for those in some dep_group. Spot-check by comparing what dep_groups = {project_name: lock_manifest_members} evaluated to pre-PR vs. what {default, project/<name>} produces now.

uv/private/_ambiguous_dep/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+# Helpers for the hub's ambiguous-package stubs.
```
jbedard: Is this file no longer used?

docs/uv.md
gregmagolan: ✅

docs/uv.md
gregmagolan: ✅

docs/uv.md
gregmagolan: ✅

uv/private/extension/defs.bzl
gregmagolan: ✅

uv/private/extension/defs.bzl
gregmagolan: Claude 😡

gregmagolan: ✅

uv/private/extension/defs.bzl
gregmagolan: ✅

uv/private/_ambiguous_dep/BUILD.bazel
gregmagolan: File is now deleted now

docs/uv.md
```diff
@@ -241,15 +589,20 @@ platform_transition_filegroup(
 ## Example: Constraining library compatibility
 
 By default uv hubs let you write `py_library` and other targets which are
-compatible with _any_ dependency group providing all the needed requirements.
+compatible with _any_ dependency group providing all the needed
```
jbedard, line 633: tell claude to reduce unnecessary changes :|

e2e/cases/firebase-admin-import/BUILD.bazel
```diff
@@ -38,7 +38,7 @@ load("@bazel_skylib//rules:build_test.bzl", "build_test")
 py_test(
     name = "test",
     srcs = ["test.py"],
-    dep_group = "firebase-admin-import",
+    dep_group = "firebase_admin_import",
```
jbedard, line 42: Did defaults change in some way? Why did all these change?

uv/private/extension/defs.bzl
```diff
@@ -598,16 +717,28 @@ def _uv_impl(module_ctx):
...
+                "groups": p.groups,
+                "packages": p.packages,
+            }
+            for project_id, p in hub_cfg.projects.items()
+        }
         uv_hub(
             name = hub_id,
-            configurations = hub_cfg.configurations,
-            packages = json.encode(hub_cfg.packages),
+            projects = json.encode(projects_json),
```
jbedard, line 934: Not in this PR, but I'm curious... would writing to a file not be more efficient? A file write vs giant-arguments for bazel caching etc? 🤔

jbedard:
Can we add some snapshot tests? Some `write_source_files` that output the generated .bzl for some setups. Then when we do refactors/fixes/test-additions we can see the diffs in the generated code - I find it very useful for all the generated .bzl in rules_js

I'm not sure if it would be useful to do this in a pre-commit and then this PR will show the diff? I assume the diff will be too huge to be useful with such a big refactor but maybe I'm wrong?

EDIT(10:40pm): the more I review this PR the more I think this is a good idea - then you can test cleanup+perf+fixes of the code generator and along with your change you also see the diff of the generated codewhich is extremely useful when dealing with generated code.

EDIT: I added some in https://github.com/aspect-build/rules_py/pull/990

uv/private/extension/defs.bzl
```diff
@@ -448,24 +523,65 @@ def _parse_projects(module_ctx, hub_specs):
...
+    # namespace with every group name in the hub. If a different project
+    # in the same hub declares a group with the same name as project P's
+    # stamp, `dep_group=<stamp>` simultaneously activates P's synthesized
+    # alias AND the other project's group — same broad semantics, but the
+    # mental model breaks. Both namespaces are user-controlled and
+    # locally adjustable, so we fail loudly and let the user rename one.
+    for hub_id, hub_cfg in hub_cfgs.items():
+        stamp_to_pid = {p.stamp: pid for pid, p in hub_cfg.projects.items()}
+        for pid, p in hub_cfg.projects.items():
+            for grp in p.groups:
```
jbedard, line 755: I find these nested loops scary so I asked claude how this can be improved, I pushed https://github.com/aspect-build/rules_py/pull/977/changes/3faaee8a4f5eea537890e1576faabfd96acc7d42 to your branch - feel free to edit/revert/squash that commit

uv/private/extension/defs.bzl
```diff
@@ -598,16 +714,28 @@ def _uv_impl(module_ctx):
...
             dep_to_scc = json.encode(project_cfg.dep_to_scc),
             scc_deps = json.encode(project_cfg.scc_deps),
             scc_graph = json.encode(project_cfg.scc_graph),
         )
 
     for hub_id, hub_cfg in cfg.hub_cfgs.items():
+        # Repository rules can't take nested string_dicts; flatten to JSON.
+        # Struct fields are pulled into plain dicts since json.encode
+        # doesn't traverse structs.
+        projects_json = {
```
jbedard, line 924: Theoretically making these dictionaries means they are non-deterministic, which also means we need to sort them when deserializing.

Since we're restructuring everything into plain dicts() here anyway, should this be where we make it a sorted array instead of dict()?

uv/private/uv_hub/repository.bzl
```diff
@@ -38,182 +100,260 @@ alias(
...
     visibility = ["//visibility:public"],
 )
-""".format(name = name),
-        )
+""".format(stamp = stamp, grp = grp))
+
+            # Narrow arm: `dep_group=<stamp>/<group>`.
+            if grp != "" and grp != stamp:
+                content.append("""
+config_setting(
```
jbedard, line 127: Is it worth putting this into a macro?

In rules_js we intentionally put as few string literals as possible in the per-npm-package generated files due to how bazel does string interning/management. So instead we create a macro that would be something such as:

```
def stamp_grp_config_setting(stamp, grp):
    config_setting(
        name = "{stamp}__q__{grp}".format(stamp, grp),
        flag_values = {{
            "@aspect_rules_py//uv/private/constraints/dep_group:dep_group": "{stamp}/{grp}".format(stamp, grp),
        }},
     )
```

Then that reduces it from 4 string literals per generated `config_setting()` to 2, and those 2 are `{stamp}` and `{grp}` which are probably used other times int his file. It really all depends how many of these files will be generated.

uv/private/extension/defs.bzl
jbedard: Looks like you [overwrote that](https://github.com/aspect-build/rules_py/compare/3faaee8a4f5eea537890e1576faabfd96acc7d42..9f908582c17514810e0f2a4066c989e647b27694#diff-1183dcf0fd627c704c6c18f2535d29e266a1c5f9ce680a90d084d5f8e05841b0R572) and made it go back O(P+G) → O(P²·G) in complexity

uv/private/uv_hub/repository.bzl
jbedard: For reference: https://github.com/aspect-build/rules_js/issues/2138

jbedard: That also puts less logic into ugly string templates and into these simple internal wrapper macros.

For example see some of the [templates in rules_js](https://github.com/aspect-build/rules_js/blob/315b6dd6d39cfb980b10b486d8d16db0977ef42d/npm/private/npm_import.bzl#L337-L355) where its just a single call to a macro with as few string literals as possible.

jbedard: TLDR: don't repeat the same strings in multiple generated files (especially across repos like npm packages do)

uv/private/uv_hub/repository.bzl
```diff
@@ -38,182 +100,260 @@ alias(
...
-        content.append(
-            """
-# This target is for a "hard" dependency.
-# Dependencies on this target will cause build failures if it's unavailable.
+        for package, group_versions in p["packages"].items():
+            groups = group_versions.keys()
+            target = "@{}//:{}".format(project_id, package)
+            select_arms = {}
+            compat_arms = {}
+            for grp in groups:
```
jbedard, line 198: More `group X package X project` loops. Note these also do `.items()` which scares me in terms of memory/GC.

This again makes me think this data structure should change...

uv/private/extension/defs.bzl
jbedard: Pushed again: https://github.com/aspect-build/rules_py/pull/977/commits/95ef83903470bd0e05eddd40952bb1cd0dccf00d

docs/uv.md
gregmagolan: Claude 😡

uv/private/extension/defs.bzl
jbedard: You overwrite it again...

uv/private/extension/defs.bzl
dzbarsky: Starlark dicts preserve insertion order (very sad property when implementing an interpreter)

uv/private/uv_hub/repository.bzl
dzbarsky: A reasonable interpreter can make a loop over .items() unpack in place and avoid allocating intermediates...

uv/private/extension/defs.bzl
```diff
@@ -70,6 +70,69 @@ load(":graph_utils.bzl", "activate_extras", "collect_sccs")
...
+            continue
+        for it, _marker in extract_requirement_marker_pairs(projectfile, lock_id, req, default_versions, package_versions):
+            starting_points.append(it)
+
+    visited = {}
+    for sp in starting_points:
+        visited[sp] = 1
+    worklist = list(starting_points)
+    bfs_idx = 0
+    for _ in range(1000000):
```
xangcastle, line 176: I understand this like a defensive pattern, but in a masive lockfile with ~10k packages is this enough?

uv/private/extension/defs.bzl
```diff
@@ -70,6 +70,69 @@ load(":graph_utils.bzl", "activate_extras", "collect_sccs")
...
+
+    Mutates `activated_extras` in place.
+    """
+    starting_points = []
+    for req in default_build_dependencies:
+        # Coarse name extraction matching extract_requirement_marker_pairs:
+        # strip at the first non-name char. Used only to pre-skip reqs
+        # missing from the lockfile (which would cause extract to fail).
+        bare_name = req
+        for sep in [";", "[", ">", "<", "=", "!", "~", " ", "@"]:
```
xangcastle, line 153: I think this is duplicated with extract_requirement_marker_pairs, maybe we can create a helper function to share this stripping logic

uv/private/extension/defs.bzl
xangcastle: I understand that if we have two projects with the same name it will be validated with a:
fail("Project {} declared more than once in hub {}".format(project_id, project.hub_name))
But, whats happend if they have different name but normalize to the same stamp, eg my-lib and my_lib, in this case they can colisionate

uv/private/extension/defs.bzl
```diff
@@ -264,6 +325,19 @@ def _parse_projects(module_ctx, hub_specs):
             whl_configurations.update(collect_configurations(lock_data))
 
             configuration_names, activated_extras = collect_activated_extras(project.lock, project_id, project_data, lock_data, default_versions, marker_graph, package_versions)
+
+            if "dependency-groups" not in project_data and project.default_build_dependencies:
```
xangcastle, line 425: the condition and project.default_build_dependencies use truthiness list, if the user pass default_build_dependencies = [] this will not be activated

.bazelrc
```diff
@@ -6,8 +6,9 @@ common --incompatible_disallow_empty_glob=False
 # Define value used by tests
 common --define=SOME_VAR=SOME_VALUE
 
-# Set the default virtualenv to 'default'
-common --@pypi//dep_group=aspect_rules_py
+# dep_group=default is the flag's default; no explicit setting needed
```
jbedard, line 9: Lets just delete this comment, it makes no sense on its own

PR #984, opened 2026-05-07
chore: fix gazelle, enforce on ci

jbedard:
### Test plan

- New test cases added

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,12 +1,21 @@
+load("@aspect_gazelle_prebuilt//:def.bzl", "aspect_gazelle")
 load("@buildifier_prebuilt//:rules.bzl", "buildifier")
-load("@gazelle//:def.bzl", "gazelle")
 load("@rules_python//python/pip_install:requirements.bzl", "compile_pip_requirements")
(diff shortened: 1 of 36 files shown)


BUILD.bazel
```diff
@@ -17,9 +23,10 @@ gazelle_python_manifest(
     ],
 )
 
-gazelle(
+aspect_gazelle(
     name = "gazelle",
-    gazelle = "@multitool//tools/gazelle",
+    languages = ["starlark"],
```
jbedard, line 31: Note that previously the `python` language was enabled by default and it has clearly not been run in a **long** time so I simply deleted it.

Within `rules_py[thon]` we do weird things in BUILDs for the purpose of tests etc, it's not really a python repo with standard python rules in it, so I think we're better off not generating `py_*` targets by default.

However our release scripts do query for `bzl_library` targets to generate stardoc, so I think we should maintain proper `bzl_target`s. Note this root BUILD does `#gazelle:exclude` on most directories other then `//uv` and `//py` so it won't generate `bzl_library` **everywhere** still.

gregmagolan (approved):

PR #989, opened 2026-05-08
test: various config between py_venv() and py_venv_exec() targets

jbedard:
While `py_venv_exec` is not public ATM I think it is worth testing all the different combos+scenarios possible between `py_venv` and the execution of stuff from it.

### Test plan

- New test cases added


jbedard:
@codex

PR #997, opened 2026-05-11
fix(uv): treat abi3 wheels as compatible with newer CPython minors

jbedard:
### Test plan

- New test cases added


jbedard:
@codex

uv/private/whl_install/repository.bzl
```diff
@@ -213,6 +214,40 @@ def sort_select_arms(arms):
...
     return {a: b for a, b in pairs}
 
+def compatible_python_tags(python_tag, abi_tag):
+    if abi_tag != "abi3" or not python_tag.startswith("cp"):
+        return [python_tag]
+
+    major = int(python_tag[2])
+    minor = int(python_tag[3:]) if python_tag[3:] else 0
+    compatible = []
+    for candidate_major in MAJORS:
```
xangcastle, line 224: TODO: This iterates over hardcoded MAJORS/MINORS. When Python 3.21+ 
eventually ships, the constraints system will need a broader update 
across rules_py. I need to think a way to centralizing version caps 
in a single source of truth.

uv/private/whl_install/repository.bzl
```diff
@@ -253,23 +288,35 @@ def _whl_install_impl(repository_ctx):
         "load(\"@bazel_skylib//lib:selects.bzl\", \"selects\")",
     ]
 
+    # During expansion the value is (source_specificity, target). When two
+    # abi3 wheels claim the same (compatible_python, platform, abi) key —
+    # e.g. cp38-abi3 and cp311-abi3 both expand into cp312+ — the wheel
+    # with the higher source minor must win, regardless of iteration order.
     for whl, target in prebuilds.items():
```
xangcastle, line 295: TODO: sort_select_arms() emits cp320, cp319, cp318... all pointing 
at the same cp311-abi3 wheel for a py3.12 build. Arms are redundant 
but harmless; collapsing them would shrink generated BUILD files. 
Defer until someone measures a real performance win.

xangcastle (approved):

uv/private/whl_install/repository.bzl
jbedard: This was also a concern of mine, but I'm uncertain if they are always redundant or only sometimes?

xangcastle (approved):

uv/private/whl_install/repository.bzl
xangcastle: I think it's just how abi3 wheels work. If one file satisfying many tags. It’s redundant visually, but I'm worried collapsing them might mess with the resolution logic. Safe to defer until it actually affects performance. We just need to think later on this

PR #1000, opened 2026-05-11
fix(uv): suppress nonfatal pyc compile warnings

jbedard:
### Test plan

### DIFF
--- a/py/tools/unpack/src/main.rs
+++ b/py/tools/unpack/src/main.rs
@@ -189,7 +189,9 @@ fn unpack_cmd_handler(args: UnpackArgs) -> Result<()> {
             ))
             .join("site-packages");
 
-        let status = std::process::Command::new(python)
+        // Swallow compileall's non-zero exit to match uv (and pip before it):
+        // https://github.com/astral-sh/uv/blob/374fcaa56aa46a2923eed31fe0df7a4a7a14c477/crates/uv-installer/src/pip_compileall.py#L5
+        let _ = std::process::Command::new(python)
             .args([
                 "-m",
                 "compileall",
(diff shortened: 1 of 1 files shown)


dzbarsky (approved):
I suspect nobody wants these on by default, packages shipping broken py2 code isn't really actionable

jbedard:
What if that was only discovered when updating a package and the update is broken? You'd think sometimes you'd want to know?

jbedard:
@codex

dzbarsky:
> What if that was only discovered when updating a package and the update is broken? You'd think sometimes you'd want to know?

Yeah that's not something I've ever seen in over a decade of doing python, the code will fail at import time anyway. Perhaps best to match what uv itself does? ;)

jbedard:
You know explicitly what UV does? Have a link to reference? 😅

dzbarsky:
> You know explicitly what UV does? Have a link to reference? 😅

https://github.com/astral-sh/uv/blob/374fcaa56aa46a2923eed31fe0df7a4a7a14c477/crates/uv-installer/src/pip_compileall.py#L5

jbedard:
Updated to do exactly what UV does 👍

dzbarsky (approved):

py/tools/unpack/src/main.rs
```diff
@@ -189,7 +189,9 @@ fn unpack_cmd_handler(args: UnpackArgs) -> Result<()> {
             ))
             .join("site-packages");
 
-        let status = std::process::Command::new(python)
+        // Swallow compileall's non-zero exit to match uv (and pip before it):
+        // https://github.com/astral-sh/uv/blob/374fcaa56aa46a2923eed31fe0df7a4a7a14c477/crates/uv-installer/src/pip_compileall.py#L5
```
xangcastle, line 193: TODO: add --verbose-compileall flag for future debugging

xangcastle (approved):

PR #1002, opened 2026-05-11
test: separate e2e test hubs

jbedard:
I think e2e tests adding to the same hub is dangerous and should be avoided unless that is explicitly what we are testing (the new `multi-project` test).

### Test plan

- New test cases added


jbedard:
@codex

gregmagolan (approved):

xangcastle (approved):

PR #1004, opened 2026-05-12
feat(sdist): generic toolchain env plumbing for pep517_native_whl

jbedard:

pep517_native_whl now accepts an arbitrary list of toolchain targets and
    forwards their `TemplateVariableInfo` make-variables into a per-action
    `env` dict via `$(VAR)` substitution. Pattern mirrors rules_rust's
    cargo_build_script — replaces the previous hardcoded CC toolchain reach
    inside the rule body:

* `ctx.attr.toolchains`; pull `DefaultInfo.files` / `ToolchainInfo.all_files` into action inputs and `TemplateVariableInfo.variables` into a make-variable substitution map
* `env` string_dict attr whose values may reference $(VAR) make variables sourced from any toolchain in the list

### Test plan

- New test cases added


uv/private/sdist_build/repository.bzl
```diff
@@ -211,6 +211,32 @@ def _sdist_build_impl(repository_ctx):
...
+    # sources find their compiler and JDK by default. Both targets expose
+    # `TemplateVariableInfo`; the env values below are make-variable
+    # references resolved at action analysis time.
+    #
+    # CXX defaults to $(CC) — most clang/gcc-based toolchains use a single
+    # driver binary for both languages, and meson-python / cmake-based
+    # backends look for CXX independently of CC.
+    toolchain_attrs = ""
+    if is_native:
+        toolchain_attrs = """
```
jbedard, line 242: @dzbarsky this is one part I'm uncertain of, should we be avoiding defaults like this? Or is that needed for many packages to compile correctly?

uv/private/pep517_whl/rule.bzl
```diff
@@ -30,6 +27,34 @@ def _patch_args_and_inputs(ctx):
...
+
+    Pattern mirrors rules_rust's cargo_build_script
+    (see cargo/private/cargo_build_script.bzl).
+    """
+    extra_inputs = []
+    known_variables = {}
+    for target in ctx.attr.toolchains:
+        if DefaultInfo in target:
+            extra_inputs.append(target[DefaultInfo].files)
+            extra_inputs.append(target[DefaultInfo].default_runfiles.files)
```
xangcastle: `default_runfiles` can be `None` on some targets. Consider adding a guard:

```python
if DefaultInfo in target:
    extra_inputs.append(target[DefaultInfo].files)
    if target[DefaultInfo].default_runfiles:
        extra_inputs.append(target[DefaultInfo].default_runfiles.files)
```

uv/private/extension/defs.bzl
```diff
@@ -226,11 +241,19 @@ def _parse_projects(module_ctx, hub_specs):
                     ))
 
                 has_target = override.target != None
+
+                # Toolchains / env count as modifications when they differ from
+                # the rule defaults (which equal sdist_build's defaults, so a
+                # no-op override has neither effect nor anything to validate).
+                toolchains_overridden = override.toolchains != _DEFAULT_NATIVE_TOOLCHAINS
```
xangcastle: Label comparison can be fragile when one side resolves to a string. Suggest:
```python
toolchains_overridden = [str(t) for t in override.toolchains] != [str(t) for t in _DEFAULT_NATIVE_TOOLCHAINS]
```

uv/private/extension/defs.bzl
```diff
@@ -376,6 +399,19 @@ def _parse_projects(module_ctx, hub_specs):
...
 
+                    # Only thread the toolchains/env kwargs through to
+                    # sdist_build when the override differs from the
+                    # default — otherwise let sdist_build use its own
+                    # matching defaults (avoids unnecessary kwargs and
+                    # keeps the no-override path simple).
+                    toolchains_for_sbuild = None
+                    env_for_sbuild = None
+                    if pkg_override:
+                        if pkg_override.toolchains != _DEFAULT_NATIVE_TOOLCHAINS:
```
xangcastle: Same Label-vs-string comparison issue. Matching fix:
```python
if [str(t) for t in pkg_override.toolchains] != [str(t) for t in _DEFAULT_NATIVE_TOOLCHAINS]:
```

xangcastle (changes requested):

e2e/cases/uv-sdist-jdk-build/setup.MODULE.bazel
```diff
@@ -0,0 +1,43 @@
...
+    lock = "//cases/uv-sdist-jdk-build:uv.lock",
+    pyproject = "//cases/uv-sdist-jdk-build:pyproject.toml",
+)
+uv.unstable_annotate_packages(
+    src = "//cases/uv-sdist-jdk-build:annotations.toml",
+    lock = "//cases/uv-sdist-jdk-build:uv.lock",
+)
+
+# Replace the default CC toolchain plumbing with a JDK-only set. Contrived
+# for jpype1 (its real build does need CC), but exercises the
```
dzbarsky: i think this should be augment instead of replace. as you say, you might still need CC toolchain and specifying all vars for that seems super annoying

uv/private/extension/defs.bzl
```diff
@@ -70,6 +70,21 @@ load(":graph_utils.bzl", "activate_extras", "collect_sccs")
...
 load(":projectfile.bzl", "collate_versions_by_name", "collect_activated_extras", "extract_requirement_marker_pairs")
 
+# Default toolchain + env wired into the generated pep517_native_whl(...) call
+# for native sdists. Match the matching defaults on sdist_build's repo rule so
+# a no-op override_package (toolchains/env left at their default) produces the
+# same generated BUILD as no override at all. To strip CC (or replace with
+# another toolchain), pass `toolchains = [...]` and `env = {...}` explicitly —
+# an empty list / dict is honored verbatim.
+_DEFAULT_NATIVE_TOOLCHAINS = [Label("@bazel_tools//tools/cpp:current_cc_toolchain")]
+_DEFAULT_NATIVE_ENV = {
```
dzbarsky: don't think this is gonna work, you also need CCFLAGS LDFLAGS etc. I think for cc specifically you probably want to grab the actual envs for compile + link actions off the toolchains and inject those

uv/private/extension/defs.bzl
```diff
@@ -568,6 +606,14 @@ def _uv_impl(module_ctx):
         if sbuild_cfg.pre_build_patches:
             sbuild_kwargs["pre_build_patches"] = sbuild_cfg.pre_build_patches
             sbuild_kwargs["pre_build_patch_strip"] = sbuild_cfg.pre_build_patch_strip
+
+        # `None` (sentinel for "user didn't deviate from default") leaves
+        # sdist_build to use its own matching default. An empty list/dict
+        # passes through verbatim and emits empty toolchains/env.
+        if sbuild_cfg.toolchains != None:
```
dzbarsky: i think if you just augment them a lot of this becomes simpler? Is the use case of "trust me bro I have a cc toolchain registered and I DON'T want to use it for this build because I know what I'm doing" really worth jumping through these hoops

uv/private/sdist_build/repository.bzl
```diff
@@ -265,5 +293,19 @@ sdist_build = repository_rule(
         "version": attr.string(),
         "pre_build_patches": attr.label_list(default = []),
         "pre_build_patch_strip": attr.int(default = 0),
+        "toolchains": attr.string_list(
+            default = ["@bazel_tools//tools/cpp:current_cc_toolchain"],
+            doc = "Toolchain labels forwarded to the generated pep517_native_whl(...) call. Emitted verbatim — set to [] to strip the default CC toolchain. Set via `uv.override_package(toolchains = [...])`.",
+        ),
+        "env": attr.string_dict(
```
dzbarsky: ditto as above, right? These are incomplete and we can probably define them in just one spot?

uv/private/extension/defs.bzl
dzbarsky: I guess if you really want to, you can add an explicit `skip_default_cc_toolchain` attr? but i really doubt that's a knob you want people using

uv/private/extension/defs.bzl
jvolkman: It's been years since I worked on this stuff, but I found that just setting these environment variables wasn't sufficient in `rules_pycross`. I had to take control over `sysconfig`. E.g., https://github.com/jvolkman/rules_pycross/blob/main/pycross/private/tools/wheel_builder.py#L595-L600

jbedard: Thanks for the info @jvolkman

I think this `toolchains`+`env` API here (which is trying to align with rules_rs) would not block additional APIs for `sysconfig` in the future though? IIUC?

jbedard: @dzbarsky or @jvolkman any concerns with how this is wired up now?

xangcastle (approved):

PR #1007, opened 2026-05-12
feat: suppress repository-rule spam behind RULES_PY_UV_VERBOSE flag

xangcastle:
## Summary
Eliminates unconditional `print()` spam from `sdist_build`, `git_archive`, and the uv module extension by gating diagnostic output behind the `RULES_PY_UV_VERBOSE` environment variable. Also consolidates the `override_package` noise into a single summary line.

## Motivation
Every `bazel build` was emitting multiple warnings and debug prints from repository rules:
- sdist configure-tool failures
- missing archive-path resolutions  
- native-source detections
- pure-Python fallbacks
- per-override `print()` from the module extension
- git-archive command dumps

These messages are useful for debugging but create noise during normal builds. Users should be able to toggle them on without editing `MODULE.bazel`.

### `uv/private/extension/defs.bzl`
- Removed the `_extension_tag` with a `verbose` attribute (too intrusive for quick debug).
- Replaced per-override `print()` with an accumulated counter; now prints a single line:
  `uv: N package(s) overridden` only when overrides exist.

### `uv/private/sdist_build/repository.bzl`
- Removed `verbose` attribute from the repository rule.
- Gated the 4 diagnostic prints behind `rctx.os.environ.get("RULES_PY_UV_VERBOSE", "")`.

### `uv/private/git_archive/repository.bzl`
- Removed `verbose` attribute from the repository rule.
- Gated git command/stdout/stderr dumps behind the same env var.

### `docs/uv.md`
- Added **Troubleshooting → Verbose logging for uv repository rules** section documenting the flag and the diagnostics it enables.

### `.bazelrc`
- Added a commented-out example:

### DIFF
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -144,6 +144,7 @@ def _parse_projects(module_ctx, hub_specs):
 
     install_cfgs = {}
     install_table = {}
+    override_count = 0
 
     # FIXME: Collect build deps files/annotations
 
@@ -243,7 +244,7 @@ def _parse_projects(module_ctx, hub_specs):
 
                 k = (project_id, normalize_name(override.name), v, "__base__")
                 if has_target:
-                    print("Overriding {}@{} in {} with {}".format(override.name, v, project_name, override.target))
+                    override_count += 1
                     install_table[k] = str(override.target)
 
             # Lazily evaluated cache
@@ -460,6 +461,9 @@ def _parse_projects(module_ctx, hub_specs):
             for package, cfgs in version_activations.items():
                 for cfg in cfgs.keys():
                     hub_cfg.packages.setdefault(package, {})[cfg] = "@{}//:{}".format(project_id, package)
+
+    if override_count:
+        print("uv: {} package(s) overridden".format(override_count))
 
     return struct(
         project_cfgs = project_cfgs,
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -144,7 +144,6 @@ def _parse_projects(module_ctx, hub_specs):
 
     install_cfgs = {}
     install_table = {}
-    override_count = 0
 
     # FIXME: Collect build deps files/annotations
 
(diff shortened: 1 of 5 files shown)


uv/private/extension/defs.bzl
```diff
@@ -243,7 +244,7 @@ def _parse_projects(module_ctx, hub_specs):
 
                 k = (project_id, normalize_name(override.name), v, "__base__")
                 if has_target:
-                    print("Overriding {}@{} in {} with {}".format(override.name, v, project_name, override.target))
```
jbedard, line 246: We don't want this detailed message when the verbose flag is on?

uv/private/git_archive/repository.bzl
```diff
@@ -57,16 +57,18 @@ filegroup(
...
     status = repository_ctx.execute(
         cmd,
     )
 
-    print("Git exited {}".format(status.return_code))
-    print(status.stdout)
-    print(status.stderr)
+    if repository_ctx.os.environ.get("RULES_PY_UV_VERBOSE", ""):
+        print("Git exited {}".format(status.return_code))
+        print(status.stdout)
```
jbedard, line 76: I think the stdout+err should always be outputted if we are going to `fail()` below...

uv/private/git_archive/repository.bzl
```diff
@@ -57,16 +57,18 @@ filegroup(
         target_ref,
     ]
 
-    print(cmd)
+    if repository_ctx.os.environ.get("RULES_PY_UV_VERBOSE", ""):
```
jbedard: I think we need to use `repository_ctx.getenv()` instead of accessing `repository_ctx.os.environ` in order for the vars to be tracked and the rule to be reproducible, read more at https://bazel.build/external/repo#when_is_the_implementation_function_executed

uv/private/extension/defs.bzl
xangcastle: Updated @jbedard

uv/private/git_archive/repository.bzl
```diff
@@ -57,18 +57,20 @@ filegroup(
...
     # Execute the archive command
     status = repository_ctx.execute(
         cmd,
     )
 
-    print("Git exited {}".format(status.return_code))
-    print(status.stdout)
-    print(status.stderr)
+    if repository_ctx.getenv("RULES_PY_UV_VERBOSE", ""):
+        print("Git exited {}".format(status.return_code))
```
jbedard: You don't want to still spam stdout/err in this case?

xangcastle: Yes, I am doing that a few lines below @jbedard

jbedard: Only if `return_code != 0`, where previously it spammed 100% of the time

jbedard (approved):

PR #1009, opened 2026-05-13
fix(uv): pin sdist build tools to configured Python

jbedard:

Add:
```
uv.project(python_version)
```
to declare an explicit python version used by a project

### Test plan

- New test cases added


e2e/cases/uv-sdist-python-version-1004/setup.MODULE.bazel
```diff
@@ -0,0 +1,24 @@
...
+uv.declare_hub(hub_name = "pypi_uv_sdist_python_version_1004")
+uv.project(
+    default_build_dependencies = [
+        "build",
+        "setuptools",
+    ],
+    hub_name = "pypi_uv_sdist_python_version_1004",
+    lock = "//cases/uv-sdist-python-version-1004:uv.lock",
+    pyproject = "//cases/uv-sdist-python-version-1004:pyproject.toml",
+    python_version = "3.10",
```
jbedard, line 22: Is this really the greatest API? I feel like this is odd...

@dzbarsky any idea what the real use case was here? Should this at least be moved to the `declare_hub` so it is a hub that has a specific version and then `py_binary|test(python_version)` must align with it? Or is it ok if they don't align?

dzbarsky: did I write this? Sorry I don't recall this patch

jbedard: Yes you did. I believe for cases where the default python version is not usable by a pip package, so it can be overridden here for pip package setup/compilation?

The odd part is when you then must specify that python version _again_ in `py_test|binary` when using the pip package at execution time. That API just seems off or as if it should be unnecessary...

dzbarsky: Its possible it was a workaround for issues with previous rules_python toolchain. Might also have been a fix for the config sprawl eating up disk. You can try asking for us to drop this patch and see what breaks?

jbedard: Seems like we can just drop it for now 🤷

jbedard:
Confirmed this seems no longer necessary

PR #1010, opened 2026-05-13
fix(py): reset dep_group on data edges

jbedard:
Runfiles `data` deps are packaged artifacts, not part of the parent's resolved venv — they should not inherit the parent binary's `dep_group`. Reset it on `data` edges; new tests verify probes in runfiles see the default lock venv.

### Test plan

- New test cases added

### DIFF
--- a/e2e/cases/multi-project-hub/BUILD.bazel
+++ b/e2e/cases/multi-project-hub/BUILD.bazel
@@ -2,7 +2,8 @@
 # e2e snapshots. Three sub-projects (alpha, beta, gamma) all depend on
 # cowsay, exercising the N>1 select-arm and dep_group config_setting
 # emission that the single-project @pypi_single snapshot cannot cover.
-load("@aspect_rules_py//py:defs.bzl", "py_test")
+load("@aspect_rules_py//py:defs.bzl", "py_binary", "py_library", "py_test", "py_venv")
+load("@aspect_rules_py//py/private/py_venv:py_venv_exec.bzl", "py_venv_exec_test")
 
 # Alpha defines `[dependency-groups]` (`dev` + hyphenated `unit-tests`)
 # instead of an implicit project-name group, so the snapshot covers both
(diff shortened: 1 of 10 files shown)


py/private/py_library.bzl
```diff
@@ -187,8 +188,11 @@ _attrs = dict({
         The transitive closure of the `data` dependencies will be available in the `.runfiles`
         folder for this binary/test. The program may optionally use the Runfiles lookup library to
         locate the data files, see https://pypi.org/project/bazel-runfiles/.
+        Data deps are packaged runfiles artifacts and should not inherit the parent's
```
jbedard: @dzbarsky I'm curious about this one as well. What's the point of the hub/project API if the `dep_group` (renamed from `venv` in 2.x) gets reset every time it crosses a library?

I would think if a `py_library` includes a project configured for a certain `dep_group` it should use the `dep_group` configured in the hub/project API.

Maybe I'm misunderstanding something here?

py/private/transitions.bzl
```diff
@@ -57,3 +60,18 @@ python_transition = transition(
 
 # The old name, FIXME: refactor this out
 python_version_transition = python_transition
+
+# Data deps are packaged into runfiles and should not inherit custom py_binary/py_venv
+# dep_group settings. We reset back to the default to match the default .bazelrc settings.
+def _reset_dep_group_transition_impl(_settings, _attr):
```
dzbarsky: i don't think the comment matches what is actually happening? Our use case (and perhaps other common ones for monorepos) is to have a default closure in .bazelrc, and individual binaries can opt into custom ones. So you do actually want to reset back to the default rather than unsetting the flag.

jbedard: Doesn't that just mean binaries declare the `dep_group` they want to use (to override the bazelrc default which overrides the default default) and that gets transitioned throughout the graph for that binary? I'm not sure why we need a "reset" at the library level?

dzbarsky: the point is to collapse all configs of data down to a single one. I don't see how the code here accomplishes that unless I'm missing something?

dzbarsky: generally on an outgoing `data` edge you want to reset all ruleset attributes. Imagine your `data` is another binary or lib you build from source, like C/Rust/etc...

jbedard: > I don't see how the code here accomplishes that unless I'm missing something?

You/codex wrote it, I/claude rebased it 🤷 

> Imagine your data is another binary or lib you build from source, like C/Rust/etc...

Don't you want to just consume that library as it was build from source, instead of changing options on it and building it again?

dzbarsky: this patch is not quite what we use. We actually force-set it to our default virtualenv name, but that's just a hack. the ruleset needs a more real fix

and yes, my point is exactly that you want to only build the lib once, not multiple times... the whole point of the rest transition is to unify the configs so you don't end up with multiple copies of the data deps

jbedard: Ok, I think we agree with the "only build the lib once" part then. It's just a question of how to put that in a clear API?

No matter how many times an underlying package is used it must only be built once. Maybe a `--config` or `--@<hub>//dep_group=<name>` could change it, but under a single bazel invocation the target representing a package will only be built once?

uv/private/whl_install/rule.bzl
```diff
@@ -5,6 +5,22 @@ load("@rules_python//python:defs.bzl", "PyInfo")
 load("//py/private:providers.bzl", "PyWheelsInfo")
 load("//py/private/toolchain:types.bzl", "EXEC_TOOLS_TOOLCHAIN", "PY_TOOLCHAIN", "UNPACK_TOOLCHAIN")
 
+_DEP_GROUP_FLAG = "@aspect_rules_py//uv/private/constraints/dep_group:dep_group"
```
xangcastle, line 8: If DEP_GROUP_FLAG is scope="universal". Are there other universal flags that could still cause duplicate whl_install actions under different configurations (e.g., platform variants, Python version transitions)?

e2e/cases/multi-project-hub/BUILD.bazel
```diff
@@ -50,3 +51,108 @@ py_test(
         "@pypi//cowsay",
     ],
 )
+
+# Within a single uv project, a hub package should be installed once
+# regardless of how many distinct dep_group consumers reach it. Each
+# `record_path_*` binary prints `Path(cowsay.__file__).resolve()` and
+# `:hub_dep_single_compile_test` asserts the recorded paths match.
+[
+    py_binary(
```
xangcastle, line 60: The record_path_* binaries use dep_group directly on py_binary. Is dep_group a first-class attribute of py_binary, or is it sugar that routes through an auto-generated py_venv sibling (like py_test does)?

e2e/cases/multi-project-hub/record_path.py
```diff
@@ -0,0 +1,9 @@
+"""Print the resolved on-disk location of `cowsay.__file__` for the current
+build configuration. Used by `:hub_dep_single_compile_test` to compare what
+multiple `dep_group` consumers see for the same hub package."""
+
+from pathlib import Path
+
```
xangcastle, line 6: If anything else writes to stdout (e.g., a warning from cowsay or an import side-effect), the path parsing in compare_paths.py will break silently

jbedard:
Need better solution from @xangcastle

PR #960, opened 2026-05-01
feat: optimized py_image_layer

thesayyn:
Rewrite py_image_layer with analysis-time grouped OCI layers and
globally-shared pip tars via two propagated aspects:

- _layer_aspect propagates through deps/data/actual; pip per-package
  tars are declared at the pip target's namespace so they action-share
  across every rule using that package. First-party PyInfo targets
  matched by layer_tier.groups are captured for per-group tarring at
  the binary.
- _merge_aspect runs only on the binary, reads pip_packages closure,
  buckets install_dirs by merge_group, and emits one bsdtar action per
  group from raw install_dirs (single pass, no intermediates).

A new layer_tier rule carries grouping + compression as LayerTierInfo;
aspects read it via a //py:layer_tier label_flag so users can switch
tiers globally with --//py:layer_tier=//path:custom_tier or pin one
per rule via the layer_tier attr. A Python validation action surfaces
layer_tier(groups=...) suggestions when the squashed pip layer or
individual packages exceed thresholds, or the OCI 127-layer limit is
breached.

Sharing model:
- Solo whole-group + subpath-split pip tars: action-shared across
  every rule using that package (declared at the pip target's
  namespace).
- Multi-member merged tars: per-binary action, deterministic content
  (canonical mtree, fixed bsdtar options) so remote CAS / OCI registry
  dedupes by digest.
- First-party grouped tars: per-binary, one tar per group, collected
  from matched py_library targets in the binary's dep closure.
- Ungrouped pip packages: squashed by the rule into one per-rule tar.
- Optional dedicated interpreter layer when layer_tier.interpreter_group
  is set; the tar is declared at the py toolchain target's namespace
  so it action-shares across every py_image_layer using that toolchain
  config.

### Changes are visible to end-users: yes

- Searched for relevant documentation and updated as needed: yes
- Breaking change (forces users to change their own code or config): yes
- Suggested release notes appear below: yes

New public api of `py_image_layer` is quite different than the previous one. 

### Test plan

- Covered by existing test cases
- New test cases added

PR #1015, opened 2026-05-14
test: put oci image snapshots into /snapshots/ dir

jbedard:
### Test plan


e2e/cases/oci/py_image_layer/BUILD.bazel
```diff
@@ -12,6 +12,13 @@ platform(
         # rules_rs Rust toolchains for linux-gnu targets require gnu.2.28.
         "@llvm//constraints/libc:gnu.2.28",
     ],
+    # aspect_rules_py's interpreter toolchains are gated on the
```
xangcastle, line 15: Is it intentional that platform_libc=glibc is declared as a flags entry on the platform() rather than being derived from the existing @llvm//constraints/libc:gnu.2.28 constraint? 

It feels slightly redundant to declare both the constraint and the flag. If they ever diverge (e.g., a platform with musl constraint but glibc flag), which one wins?

e2e/cases/oci/py_image_layer/asserts.bzl
```diff
@@ -24,15 +23,12 @@ for f in $(SRCS); do
 done > $@
 """.format(actual),
         toolchains = ["@bsd_tar_toolchains//:resolved_toolchain"],
-        # HACK: avoid running tests that depend on this output on Bazel 9.
-        # For some reason, the listing is different than on Bazel 8, in a way that's hard to scrub.
-        target_compatible_with = ["@platforms//:incompatible"] if bazel_features.rules.merkle_cache_v2 else [],
     )
 
     write_source_file(
```
xangcastle, line 28: With the pre-commit hooks that auto-update these listings now removed (.pre-commit-config.yaml), how are developers expected to regenerate these snapshots when the layer contents change? 

write_source_file supports bazel run for manual updates, but without an automated check it's easy for the snapshots to drift. Is there a CI check that fails when the snapshots are stale, or should a note be added to the module's README/CONTRIBUTING docs?

jbedard: `write_source_file` also has tests to ensure the file is in sync

e2e/cases/oci/py_image_layer/BUILD.bazel
jbedard: I honestly don't understand this. Without this the oci images produced on mac fallback to using files out of `rules_python` instead of `aspect_rules_py`? I could try revert and simplifying it...

jbedard: Claudes reasoning, which explains why without this when running on mac I see `rules_python`:

```
Why the explicit =glibc pin matters here:
  - The two linux platforms (x86_64_linux, aarch64_linux) declare cpu+os+libc constraints but, without the pin, platform_libc keeps the host-derived default.
  - Cross-build from a darwin host → linux target: default = libsystem. Neither glibc nor musl toolchain's target_settings match, so aspect_rules_py's interpreters are filtered out
  and Bazel falls back to rules_python's auto-registered interpreter (rules_python++python+python_3_11_x86_64-unknown-linux-gnu). 
  - Cross-build from a linux glibc host: default = glibc, so things work "by accident".
  - That's exactly why CI (linux) and local (darwin) produced different runfiles paths in the snapshot before the pin.
```

jbedard: Looks like the same was already done in https://github.com/aspect-build/rules_py/pull/960/changes#diff-74b666ac756de00709b83d55fcd4de01ba88f539e694f9bfd92cd39cbe05212a ...?

xangcastle (approved):

PR #1022, opened 2026-05-16
fix(sdist): prefer setup.py when pyproject metadata is incomplete

jbedard:
### Test plan

- New test cases added


jbedard:
replaced with https://github.com/aspect-build/rules_py/pull/1030

PR #1030, opened 2026-05-18
fix(sdist): prefer setup.py when pyproject metadata is incomplete

jbedard:
### Test plan

- New test cases added


uv/private/pep517_whl/build_helper.py
```diff
@@ -8,12 +8,165 @@
...
+# to `execvp` by basename via PATH.
+_DEBUG_FLAG = "-fdebug-default-version=4"
+_COMPILER_WRAPPER = """#!/usr/bin/env python3
+import os
+import sys
+
+filtered_args = [arg for arg in sys.argv[1:] if arg != "{debug_flag}"]
+compiler_path = "{compiler_path}"
+if os.path.exists(compiler_path):
+    os.execv(compiler_path, ["{name}"] + filtered_args)
```
jbedard: An additional `os.exec`?

jbedard: and shouldn't we `return` if it was successful and avoid hitting the `os.execvp` below?

xangcastle (approved):

PR #1035, opened 2026-05-20
perf: remove <wheel>/<i> intermediate symlinks

jbedard:
---

### Test plan

- New test cases added


xangcastle:
This PR is now superseded by #1038, which solved the same cache-bust problem (unstable _wheels/<i> indices) by replacing numeric indices with stable 8-hex hashes (_wheels/<hash>/) instead of removing the aliases entirely.
Removing the two-hop symlink aliases breaks py_image_layer and py_venv_image_layer.

I'd recommend closing it. Happy to reopen if I'm missing something.

PR #1036, opened 2026-05-20
refactor: move VIRTUAL_ENV declaration into bazel RunEnvironmentInfo

jbedard:
### Test plan

### DIFF
--- a/py/private/py_venv/py_venv.bzl
+++ b/py/private/py_venv/py_venv.bzl
@@ -123,6 +123,9 @@ def _py_venv_rule_impl(ctx):
         is_executable = True,
     )
 
+    if "VIRTUAL_ENV" in ctx.attr.env:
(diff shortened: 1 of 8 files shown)


py/private/py_venv/py_venv.bzl
```diff
@@ -123,6 +123,9 @@ def _py_venv_rule_impl(ctx):
         is_executable = True,
     )
 
+    if "VIRTUAL_ENV" in ctx.attr.env:
```
jbedard, line 126: The `.sh` always overwrote this before, just like now `passed_env["VIRTUAL_ENV"]` always overwrites it.

So this `fail()` isn't directly related to the refactor but I think is worth doing to avoid confusion?

xangcastle (approved):

PR #1044, opened 2026-05-22
refactor: replace dependency on shell awk+sort with gawk bcr module

jbedard:
### Test plan


py/private/py_image_layer.bzl
```diff
@@ -710,33 +720,53 @@ _platform_cfg = transition(
...
     mtree_args.add_all(files_depset, map_each = map_each, expand_directories = False, allow_closure = True)
 
     awk_script = ctx.file._awk_script
-    ctx.actions.run_shell(
-        inputs = depset(direct = [awk_script], transitive = [files_depset, bsdtar_files.files]),
+    awk = ctx.executable._awk
+    sorted_mtree = ctx.actions.declare_file(tar_out.basename + ".mtree", sibling = tar_out)
+
+    gawk_args = ctx.actions.args()
+    gawk_args.add("-v", "outfile=" + sorted_mtree.path)
```
dzbarsky: don't use `.path`

dzbarsky (commented):
I know Alex liked the idea of using awk but honestly even with "hermetic gawk" i've seen issues where your awk scripts somehow weren't compatible. Unless any of you really enjoy awk I personally would just rewrite this as a simple Go/rust/zig/etc tool...

Last time I wrote a Go-based mtree rewriter it was like 20 lines.

MODULE.bazel
```diff
@@ -103,15 +104,15 @@ use_repo(tel, "aspect_tools_telemetry_report")
 # Dev deps are the remainder of the file
 
 # LLVM configuration
-bazel_dep(name = "llvm", version = "0.7.1", dev_dependency = IS_RELEASE)
+bazel_dep(name = "llvm", version = "0.8.3", dev_dependency = IS_RELEASE)
 
 # Patch http_bsdtar_archive to tolerate missing `--include` patterns.
```
dzbarsky: You can remove this patch and the OSX override yall added

jbedard:
Note the goal of this PR is to not break on older linux distributions, like what rules_py currently runs on for BCR (debian2002?). I just want to make it (more) hermetic for now, we can debate replacing it with something else later.

py/private/py_image_layer.bzl
jbedard: Sorry that's me being stupid and trusting the robots, I should have caught that before requesting review 😅

xangcastle (approved):

PR #1045, opened 2026-05-22
perf: migrate bash launcher to hermetic-build

jbedard:

Use https://github.com/hermeticbuild/hermetic-launcher to launch the python toolchain instead of depending on a shell wrapper.

### Test plan

- Manual testing


py/private/py_venv/_virtualenv.py
```diff
@@ -3,6 +3,12 @@
 import os
 import sys
 
+# Make wheel-declared console scripts reachable via `subprocess.run("name", ...)`.
+_venv_bin = os.path.join(sys.prefix, "bin")
+if _venv_bin not in os.environ.get("PATH", "").split(os.pathsep):
+    os.environ["PATH"] = _venv_bin + os.pathsep + os.environ.get("PATH", "")
```
jbedard, line 11: Does anyone know of a better way of doing this? Should it be opt-in?

py/private/py_venv/types.bzl
```diff
@@ -1,5 +1,12 @@
 """quasi-public types."""
 
+PyExecutableInfo = provider(
```
jbedard: This replaces the `__PEX_PY_BINARY_ENTRYPOINT__` we were inserting into the shell script.

I don't know the history behind `py_pex_binary` or if this is the ideal solution. @thesayyn?

py/private/py_venv/types.bzl
```diff
@@ -1,5 +1,12 @@
...
+    doc = "Provider emitted by `py_venv_exec` carrying launcher-level metadata that downstream consumers can't infer from the binary itself.",
+    fields = {
+        "entrypoint_rlocation": "str — runfiles path of the `main` Python file the launcher exec's. Read by `py_pex_binary` to set the pex entrypoint without scraping the launcher.",
```
thesayyn: should be a File instead of a str.  downstream consumers can decide what they want to do with it.

jbedard: See fixup: https://github.com/aspect-build/rules_py/pull/1045/changes/b5ecd19cfaf3b7945fba945ba8486b2c51d3ac62

py/private/py_venv/types.bzl
thesayyn: this should not be needed now that we have https://github.com/aspect-build/rules_py/commit/b5ecd19cfaf3b7945fba945ba8486b2c51d3ac62

jbedard: Because we can just use `DefaultInfo.executable` or something like that?

thesayyn (approved):

uv/private/pep517_whl/build_helper.py
```diff
@@ -46,6 +46,13 @@
 sysroot = {sysroot!r}
 if sysroot and "-isysroot" not in filtered_args:
     filtered_args = ["-isysroot", sysroot] + filtered_args
+# Force `lld` as the linker so clang doesn't search for a system `ld`
+# (which isn't on PATH inside the action sandbox — only the venv bin
+# is). The hermetic LLVM toolchain ships `ld.lld` alongside `clang`,
+# so this is a no-op argv on compile-only calls and resolves the
+# linker hermetically on link calls.
+if not any(a == "-fuse-ld" or a.startswith("-fuse-ld=") for a in filtered_args):
```
jbedard: After feedback from @dzbarsky I want to understand this more. Both why switching from a bash launcher to hermetic launcher required this and why only on bazel 8.

jbedard: Using `common --incompatible_strict_action_env` (defaults in bazel 9, off in bazel 8) fixes the e2e tests in bazel 8. I would prefer that over injecting flags like this

py/private/py_venv/_virtualenv.py
jbedard: I've found no better way of doing this. Alternatives would be:
* wrapper script, exactly what this PR is dropping
* `hermetic_launcher` supporting env vars (or appending to them specifically) which seems like a bad idea

PR #1049, opened 2026-05-22
test: declare module test uv extension as isolated

jbedard:
### Test plan


jbedard:
Seems to be hidden behind a bazel flag for now so we can't depend on this

PR #1050, opened 2026-05-22
test: bazel_env integration with the current Python toolchain

xangcastle:
Verifies that bazel_env can consume @python_interpreters//:current_py_toolchain and correctly generates the expected tool launchers and toolchain symlinks.


xangcastle:
Hi @ctcjab I would like to see your comments on this one

ctcjab (approved):
Thanks for adding this @xangcastle — nice to have an e2e test proving the `bazel_env` integration works end-to-end with `current_py_toolchain`.

One suggestion: the test currently checks that the launcher/symlink files *exist*, but doesn't actually invoke the interpreter through them. Adding something like `"${python3_launcher}" -c 'import sys; print(sys.version)'` would catch cases where the symlink points somewhere broken. Not a blocker though — LGTM as-is.

jbedard (approved):

PR #1058, opened 2026-05-29
fix: point pyvenv.cfg home= directly at PBS bin/ to fix sys.base_prefix on Python 3.11/3.12

### LINKED ISSUE #1048: [Bug]: v2.0.0-alpha.1 breaks sys.base_prefix for Python 3.11/3.12 with native extensions
### What happened?

After upgrading from `aspect_rules_py` 1.11.5 to 2.0.0-alpha.1, `py_test` targets that load native C++ extensions (built via `nanobind_extension`) crash on Python 3.11 and 3.12 with `ModuleNotFoundError: No module named 'encodings'`. Python 3.13 is unaffected.

The error output shows `sys.base_prefix = '/install'` and `stdlib dir = '/install/lib/python3.11'` — the interpreter cannot resolve its standard library through the new Starlark-assembled venv's symlink chain.

```
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = '.../test_native_module_test.py311.runfiles/_main/.../._test_native_module_test.py311.venv/bin/python'
  isolated = 1
  sys._base_executable = '.../rules_python++python+python_3_11_x86_64-unknown-linux-gnu/bin/python3.11'
  sys.base_prefix = '/install'
  sys.base_exec_prefix = '/install'
  sys.prefix = '/install'
  sys.exec_prefix = '/install'
  sys.path = [
    '/install/lib/python311.zip',
    '/install/lib/python3.11',
    '/install/lib/python3.11/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'
```

This works correctly on 1.11.5 for all three Python versions.

### Version

- **Host OS:** RHEL 9 / x86_64 (Linux 5.14.0)
- **Bazel:** 9.1.0
- **aspect_rules_py:** 2.0.0-alpha.1 (regression vs 1.11.5)
- **rules_python:** 2.0.1 (providing hermetic Python toolchains for 3.11, 3.12, 3.13)
- **Native extension tooling:** nanobind_bazel 2.12.0 + hermetic_cc_toolchain 4.1.0

### How to reproduce

1. Register multiple Python toolchains via `rules_python` (3.11, 3.12, 3.13):
```starlark
python.toolchain(python_version = "3.11")
python.toolchain(python_version = "3.12")
python.toolchain(is_default = True, python_version = "3.13")
```

2. Create a `nanobind_extension` (or any native C++ Python extension) and a `py_test` that imports it, configured for multi-version testing:
```starlark
ctc_py_test(  # wrapper around py_test
    name = "test_native_module_test",
    srcs = ["test_native_module_test.py"],
    data = [":example_ext.so"],
    python_versions = ["3.11", "3.12", "3.13"],
)
```

3. Run the test for Python 3.11 or 3.12:
```shell
bazel test //:test_native_module_test.py311
# FAILS: No module named 'encodings'

bazel test //:test_native_module_test.py313
# PASSES
```

### Any other information?

This appears related to the v2 migration from the Rust-based venv launcher to the Starlark-assembled venv with two-hop symlinks. The `bin/python` relative symlink change (aimed at fixing `sys.base_prefix` in containers) seems to break prefix resolution for the older `rules_python`-provided Python 3.11 and 3.12 interpreters, while 3.13 (possibly due to changes in CPython's path discovery logic) resolves correctly.

Possibly related to #703, though that issue was specifically about the `-I` flag. This regression occurs during normal `bazel test` execution without any special flags.

xangcastle:
Fix #1048

## Problem: [#1048 ](https://github.com/aspect-build/rules_py/issues/1048)

  Python 3.11/3.12 reimplemented prefix discovery as pure Python (`Modules/getpath.py`).
  Its `resolvedpath()` correctly handles one-level local symlinks but has a bug with
  multi-hop relative symlinks that traverse `..` components across repository boundaries —
  exactly the shape that `venv/bin/python` produces:

  venv/bin/python → ../../../../interpreter_repo/bin/python3.11

  When resolution fails, Python falls back to the PBS compile-time prefix `/install`,
  which does not exist at runtime:

  ModuleNotFoundError: No module named 'encodings'

  Python 3.9/3.10 used C code (correct). Python 3.13 fixed the bug. Only 3.11 and 3.12
  are affected.

  ## Fix

  Change `pyvenv.cfg`'s `home =` key for hermetic (runfiles-based) interpreters from
  `./bin/` to a relative path that points **directly** at the PBS `bin/` directory:

  home = ../../../../aspect_rules_py++python_interpreters+.../bin

  This means `getpath.py` only needs to resolve `bin/python → python3.11` — a single
  local hop within the PBS package — which works correctly in all Python versions.

  System (non-hermetic) interpreters keep `home = ./bin` unchanged.

  ## Changes

  - `py/private/py_venv/venv.bzl` — compute `pyvenv_home` from the interpreter's
    rlocation path for hermetic toolchains; fall back to `./bin` for system interpreters
  - `e2e/tools/verify_venv/verify_venv.py` — add `verify_base_prefix()` asserting
    `sys.base_prefix != "/install"` and that the stdlib dir exists; call it from
    `verify_all()`
  - `e2e/cases/pbs-symlink-prefix-1048/` — regression tests for Python 3.11 and 3.12
  - OCI layer snapshots updated to reflect the new `pyvenv.cfg` and `verify_venv.py` sizes

### DIFF
--- a/e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
+++ b/e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
@@ -1,0 +1,48 @@
+"""Regression test for issue #1048: PBS symlink chain breaks sys.base_prefix.
+
+Python 3.11/3.12 reimplemented prefix discovery in pure Python (getpath.py).
+Its resolvedpath() has a bug with multi-hop relative symlinks that traverse
+.. components across repo boundaries — the exact shape of the venv's
+bin/python → ../../../../interpreter_repo/bin/python3.11 chain.
+
+When resolution fails, Python falls back to the compiled-in /install prefix
+(absent at runtime), causing:
+    ModuleNotFoundError: No module named 'encodings'
+
+The fix: pyvenv.cfg's home= key now points directly to the PBS bin/ directory,
+so Python only resolves the local python → python3.11 symlink (one hop).
+"""
+
+import os
+import sys
+
+from verify_venv import verify_all, verify_base_prefix, verify_in_venv
+
+
+def test_base_prefix_not_install():
+    assert sys.base_prefix != "/install", (
+        f"sys.base_prefix is '/install' (the PBS compile-time prefix). "
+        f"Python {sys.version_info.major}.{sys.version_info.minor} failed to "
+        f"resolve the pyvenv.cfg home= symlink chain."
+    )
+
+
+def test_base_prefix_has_stdlib():
+    stdlib = os.path.join(
+        sys.base_prefix,
+        "lib",
+        f"python{sys.version_info.major}.{sys.version_info.minor}",
+    )
+    assert os.path.isdir(stdlib), (
+        f"stdlib missing: {stdlib!r} (sys.base_prefix={sys.base_prefix!r})"
+    )
+
+
+if __name__ == "__main__":
+    verify_all()
+    test_base_prefix_not_install()
+    test_base_prefix_has_stdlib()
+    print(
+        f"OK: sys.base_prefix={sys.base_prefix!r}, "
+        f"Python {sys.version_info.major}.{sys.version_info.minor}"
+    )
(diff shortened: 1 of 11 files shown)


xangcastle:
@ctcjab please take a look

e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
```diff
@@ -0,0 +1,48 @@
+"""Regression test for issue #1048: PBS symlink chain breaks sys.base_prefix.
```
jbedard, line 1: Is this test (and associated BUILD) simple enough to just put into `py/tests` as a unit test?

py/tests/pbs-symlink-prefix-1048/verify_venv.py
```diff
@@ -0,0 +1,171 @@
+"""Venv / interpreter layout assertions for e2e tests.
```
jbedard: This file was duplicated? Lets not duplicate anything, if this is required then keep the files in `e2e`...

e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
jbedard: nm, the test depends on other things in e2e

jbedard (approved):

PR #1065, opened 2026-06-01
chore: bumping bazel_lib

xangcastle:
bumping bazel_lib
---

### Test plan


jbedard:
What for? These should be minimum versions and we shouldn't bump them just to keep up to date

PR #1077, opened 2026-06-04
fix: normalize default dependency group name to underscores to prevent configuration mismatches

xangcastle:
---

### Test plan

- New test cases added


e2e/cases/rdflib-linux/__test__.py
```diff
@@ -0,0 +1,14 @@
+"""Regression: rdflib must be importable on Linux.
```
jbedard, line 1: What makes this package unique that caused the bug? Outline it here...

e2e/cases/rdflib-linux/check_dep_group_names.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# Regression: dep_group config_setting names must be normalized (hyphens → underscores).
```
jbedard, line 2: Is the `.sh` really necessary? The `rdflib_linux` string is already in the BUILD and I think a comment at the top of the BUILD explaining why `rdliflib_linux` instead of `rdflib-linux` is enough and we can avoid the shell script?

xangcastle: rdflib-linux is not enough bc instead of fail is skipped

PR #1078, opened 2026-06-04
fix: sdist build pipeline, pep508 bracket tokenizer, whl visibility, and pytest coverage

xangcastle:
Applies and hardens a set of bug fixes reported in v1.11.5.

  ### Issues fixed
  - `build_helper.py`: descend into `subdirectory` for monorepo archives (`source.subdirectory` in `uv.lock`)
  - `build_helper.py`: fall back to `setuptools.build_meta` when the declared backend (e.g. `hatchling`, `flit_core`) is not importable in the sandbox
  - `build_helper.py`: generate `setup.cfg` with `[options.package_data] * = *` for setuptools < 61 so non-Python data files are included in the wheel
  - `build_helper.py`: set `SOURCE_DATE_EPOCH=315532800` to prevent `wheel < 0.38` from failing on Bazel's epoch-0 file timestamps
  - `build_helper.py`: add `--skip-dependency-check` to `python -m build`
  - `build_helper.py`: resolve `$CC`/`$CXX` to absolute paths before `cwd` changes into the extracted source tree
  - `sdist_build/repository.bzl`: add `subdirectory` attribute; skip build deps already provided via `default_build_dependencies` to prevent label collisions
  - `extension/defs.bzl`: add `setuptools` and `wheel` to `default_build_dependencies`; guard their lockfile resolution so projects that don't lock build tools are not broken

  **PEP-508 marker evaluation (issue 4 from second report)**
  - `pep508_evaluate.bzl`: fix tokenizer losing a variable token when a bracket character immediately follows without whitespace (e.g. `'linux' in sys_platform)` would crash with `key ")" not found`)
  - `markers/defs.bzl`: evaluate the marker once per candidate extra and OR the results, fixing false negatives for multi-extra conflict-group lockfiles

  **Dependency version resolution**
  - `extension/projectfile.bzl`: use `setdefault` instead of direct assignment when building `group_preferences` so the first (lockfile-derived) resolution wins when a package appears multiple times across resolved specs

  **whl visibility (issue 4 from first report)**
  - `whl_install/repository.bzl`: change `whl` select-chain target from `//visibility:private` to `//visibility:public` so spoke alias repos (`depctx_aliases`) can reference it

  **pytest runner**
  - `pytest.py.tmpl`: disable `anyio` plugin by default to prevent it from hijacking the event loop in non-async test suites
  - `pytest.py.tmpl`: check `coverage.Coverage.current()` before creating a new instance, fixing double-initialisation when multiple entry points are invoked (e.g. ROS launch tests)
  - `pytest.py.tmpl`: write coverage data to `$COVERAGE_DIR` when `--experimental_split_coverage_postprocessing` is active

  **Rust tooling**
  - `pth.rs` / `venv.rs`: skip `__pycache__` directories to prevent spurious collision errors from bytecode caches that embed source paths

  ### Tests added

  | File | Coverage |
  |------|----------|
  | `uv/private/markers/pep508_evaluate_test.bzl` | 11 new assertions for bracket-immediately-after-variable tokenizer fix |
  | `uv/private/extension/test_projectfile.bzl` | `collect_activated_extras_setdefault_preserves_first_resolution_test` |
  | `uv/private/pep517_whl/build_helper_test.py` | 15 new Python unit tests: backend fallback, `setup.cfg` generation, `[options.package_data]`, src-layout, `[project.scripts]` → `console_scripts`, CC/CXX absolute path resolution |

---

### Test plan
- New test cases added

### DIFF
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -360,6 +360,7 @@ def _parse_projects(module_ctx, hub_specs):
                         lock_build_deps = [
                             it[0]
                             for req in project.default_build_dependencies
+                            if normalize_name(req.split(";")[0].split("[")[0].split(">")[0].split("<")[0].split("=")[0].split("!")[0].split("~")[0].split(" ")[0].strip()) in package_versions
                             for it in extract_requirement_marker_pairs(project.lock, project_id, req, default_versions, package_versions)
                         ]
 
@@ -391,6 +392,7 @@ def _parse_projects(module_ctx, hub_specs):
                         pre_build_patch_strip = pre_build_patch_strip,
                         available_deps = project_available_deps,
                         configure_command = project.unstable_configure_command,
+                        subdirectory = package.get("source", {}).get("subdirectory", ""),
                     )
 
                     has_sbuild = True
@@ -574,6 +576,8 @@ def _uv_impl(module_ctx):
         if sbuild_cfg.pre_build_patches:
             sbuild_kwargs["pre_build_patches"] = sbuild_cfg.pre_build_patches
             sbuild_kwargs["pre_build_patch_strip"] = sbuild_cfg.pre_build_patch_strip
+        if sbuild_cfg.subdirectory:
+            sbuild_kwargs["subdirectory"] = sbuild_cfg.subdirectory
         sdist_build(**sbuild_kwargs)
 
     for install_id, install_cfg in cfg.install_cfgs.items():
@@ -628,6 +632,8 @@ _project_tag = tag_class(
             mandatory = False,
             default = [
                 "build",
+                "setuptools",
+                "wheel",
             ],
         ),
         "unstable_configure_command": attr.string_list(
--- a/uv/private/pep517_whl/build_helper.py
+++ b/uv/private/pep517_whl/build_helper.py
@@ -20,6 +20,7 @@ PARSER.add_argument("outdir")
 PARSER.add_argument("--validate-anyarch", action="store_true")
 PARSER.add_argument("--patch-strip", type=int, default=0, help="Strip count for patch (-p)")
 PARSER.add_argument("--patch", action="append", default=[], dest="patches", help="Patch file to apply (repeatable)")
+PARSER.add_argument("--subdirectory", default="", help="Subdirectory within the archive containing pyproject.toml")
 opts, args = PARSER.parse_known_args()
 
 tmp_root = opts.outdir.lstrip("/") + ".tmp"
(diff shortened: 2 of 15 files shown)


uv/private/extension/defs.bzl
```diff
@@ -360,6 +360,7 @@ def _parse_projects(module_ctx, hub_specs):
                         lock_build_deps = [
                             it[0]
                             for req in project.default_build_dependencies
+                            if normalize_name(req.split(";")[0].split("[")[0].split(">")[0].split("<")[0].split("=")[0].split("!")[0].split("~")[0].split(" ")[0].strip()) in package_versions
```
jbedard, line 363: This needs an explanation lol, such as a method with unit tests lol

uv/private/pep517_whl/build_helper.py
```diff
@@ -41,15 +46,144 @@
...
+    name = None
+    version = None
+    build_backend = None
+    in_project = False
+    in_build_system = False
+    in_project_scripts = False
+    has_src_layout = path.isdir(path.join(src_dir, "src"))
+    scripts = {}
+
+    for line in lines:
```
jbedard, line 72: This parser scares me

uv/private/pep517_whl/build_helper.py
```diff
@@ -41,15 +46,144 @@
...
+                    cfg_content += "    {} = {}\n".format(script_name, script_val)
+            with open(setup_cfg_path, "w") as f:
+                f.write(cfg_content)
+
+_ensure_build_backend(t)
+
 # Get a path to the outdir which will be valid after we cd
 outdir = path.abspath(opts.outdir)
 
+# Resolve CC/CXX to absolute paths so they remain valid after cwd changes
```
jbedard, line 171: Why is this required now? Can this be an isolated fix?

uv/private/pep517_whl/build_helper.py
```diff
@@ -62,6 +196,7 @@
         "-m", "build",
         "--wheel",
         "--no-isolation",
+        "--skip-dependency-check",
```
jbedard, line 199: Why is this required now? Can this be an isolated fix?

uv/private/sdist_build/repository.bzl
```diff
@@ -258,6 +276,10 @@ sdist_build = repository_rule(
                   "two arguments. See //uv/private/sdist_configure:defs.bzl.",
         ),
         "version": attr.string(),
+        "subdirectory": attr.string(
```
jbedard, line 279: Is this a new feature that we want in 2.x? Can this be done on its own?

PR #1079, opened 2026-06-04
feat: add project-qualified hub labels @hub//project/<stamp>:<pkg>

xangcastle:
Each hub now generates a `//project/<stamp>/` subdirectory with aliases for every package in that project.

  @hub//project/my_project:requests   # qualified
  @hub//requests                      # unqualified (unchanged)

Both forms resolve to the same target in single-project hubs. The qualified form becomes useful in multi-project hubs: when two projects provide the same package under the same dep_group name, the unqualified alias raises Bazel's "multiple keys match" error. The qualified alias scopes the `select()` to a single project, resolving unambiguously.

The `//project/<stamp>/BUILD.bazel` is derived entirely from data already available in `uv_hub` — no changes to the extension tag API or the `uv_hub` rule attributes are required.

---

### Test plan

- New test cases added


uv/private/uv_hub/repository.bzl
```diff
@@ -169,6 +169,56 @@ exports_files(
...
+)
+""".format(
+                    name = pkg_name,
+                    select = indent(pprint(select_spec), "      ").lstrip(),
+                    compat = repr(proj_cfg_names),
+                ),
+            )
+
+        repository_ctx.file(
+            "project/{}/BUILD.bazel".format(stamp),
```
jbedard, line 209: I still don't like this magic hardcoded "project/" keyword and think this is an indication of a bad API. Is this bad API simply aligning with a python/pip/uv API and the alignment is more important then it being bad? Or is there a better way to do this without hardcoding a magci "project/" keyword?

@gregmagolan and I had a similar discussion in #977 where he was able to find a way to avoid such a keyword...

PR #1091, opened 2026-06-06
fix(platform detection): detect Gentoo glibc

13steinj:
Gentoo's glibc / ldd output doesn't match the pattern that the rest does. In particular, it includes parenthesis in its platform-defined version string, and does not call out that it is glibc at all.

Fixes this hole in libc detection.

Also updates the comments to be more accurate.

---

### Test plan

- Manual testing; use the uv ruleset in a Gentoo, glibc environment (or run `ldd --version` in a bunch of docker images)
  In particular the coworker that reported this uses a [Gentoo Prefix](https://wiki.gentoo.org/wiki/Project:Prefix), but this matches.
  I have not verified (or fixed if it is broken) the output on nixos because it is too annoying for me to try (couldn't find a true container image, only VMs).
  I have also verified that in a musl-based Gentoo, the output for musl is still correctly detected.

### DIFF
--- a/uv/private/host/repository.bzl
+++ b/uv/private/host/repository.bzl
@@ -34,20 +34,21 @@ def _platform(rctx):
 
         out = res.stdout.lower()
 
-        # - Alpine: "musl libc\nversion <ver>\n ..."
+        # - Alpine: "musl libc ...\nversion <ver>\n ..."
         if "musl" in out:
             ver = res.stdout.split("\n")[1].split(" ")[-1].split(".")
             return "musl", "{}.{}".format(ver[0], ver[1])
 
             # - Amazon Linux: "ldd (gnu libc ...) <ver>\n..."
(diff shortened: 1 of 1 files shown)


13steinj:
I use devmoji/gitmoji-- happy to remove the emoji from the PR if that's a problem.

13steinj:
To my knowledge an additional small patch is needed for 1.x
```patch
--- a/uv/private/tomltool/toml.bzl
+++ b/uv/private/tomltool/toml.bzl
@@ -47,5 +47,8 @@
         if "gnu" in ldd or "glibc" in ldd:
             return "gnu"

+        elif "gentoo" in ldd:
+            return "gnu"
+
         elif "musl" in ldd:
             return "musl"
```
or equivalent, but 2.0 uses a different mechanism that doesn't rely on libc for this bit.

jbedard (approved):

PR #1092, opened 2026-06-07
fix(uv): consolidate per-wheel metadata + namespace/regular-span venv merge

jbedard:
Schema foundation: wheel metadata (top_levels, namespace_top_levels, console_scripts) is keyed per wheel by file basename instead of unioned across platform wheels, and the whl_install build rule selects the entry for the wheel the select chain resolved to for the active config. A union leaked an inactive wheel's surface into the active one (another platform's C-extension suffix as a dangling symlink, a win32-only console script wrapper everywhere). A lookup miss (sbuild fallback, failed extraction) emits no PyWheelsInfo and consumers fall back to .pth resolution.

Namespace packages: PEP 420 namespace packages are now materialised CONCRETELY in site-packages via per-entry symlinks under a real namespace directory, from the new per-wheel `namespace_entries` metadata, so tools that inspect site-packages directly (mypy, pyright) see the packages and their py.typed markers — runtime .pth resolution alone left them invisible. Wheels lacking entry metadata (hand-written py_unpacked_wheel) keep the historical .pth-only fallback.

Regular packages spanning wheels: a regular package split across wheels (azure-core owns `azure/core/` while azure-core-tracing-opentelemetry grafts `azure/core/tracing/ext/...` into it) can't be merged by symlinks or .pth — Python locks a regular package's __path__ to one directory. The new per-wheel `namespace_dirs` / `regular_roots` metadata is cross-referenced across wheels to detect the overlap, and the conflicted subtree is physically merged by a new site_merge tool run under the (optional) exec-tools toolchain.

The two assembly mechanisms coexist in _resolve_wheel_collisions: an all-namespace top-level with a spanning regular package goes to the physical merge + .pth (azure); a pure PEP 420 namespace goes to the per-entry concrete symlinks. They never declare conflicting outputs for the same path.

All metadata fields are now `attr.string_list_dict` keyed by wheel basename. e2e: pth-namespace-547 (concrete namespace, mypy-visible), azure-core-tracing-overlap (regular package spanning wheels), and the cffi abi3 snapshot (per-wheel surface) all pass; snapshots regenerated.

---

### Test plan

- New test cases added

### DIFF
--- a/py/private/py_venv/venv.bzl
+++ b/py/private/py_venv/venv.bzl
@@ -23,6 +23,9 @@ distutils, etc.) treat it as a real venv:
         _virtualenv.py                          distutils-compat shim
         _virtualenv.pth                         loads the shim at site init
         <top_level>                             symlink to a wheel's subdir
+        <ns_pkg>/<entry>                        merged PEP 420 namespace: real
+                                                <ns_pkg>/ dir, per-entry symlinks
+                                                into each contributing wheel
         <dist>-<ver>.dist-info                  symlink to a wheel's dist-info
 
 The whole tree is declared at analysis time as individual
@@ -33,6 +36,7 @@ Bazel's action cache treats each piece independently (no tree-artifact
 
 load("@bazel_lib//lib:paths.bzl", "to_rlocation_path")
 load("//py/private:py_library.bzl", _py_library = "py_library_utils")
+load("//py/private/toolchain:types.bzl", "EXEC_TOOLS_TOOLCHAIN")
 
 # Template for console-script wrappers under <venv>/bin/<name>. A tiny shell
 # script that execs the venv's own bin/python with an inline `-c` to import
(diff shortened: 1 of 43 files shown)


py/private/py_venv/venv.bzl
```diff
@@ -371,10 +596,83 @@ def assemble_venv(
...
+                ))
+            tree = wheel_by_key[key].install_tree
+            trees.append(tree)
+            arguments.add_all(
+                [tree],
+                expand_directories = False,
+                before_each = "--src",
+                format_each = "%s/lib/{}/site-packages/{}".format(wheel_py_ver, group.root),
+            )
+        ctx.actions.run(
```
jbedard, line 660: Note this adds an action

xangcastle (approved):

py/private/py_venv/venv.bzl
xangcastle: No performance impact from this action

PR #1098, opened 2026-06-10
fix: disable anyio pytest plugin

xangcastle:
## Problem

When `anyio` is a direct or transitive dependency of a `py_test` target, its pytest plugin auto-registers via the `pytest11` entry point. This causes interference with test collection and execution in environments that don't explicitly configure anyio's async backend.

## Fix

Add and e2e sample with `-p no:anyio` to the pytest args.

---

### Test plan

- New test cases added


e2e/MODULE.bazel
```diff
@@ -9,6 +9,28 @@ bazel_dep(name = "tar.bzl", version = "0.10.1")
 bazel_dep(name = "rules_python", version = "1.0.0")
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "llvm", version = "0.7.1")
```
jbedard: Can we just upgrade this instead of patching it? A newer version started using archive.org as a mirror which I think solves the problem of apple deleting them?

jbedard: and using a newer version for e2e tests is fine, doesn't force newer versions on rules_py users etc

py/private/pytest.py.tmpl
```diff
@@ -58,6 +58,8 @@ if __name__ == "__main__":
     args = [
         "--verbose",
         "--ignore=external/",
+        "-p",
```
jbedard: This is an opt-out of a common package? That sounds like a hack, yet we already do it for another package?

jbedard:
Now this is just an example in e2e? Is that worth merging into 1.x?

PR #1099, opened 2026-06-11
fix: evaluate extra == markers per-extra with OR semantics

xangcastle:
`_decide_marker_impl` was joining multiple extras into a single comma-separated string before passing it to the PEP-508 evaluator, so `extras = ["dev", "other"]` became `extra = "dev,other"` which never matched a marker like `extra == 'dev'`, causing optional dependencies gated by extras to be silently excluded. 
This PR fixes that by evaluating the marker once per extra and ORing the results, so a dependency is included if any of the active extras matches. It also adds venv-derived extras inference: since generated decide_marker calls never pass extras explicitly, the fix now reads the active venv name from the `_venv` flag and extracts matching extra values from the marker string, which handles conflict group markers of the form `extra == 'group-N-project-venvname'`. 
Tests added to `pep508_evaluate_test.bzl` cover positive/negative extra == cases including the comma-join regression, and new decide_marker targets in BUILD.bazel exercise the OR evaluation path directly.

---

### Test plan

- New test cases added

### DIFF
--- a/uv/private/markers/BUILD.bazel
+++ b/uv/private/markers/BUILD.bazel
@@ -170,6 +170,27 @@ decide_marker(
 
 # { name = "psutil", marker = "sys_platform != 'cygwin'" },
 
+decide_marker(
+    name = "test_extra_single_match",
+    extras = ["dev"],
+    marker = "extra == 'dev'",
+)
+
+decide_marker(
+    name = "test_extra_multi_match",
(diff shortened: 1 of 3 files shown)


jbedard (approved):

jbedard:
Please add the same tests (and possibly fixes) to `main` if applicable.

PR #1100, opened 2026-06-11
fix: handle variable tokens immediately followed by brackets in tokeizer

xangcastle:
The PEP-508 marker tokenizer would lose a variable token when a bracket character appeared immediately after it without whitespace. For example, in `'linux' in sys_platform)` the `)` branch would overwrite the accumulated `sys_platform` token, causing the evaluator to receive `)` as a variable name and fail with `key ")" not found in dictionary`.

The fix checks whether the tokenizer is mid-token (VAR or OP state) when a bracket is encountered. If so, it emits the current token first by transitioning to NONE and re-processing the bracket on the next iteration, matching the same emit-then-reprocess pattern already used for whitespace-terminated tokens.

---

- New test cases added

### DIFF
--- a/uv/private/markers/pep508_evaluate.bzl
+++ b/uv/private/markers/pep508_evaluate.bzl
@@ -95,6 +95,9 @@ def tokenize(marker):
 
         char = marker[0]
         if char in _BRACKETS:
+            if state in [_STATE.VAR, _STATE.OP]:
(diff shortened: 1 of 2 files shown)


jbedard (approved):

jbedard:
Please add the same tests (and possibly fixes) to `main` if applicable.

PR #1105, opened 2026-06-11
feat: support packages in git monorepo subdirectories for sdist builds

xangcastle:
When a dependency is pinned from a git monorepo (e.g. pkg @ git+https://github.com/org/repo.git#subdirectory=packages/pkg) uv records a subdirectory field in the lock file source entry. Previously the sdist build pipeline ignored this field entirely, so the wheel build would fail because build_helper.py was trying to run python -m build from the archive root instead of the correct subdirectory where pyproject.toml actually lives. 

This change threads subdirectory from the lock file through the extension layer (defs.bzl), into the sdist_build repository rule as a new attribute, and finally into build_helper.py as a --subdirectory flag that descends into the right directory before invoking the build backend.

---

### Test plan

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -296,6 +296,13 @@ An sdist (if available) will be built into a wheel for installation if no wheels
 are available, or no wheels matching the target configuration are found. Sdist
 builds occur using the configured Python and Cc toolchains.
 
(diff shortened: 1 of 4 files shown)


jbedard (approved):
Able to add any tests that fail without this? Especially if those tests can be ported to 2.x...

PR #1107, opened 2026-06-11
 fix: set SOURCE_DATE_EPOCH and skip dependency check in sdist wheel builds

xangcastle:
---

### Test plan

- New test cases added


xangcastle:
Discarding in pro of the usage of wheel>=0.38

PR #1108, opened 2026-06-11
fix: absolutize CC/CXX paths before chdir into extracted sdist source tree

xangcastle:
Bazel provides the C/C++ compiler path in CC/CXX as a path relative to the exec root. After unpacking the sdist archive, build_helper.py changes the working directory into the extracted source tree so the build backend can find pyproject.toml and run its build hooks. Any relative CC/CXX value becomes invalid the moment cwd changes, causing native extension builds to fail with a "compiler not found" error that is otherwise hard to diagnose. 

The fix resolves both variables to absolute paths immediately after computing outdir (which is itself absolutized for the same reason) and before any chdir occur

---

### Test plan

### DIFF
--- a/uv/private/pep517_whl/build_helper.py
+++ b/uv/private/pep517_whl/build_helper.py
@@ -44,6 +44,12 @@ if opts.patches:
 # Get a path to the outdir which will be valid after we cd
 outdir = path.abspath(opts.outdir)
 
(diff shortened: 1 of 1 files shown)


jbedard (approved):
Do we have some packages we know fail that we can write simple tests for? Just load the packages basically?

PR #1115, opened 2026-06-16
fix: resolve MPICC from PATH instead of defaulting to the Bazel C compiler

xangcastle:
`build_helper.py` previously set `MPICC` to a wrapper around the Bazel C compiler. This breaks MPI sdist builds (e.g. `mpi4py`) because those packages check `$MPICC` before searching `PATH`, and a plain C driver cannot compile MPI code.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                      
Only set `MPICC` when a real `mpicc` exists on the system `PATH`. In that case wrap it so debug-flag stripping is preserved. When no system `mpicc` exists, leave the variable unset so the package can fall back to its own discovery logic.                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                      
This change also adds the `uv-sdist-mpicc` e2e regression test, which drives `build_helper.py` against a synthetic sdist to verify both the "system mpicc found" and "no system mpicc" cases.

---

<!-- If no, please delete this section. -->

### Test plan

- New test cases added


uv/private/pep517_whl/build_helper.py
```diff
@@ -118,7 +118,13 @@ def _compiler_env(tmpdir):
 
     env.setdefault("CC", cc)
     env.setdefault("CXX", cxx)
-    env.setdefault("MPICC", _make_compiler_wrapper(tmpdir, "mpicc", cc_path, sysroot))
+
+    # MPI builds (e.g. mpi4py) consult $MPICC before searching PATH, so a
+    # plain C compiler here would shadow the real mpicc. Only set it when
+    # a system mpicc exists, wrapped to keep the debug-flag stripping.
+    mpicc_path = shutil.which("mpicc", path=env["PATH"])
```
jbedard, line 125: So now we _sometimes_ don't set the `MPICC` env var - when do we _require_ setting it? When there is a custom one manually put on the path?

jbedard (approved):

uv/private/pep517_whl/build_helper.py
jbedard: Does this call have a cost to it? If it's actually invoking `which`?

PR #1117, opened 2026-06-17
feat(uv): add SHA256 hashes for uv 0.11.21 and 0.11.7

daoanhkhoa123:
Add SHA256 mappings for uv release 0.11.21

- https://github.com/astral-sh/uv/releases/tag/0.11.21

No changes aside from this version file for uv.

---

### Test plan

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -79,7 +79,7 @@ to swap a locked requirement (`cowsay`) for a local one.
 bazel_dep(name = "aspect_rules_py", version = "1.6.7") # Or later
 
 uv_bin = use_extension("@aspect_rules_py//uv:extensions.bzl", "uv_bin")
-uv_bin.toolchain(version = "0.11.6")
+uv_bin.toolchain(version = "0.11.21")
 use_repo(uv_bin, "uv")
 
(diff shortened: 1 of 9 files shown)


xangcastle:
Thanks for using our HEAD version, please follow [this example](https://github.com/aspect-build/rules_py/blob/d7e3fea7c27d5fa2b03f66856a393b3dbd32ee2c/e2e/cases/uv-toolchain/setup.MODULE.bazel#L32-L44) to use any UV release, even custom forks

daoanhkhoa123:
May i know the reason why you closed the commits? Since i thought it would be a lot cleaner to add uv versions and shorten the bazel code.

jbedard:
@daoanhkhoa123 I think we want to think about this a bit. Ideally this will be automated, but I think we also don't want to list every version in the world and only list a subset of them.

IMO rules_py should ship with the hashes for the "latest patch of each minor release". Then if you want a different version either:
* you hardcode the hashes yourself
* you hardcode no hashes, they are computed the first time and put into the MODULE.bazel.lock for next time (you must checkin the lockfile so we know the hashes next time)

xangcastle (approved):

PR #1125, opened 2026-06-19
fix: quotes in RECORD entries

jbedard:
RECORD docs: https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file

### Test plan

- New test cases added


jbedard:
I don't know of any real wheels where this is an issue but it was flagged by a user, so I have unit tests but no real e2e test involving a real wheel 🤷

xangcastle (approved):

Issue #1095, opened 2026-06-10
[Bug]: v2 python_interpreters extension doesn't register a py_cc (python/cc:toolchain_type) toolchain, breaking native extension builds


ctcjab:
### What happened?

The v2 `python_interpreters` module extension is positioned as a replacement for `rules_python`'s `python.toolchain()` (it sources hermetic PBS interpreters and registers Python toolchains, so you can drop the `rules_python` dependency). But it registers fewer toolchain types than `python.toolchain()` does — specifically, it does **not** register a Python C/C++ headers toolchain (`@rules_python//python/cc:toolchain_type`).

As a result, once you switch from `rules_python`'s `python.toolchain()` to `python_interpreters` and drop `rules_python`, building any native Python extension (e.g. via [`nanobind-bazel`](https://github.com/nicholasjng/nanobind-bazel)'s `nanobind_extension`, or pybind11) fails at analysis time:

```
ERROR: .../external/rules_python+/python/cc/BUILD.bazel:15:22: While resolving toolchains
for target @@rules_python+//python/cc:current_py_cc_headers: No matching toolchains found
for types:
  @@rules_python+//python/cc:toolchain_type
```

This is because `nanobind-bazel` (and pybind11 setups) resolve Python headers through `@rules_python//python/cc:current_py_cc_headers`, which consumes `@rules_python//python/cc:toolchain_type` — a toolchain type that nothing registers when only `python_interpreters` is in play.

### Root cause

`rules_python`'s `python_register_toolchains` registers **three** toolchains per platform ([`python_register_toolchains.bzl` L161–169](https://github.com/bazel-contrib/rules_python/blob/main/python/private/python_register_toolchains.bzl#L161-L169)):

```python
native.register_toolchains("@{repo}//:{platform}_toolchain")                  # runtime
native.register_toolchains("@{repo}//:{platform}_py_cc_toolchain")            # <-- C headers/libs
native.register_toolchains("@{repo}//:{platform}_py_exec_tools_toolchain")    # exec tools
```

[...]

PR #1131, opened 2026-06-19
fix: support py_cc_toolchain

jbedard:

### Test plan

- New test cases added

### DIFF
--- a/e2e/cases/py-cc-toolchain-1095/BUILD.bazel
+++ b/e2e/cases/py-cc-toolchain-1095/BUILD.bazel
@@ -1,0 +1,72 @@
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
+# Reproduces https://github.com/aspect-build/rules_py/issues/1095
+#
+# The `python_interpreters` module extension is positioned as a drop-in
+# replacement for rules_python's `python.toolchain()`: it sources hermetic PBS
+# interpreters and registers Python toolchains, so a project can drop the
+# `rules_python` dependency. But the hub repo it generates
+# (`@python_interpreters//:BUILD.bazel`) registers only two toolchain types per
+# platform:
+#
+#   - @bazel_tools//tools/python:toolchain_type                      (runtime)
+#   - @aspect_rules_py//py/private/toolchain:exec_tools_toolchain_type (exec tools)
+#
+# It never registers a Python C/C++ headers toolchain
+# (`@rules_python//python/cc:toolchain_type`), which `python_register_toolchains`
+# in rules_python *does* register. nanobind-bazel and pybind11 resolve Python
+# headers through `@rules_python//python/cc:current_py_cc_headers`, which consumes
+# `@rules_python//python/cc:toolchain_type`. With only `python_interpreters` in
+# play, that toolchain type has no registered toolchain and native extension
+# builds fail at analysis time with:
+#
+#   ERROR: .../external/rules_python+/python/cc/BUILD.bazel:15:22: While resolving
+#   toolchains for target @@rules_python+//python/cc:current_py_cc_headers: No
+#   matching toolchains found for types:
+#     @@rules_python+//python/cc:toolchain_type
+#
+# The PBS interpreters already ship the `include/` headers, so (per the issue)
+# "the missing piece is purely the registration".
+#
+# This test captures the contents of the generated hub BUILD file and asserts it
+# registers a `@rules_python//python/cc:toolchain_type` toolchain. It FAILS today
+# (reproducing the bug) and should PASS once `python_interpreters` registers a
+# py_cc toolchain per platform.
+#
+# NOTE: We assert on the registration rather than building a real native
+# extension because, in a graph that still has `rules_python` present, the gap is
+# masked: rules_python's own MODULE.bazel auto-registers py_cc toolchains for its
+# default interpreters, so `current_py_cc_headers` resolves to *those* instead of
+# failing. The registration is the thing `python_interpreters` is missing and the
+# thing it must add to be a true replacement.
+
+genrule(
+    name = "hub_build",
+    srcs = ["@python_interpreters//:BUILD.bazel"],
+    outs = ["hub_build.txt"],
+    cmd = "cp $< $@",
+)
+
+sh_test(
+    name = "test",
+    srcs = ["test.sh"],
+    data = [":hub_build"],
+)
+
+# End-to-end guard: compiling against <Python.h> through
+# @rules_python//python/cc:current_py_cc_headers (exactly how nanobind/pybind11
+# resolve Python headers) requires the py_cc toolchain registered by
+# python_interpreters. This is the concrete native-extension build the issue is
+# about.
+cc_library(
+    name = "python_h_smoke",
+    srcs = ["python_h_smoke.cc"],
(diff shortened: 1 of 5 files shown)


tamird:
My clanker exercised this path through a rules_py 2 integration and
found that a compile-only test can silently resolve rules_python's
fallback C toolchain instead of the generated PBS one.

The replacement commit is
https://github.com/tamird/rules_py/commit/251eb970e87e9b40a770c51e968031bac1aa7f36. It compares the
selected runtime and C-toolchain implementation directly, covers regular
and free-threaded selection, keeps the Windows full and stable import
libraries disjoint, and imports an extension whose `PY_VERSION_HEX`
must match `sys.hexversion`. Free-threaded toolchains omit
`headers_abi3` because CPython rejects `Py_LIMITED_API` there.

That commit is currently stacked on several open prerequisite changes, so
I have not opened a competing PR. Are you okay with me superseding this
draft once those land, or would you prefer to fold the commit into this
branch?

jbedard:
I'm not attached to this branch at all if you have an alternate. Your link is a 404 though...

tamird:
The reviewed replacement is now
https://github.com/tamird/rules_py/commit/4855688ae124b9854cc82bc6ebb005cf9194196e.
It verifies exact platform, runtime, and C-payload selection through public
provider data, exercises regular and free-threaded compile/import paths, and
does not advertise the stable ABI for free-threaded runtimes.

ctcjab:
**Downstream data point: this fixes a real `nanobind_extension` on Linux — not just the `<Python.h>` smoke test.**

We hit #1095 in a monorepo that sources Python toolchains entirely from `rules_py`'s `python_interpreters` (PBS) extension, with **no direct `rules_python` dep**, so nanobind's `@rules_python//python/cc:current_py_cc_headers` had no registered `py_cc_toolchain`.

Tested this branch (`58fc7e46`) via `git_override` on Linux x86_64 with a Zig hermetic cc toolchain, against a real `nanobind_extension` target:

- Analysis resolves cleanly — the prior `No matching toolchains found for @@rules_python+//python/cc:toolchain_type` is gone.
- The extension **compiles + links** against the PBS interpreter's headers and **imports at runtime**, passing under CPython **3.11, 3.12, and 3.13**.

Because this drives the real `current_py_cc_headers` path (compile → link → import), it shouldn't be satisfiable by rules_python's fallback C toolchain — which I believe is the masking risk @tamird raised about the bare `<Python.h>` compile in the PR's e2e case. If it'd help, I'm happy to contribute this nanobind case as an e2e fixture (`nanobind_extension` → import smoke test, ~15 lines) so there's coverage that exercises the generated PBS toolchain end-to-end rather than a bare header compile.

Separately (not a claim about this PR): with #1095 previously blocking analysis, we'd never reached runtime to observe #1048 on this path — and at this commit we see no sign of the `sys.base_prefix` issue on any of 3.11/3.12/3.13.

This (or @tamird's alternative) reaching a mergeable state is the last blocker for our rules_py v2 adoption on native-extension packages, so glad to help validate — happy to re-run against `4855688` once its prerequisites land.

ctcjab:
**Correction + a sharper result after digging into toolchain resolution.**

My earlier "shouldn't be satisfiable by rules_python's fallback C toolchain" was too strong — the masking is real, but *version-dependent*, and I can now show exactly where it does and doesn't bite.

Setup: `python_interpreters` as the rules_python drop-in; `rules_python` present only transitively (resolves to 1.7.0 here), its default `pythons_hub` registering py_cc toolchains for **3.11 only** in this graph. Building a real `nanobind_extension`, per target Python version:

| Target Py | Without this PR | With this PR |
|---|---|---|
| 3.13 | ❌ analysis fails: `No matching toolchains found for @@rules_python+//python/cc:toolchain_type` (rules_python's hub has no 3.13 py_cc) | ✅ resolves `@@aspect_rules_py++python_interpreters+…//:py_cc_toolchain` |
| 3.11 | ⚠️ builds, but **masked** — resolves `@@rules_python++python+python_3_11_…//:py_cc_toolchain`, i.e. rules_python's *own* 3.11 headers, not the configured PBS interpreter | ✅ resolves the PBS `python_interpreters` toolchain |

So the NOTE in the e2e case is right that a real native build can be masked when `rules_python` is present — but only for the Python version(s) its transitive hub happens to register (3.11 here). For any other version (3.13) the gap is unmasked and the build genuinely fails without the registration. And even in the masked case the fix matters: without it, `@rules_python//python/cc:current_py_cc_headers` silently resolves rules_python's *own* interpreter headers rather than the PBS interpreter you configured — a latent header/runtime-interpreter mismatch (the concern @tamird raised, made concrete). With the PR, resolution correctly points at the `python_interpreters` toolchain across 3.11/3.12/3.13. All verified via `--toolchain_resolution_debug`.

Happy to turn this into the repro PR mentioned above: a real `nanobind_extension` e2e case that (a) fails on 3.13 without the registration and (b) asserts the *selected* py_cc toolchain is the `python_interpreters` one rather than rules_python's — so the masked-but-wrong 3.11 case is caught too. Let me know if you'd like it against `e2e/cases/`.

ctcjab:
One clarification for anyone finding this thread later: my testing above was against this branch (`58fc7e46`), which is now ~117 commits behind `main`. Current `main` already registers a `py_cc_toolchain` in `python_interpreters` (`py/private/interpreter/repository.bzl`) and ships a passing `e2e/cases/pbs-cc-toolchain` case that compiles a real extension against `@rules_python//python/cc:current_py_cc_headers` / `current_py_cc_libs` and asserts the exact PBS interpreter (plus abi3 / no-libpython-on-macOS) — so #1095 looks resolved on `main` via that route, not this branch.

Leaving my findings above for reference, but for anyone hitting the py_cc gap: point at `main`. I'm separately checking whether the nanobind-specific path needs any additional coverage on current `main` and will follow up if so.

ctcjab:
Follow-up to the above: opened #1299 with the nanobind e2e case — a real `nanobind_extension` (a `build_test` plus a runtime import/call under a PBS 3.13 runtime) alongside `pbs-cc-toolchain`, covering the `@rules_python//python/cc:current_py_cc_headers` path the way most native extensions actually reach it. Confirms the py_cc registration on `main` works for the nanobind consumer and guards it against regression.

jbedard:
I think #1187 (and maybe #1202?) accomplished everything this PR was trying to do, and the remaining tests here seem to overlap with existing tests.

@ctcjab the test you added in #1299 still shows reproduces a bug that needs fixing, I'm not sure if @tamird is aware of that issue.

tamird:
@jbedard which bug do you mean? #1299's Bazel 8/9 e2e checks ran both nanobind targets successfully. Its deps edge is intentional: version-sensitive extensions must retain the terminal Python configuration, while data resets to the inherited configuration. The explicit imports = ["."] is also intentional under rules_py's -I launcher so the package-local .so has a declared import root. Current pbs-cc-toolchain and cc-deps coverage already verify that the selected runtime and C payload match, including PY_VERSION_HEX == sys.hexversion. Is there another failing configuration or invariant you mean?

jbedard:
@tamird oh, you are correct, it looks like your CC changes also fixed the test in #1299.

@ctcjab can you confirm the latest HEAD fixes your issue?

ctcjab:
@jbedard confirmed — latest HEAD fixes it. Validated against `2.0.0-alpha.5`: the py_cc toolchain rebuilds our nanobind extension per Python version, and the #1299 coverage passes on 3.11/3.12/3.13. Thanks to you and @tamird!

jbedard:
Thanks for verifying 👍

PR #1133, opened 2026-06-19
feat(uv): reserve wheel build resources via uv.override_package(resource_set)

jbedard:

Enable overriding wheel build resources via `uv.override_package(resource_set)`

### Test plan

- New test cases added

### DIFF
--- a/docs/uv-patching.md
+++ b/docs/uv-patching.md
@@ -124,6 +124,11 @@ uv.override_package(
 Resource reservations are only honored when Bazel runs with
 `--experimental_action_resource_set`; add it to your `.bazelrc`.
 
+`resource_set` only applies to packages built from an sdist. Setting it on a
+package that resolves to a prebuilt wheel (no source build) fails the build
+rather than silently dropping the reservation — force a source build with
+`[tool.uv] no-binary-package` if you need the reservation to apply.
+
 ### Full replacement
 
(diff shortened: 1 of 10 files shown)


tamird (commented):
Two accepted paths currently discard `resource_set`:

* a wheel-only lock record never creates an sdist build target; and
* an sdist configure tool that returns `build_file_content` exits before the
  generated `pep517_*_whl` attributes are emitted.

Both should fail when a non-default `resource_set` is present rather than
silently accepting a reservation that no action receives. The geohash e2e
case only proves that the wheel still builds; please also assert the generated
BUILD content or action carries `resource_set = "mem_4g"`.

tamird (commented):
One correction: the flag requirement is stale for the supported Bazel versions. [Bazel c570f978](https://github.com/bazelbuild/bazel/commit/c570f978e5bd67d90318a12e353c6e5ba099da36) made `resource_set` callbacks unconditional and moved `--experimental_action_resource_set` to the graveyard as a default-true no-op. That behavior is in Bazel 8.5.1, which this repository pins, and remains in Bazel 9.

Please remove the `e2e/.bazelrc` addition and the flag requirement from `docs/uv-patching.md` and the attribute docs. The `resource_set` API itself should remain unchanged.

acozzette (approved):

PR #1134, opened 2026-06-22
fix(uv): preserve compiler commands through native wheel builds

jbedard:

Use native bazel c toolchains for CC and CXX

### Test plan

- New test cases added


tamird:
I do not think `CcToolchainInfo` exposes a reusable compiler-command
contract for arbitrary PEP 517 backends. Bazel's configured compile and link
actions have Bazel C++ semantics; distutils, Meson, CMake, and other backends
own different driver and linker behavior after changing into the source tree.
Projecting those actions requires the sysroot, token, driver, and path
heuristics this change has accumulated.

#1152 removes that projection. Packages declare `build_toolchains` and `env`
explicitly; rules_py forwards their inputs and Make variables, makes only
exact whole-value declared tool paths absolute before `chdir`, and leaves all
other values opaque.

PR #1135, opened 2026-06-22
refactor(py_venv): drop _wheels/<key> indirection to deduplicate runfiles

jbedard:
IIUC (and the robots are correct) the original double-hop design (467c24c4) is no longer required with the latest `py_image_layer` rewrite (dd317be2). 

>   site-packages symlinks and .pth entries now point directly at each wheel's natural
>   runfiles path instead of routing through an intermediate <venv>/_wheels/<key>/ tree-artifact alias.
> 
>   The two-hop design (467c24c4) made the venv self-contained in
>   bazel-bin so py_image_layer could walk it there. But py_image_layer was since
>   rewritten to source wheel content from PyWheelsInfo.install_tree directly, so
>   it no longer follows the venv symlinks in bazel-bin — the alias is dead weight.
> 
>   The alias also materialized in runfiles, where Bazel duplicated every wheel
>   file at the alias path AND at its natural rfpath — doubling on-disk runfiles
>   for every uv-built wheel. Worse, because the alias was a distinct declared
>   output, py_image_layer's source-layer dedup (keyed on the natural path) never
>   matched it, so the full tree shipped unfiltered into images — re-injecting the
>   __pycache__/.pyc and install metadata that _should_skip_pkg_path deliberately
>   strips from the pip-package layer.
> 
>   Removing the alias preserves the cache-independence #1038's hash key provided
>   (each wheel's natural rfpath is equally independent of sibling wheels) while
>   shedding the duplication and the bytecode-smuggling. The only thing lost is
>   bazel-bin self-containment of `bazel build :foo.venv`, which no real consumer
>   relies on (verified across the e2e py_image_layer / py_venv_image_layer /
>   uv-deps-650 cases).

### Test plan

- New test cases added


thesayyn (approved):

tamird (changes requested):
This can execute a wheel's root-level `.pth` files twice. For an
`install_tree` wheel that loses any top-level collision, `_format_imp` now
uses `site.addsitedir()` on the natural wheel root. Its `.pth` files are
already projected into the venv by `top_level_to_site_pkgs`, so site startup
executes the projected copy and then `addsitedir()` executes the original.

Please preserve the old split after removing `_wheels/<key>`: use a plain
natural runfiles path for known `install_tree` wheels, whose root `.pth`
files were projected, and reserve `site.addsitedir()` for metadata-unknown
wheels. A regression fixture should force a known wheel onto the fallback
path and assert its `.pth` side effect occurs once.

The PR also deliberately drops the materialized bazel-bin venv contract, so
`Changes are visible to end-users` should be `yes` even if no current e2e
consumer relies on that layout.

jbedard:
https://github.com/aspect-build/rules_py/pull/1138 adds a test that I think reproduces what @tamird describes

tamird (approved):
The current head addresses the duplicate `.pth` execution, and the merged #1138 regression exercises that path. One description fix remains: removing bazel-bin venv self-containment is user-visible, so the PR body should say `yes`.

jbedard:
@tamird I'm still uncertain if the venv self-containment is problematic or minor enough it is worth this simplification? I do like deleting so much code though 😅

tamird:
I think "self-contained" overstates what exists today. `_wheels/<key>`
is itself an absolute symlink to another bazel-bin artifact, and the venv
already escapes its own tree for the
[PBS interpreter](https://github.com/aspect-build/rules_py/blob/e14d7b7400344b0eeceb0ff678cc9be8ad5c58e8/py/private/py_venv/venv.bzl#L755-L790),
[first-party imports, and unkeyed wheels](https://github.com/aspect-build/rules_py/blob/e14d7b7400344b0eeceb0ff678cc9be8ad5c58e8/py/private/py_venv/venv.bzl#L676-L701).
The property being removed is narrower: install-tree wheels are walkable
through the bazel-bin venv.

There is one real consumer to handle before landing this:
`py_venv_link` currently points IDEs at that bazel-bin venv, where direct
runfiles-relative wheel links would dangle.
[21bdebdb](https://github.com/aspect-build/rules_py/commit/21bdebdbec3ab938d810599abee218c37402be29)
moves `py_venv_link` to the complete runfiles tree, which already owns the
interpreter, first-party sources, and wheels.

Once that lands first, I do not see a reason to preserve `_wheels`.
`py_image_layer` now reads each
[`install_tree` directly](https://github.com/aspect-build/rules_py/blob/168ca70fc4ec638b4fb5c943c83893232b542cd7/py/private/py_image_layer.bzl#L348-L365),
and supported runtime consumers have the complete runfiles closure. The
description should state that ordering rather than claim no consumer relies
on the bazel-bin layout.

tamird:
@jbedard The rebase failure is a stale fixture assertion, not an
import regression. The test requires the `airflow` loader filename to sit
under the hidden venv, while this PR deliberately replaces that path with
the wheel's natural `actual_install.install` runfiles path. Both targets
pass on `main` through `_wheels`; on cd00101a they fail only that
substring assertion.

https://github.com/tamird/rules_py/commit/6bcd89208761c38f94b230fca39990d274c2db2f
replaces the deprecated `pkgutil` path check with an
`importlib.util.find_spec` discoverability check. Both failed targets pass
locally under Bazel 8 again, with their prior 0.1-second test runtime.

tamird:
We need this to complete a downstream rules_py 2 upgrade. Without it,
each uv wheel remains in runfiles both at its natural install-tree path and
under `_wheels/<key>`. In our image closure that duplicates wheel trees and
prevents `py_image_layer`'s natural-path deduplication and filtering from
recognizing the alias copy, so content it normally omits is retained.

This is therefore load-bearing for the upgrade, not an optional cleanup. I
will redo the patched-wheel topology change from `main` after this lands
rather than stack another PR on this branch.

xangcastle (approved):

PR #1136, opened 2026-06-22
feat(uv): expose per-group deps

jbedard:
### Test plan

- New test cases added

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -134,6 +134,33 @@ py_binary(
 )
 ```
 
+Targets that need every dependency in one dependency group can use the
+group-specific lists generated in `defs.bzl`. The `group_deps()` helper follows
+the consuming target's `dep_group`, so the group name is never repeated:
+
+```starlark
+load("@pypi//:defs.bzl", "group_deps")
+
+py_binary(
+    name = "all_vendored",
+    dep_group = "vendored_say",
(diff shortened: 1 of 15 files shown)


jbedard:
This is an alternative to https://github.com/aspect-build/rules_py/commit/941bd5c917335730125a3aa1ea43e26b0e02adb0 which modified the `requirements()` API that I would rather move away from (and not diverge from rules_python where we do keep it).

tamird (commented):
The API is useful, but the generated implementation does unnecessary work at
two levels: repository generation scans every package for every group, and
every `group_deps()` call rebuilds all dependency labels plus the complete
`select()`.

Please invert the existing sparse package-to-group membership once while
generating the hub, emit canonical dependency-label lists per group, and
construct one module-level `_GROUP_DEPS = select(...)`. `group_deps()` can
then return that immutable value. The current functional fixtures are enough
for behavior; their generated `defs.bzl` snapshots should pin the precomputed
shape.

jbedard:
@tamird I think all your feedback is addressed, I don't think any of the API is effected by that which is my primary concern atm...

xangcastle (approved):

tamird (approved):
The revised implementation addresses my feedback. The per-group membership inversion, canonical labels, and module-level `select()` all look correct.

PR #1140, opened 2026-06-23
feat(uv): sample wheel build memory usage when resource_set is set

xangcastle:
Adds a memory-monitoring layer to sdist wheel builds. When a `resource_set` other than `"default"` is declared, the wheel build action now runs under a sampler that tracks aggregate resident-set size (RSS) of the build process  and all its descendants, and reports the peak to stderr.                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                            
This gives users concrete data to calibrate their `resource_set` estimates. Rather than guessing whether `"mem_2g"` is enough, they can observe what the build actually consumes. 

---

### Test plan
- New test cases added


tamird:
The `resource_set` reservation is the right scheduler API and is already
on `main`. The monitor is still useful for post-OOM tuning, but its trigger
and process ownership should be narrower:

- `resource_set` should not implicitly enable monitoring. Reservation and
  diagnostics are independent; use an explicit `monitor_memory` opt-in.
- Do not use `start_new_session` or kill the backend process group. Bazel owns
  the action process group and its cancellation semantics. Launch the backend
  normally and sample its descendants on a best-effort basis.

The output should remain explicitly approximate Linux procfs RSS, not a hard
limit or a remote-executor sizing mechanism.

tamird:
Superseded by #1174. The replacement is rebuilt on current `main`, makes
monitoring an explicit opt-in independent from `resource_set`, and leaves
process-group cancellation to Bazel.

jbedard:
@xangcastle close?

PR #1142, opened 2026-06-23
feat(uv): describe installed wheel topology metadata

jbedard:
Plumb two pieces of descriptive metadata through repository-phase
extraction for downstream venv analysis:

  * `directory_top_levels`: the subset of each wheel's top-levels that
    install as directories (vs single-file modules), exposed as a new
    `whl_install` attr.
  * console_scripts is now emitted for every active wheel, including an
    empty list, so a consumer can distinguish a wheel with no scripts
    from one whose metadata was never extracted.

Neither has a consumer yet; this is groundwork. Regenerates the affected
BUILD snapshots.

### Test plan

- New test cases added

### DIFF
--- a/e2e/snapshots/abi3_compat.whl_install.cffi.BUILD.bazel
+++ b/e2e/snapshots/abi3_compat.whl_install.cffi.BUILD.bazel
@@ -107,6 +107,56 @@ whl_install(
             "cffi-2.0.0.dist-info",
         ],
     },
+    directory_top_levels = {
+        "cffi-2.0.0-cp312-cp312-macosx_10_13_x86_64.whl": [
+            "cffi",
+            "cffi-2.0.0.dist-info",
+        ],
+        "cffi-2.0.0-cp312-cp312-macosx_11_0_arm64.whl": [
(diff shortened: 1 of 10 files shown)


tamird:
The reduced patch still underclassifies regular packages because it only
recognizes a literal `__init__.py`. Repository analysis produces metadata
for other target platforms and for static tooling, so the repository host's
active loader suffixes are not enough.

Please conservatively recognize `__init__.py`, `.pyc`, `.pyi`, `.pyw`, and
native `__init__` modules ending in `.so`, `.pyd`, or `.fwork`, and keep the
classification in one tested `wheel_layout_from_record` helper. The audited
version is
https://github.com/tamird/rules_py/commit/56aad741b9279c03d6009b9ff3219111d9f3198e.

The `.data` routing and selected-wheel basename fixes are already on `main`,
so this PR can stay focused on the remaining topology contract.

tamird:
Superseded by #1172, which isolates RECORD routing and carries the selected wheel's directory metadata through `PyWheelsInfo`.

PR #1148, opened 2026-06-23
fix(uv): isolate source-build Python

tamird:
Filter `PYTHONHOME`, `PYTHONPATH`, and `PYTHONPLATLIBDIR` from the
inherited action environment before invoking a PEP 517 backend. Preserve
unrelated variables and allow an explicit package environment to set them
again.

`pyproject-hooks` copies its process environment and launches the backend
without isolated mode, so host Python settings can replace the managed
build environment or standard library.


uv/private/pep517_whl/BUILD.bazel
```diff
@@ -22,6 +26,37 @@ bzl_library(
...
+    deps = ["@pypi//build"],
+)
+
+_PYTHON_ENV_BACKEND_SRCS = glob(["python_env_backend/*"])
+
+mtree_spec(
+    name = "__python_env_sdist_mtree_raw",
+    srcs = _PYTHON_ENV_BACKEND_SRCS,
+    include_runfiles = False,
+    tags = ["manual"],
```
jbedard: what can be tagged as `testonly = True` instead of `tags = ["manual"]`?

tamird: Done in 0e779f44. The helper, sdist, and transition-wrapper targets are now `testonly`. The two raw wheel targets remain `manual` as well because they intentionally fail under the default action environment and must only be reached through the hostile-environment transition.

uv/private/pep517_whl/BUILD.bazel
```diff
@@ -22,6 +26,37 @@ bzl_library(
...
 
+py_binary(
+    name = "__build_helper",
+    testonly = True,
+    srcs = ["build_helper.py"],
+    main = "build_helper.py",
+    deps = ["@pypi//build"],
+)
+
+_PYTHON_ENV_BACKEND_SRCS = glob(["python_env_backend/*"])
```
jbedard: If this entire `python_env_backend` dir is only for tests can we put it under a `tests/` subdir to make that clear?

tamird:
Extracted from #1151. Source-built wheels in the parent use a managed
interpreter, but `pyproject-hooks` copied host Python path variables
into backend processes, which could replace the standard library.

uv/private/pep517_whl/BUILD.bazel
tamird: Done.

jbedard (approved):

PR #1149, opened 2026-06-23
fix(py): tolerate incompatible wheel source

tamird:
Some wheels intentionally retain source for older interpreters. For
example, aenum 3.1.15 includes `aenum/_py2.py`; `compileall` compiles
compatible modules but reports a false aggregate result. Failing the
install on that result rejects a usable wheel.

Invoke `compileall.main` in a child interpreter to preserve per-file
diagnostics without turning its boolean into process status. Keep
`check=True` so abnormal interpreter exits still fail the action,
matching pip.


jbedard:
Can we add an e2e test reproducing the original issue and proving the fix? You mention `aenum 3.1.15` having the issue, how can that be used to reproduce the issue?

tamird:
@jbedard Added in 61584a5f. `aenum==3.1.15` is the real reproducer: its Python 3 wheel retains a Python 2-only [`_py2.py`](https://github.com/ethanfurman/aenum/blob/3.1.15/aenum/_py2.py#L1-L7), while [`__init__.py`](https://github.com/ethanfurman/aenum/blob/3.1.15/aenum/__init__.py#L31-L36) selects `_py3.py` at runtime. The new e2e installs that published wheel and exercises `aenum.Enum`; before this change, recursive `compileall` rejects the wheel before the test runs.

jbedard (approved):

tamird:
Extracted from #1151. The parent precompiles wheel sources, and aenum
3.1.15 retains Python 2-only `_py2.py`; `compileall` reported failure
despite compiling every compatible module and leaving the wheel usable.

PR #1150, opened 2026-06-23
fix(py): support dotted console entry points

tamird:
Entry-point object references may address a dotted attribute, and script
names are case-sensitive. Installed-wheel and generated-venv wrappers
currently render `from module import object`, which is invalid when the
object contains a dot; `ConfigParser` also lowercases installed script
names by default.

Preserve names while parsing `entry_points.txt` and resolve objects with
`import_module` plus `attrgetter` in both wrapper paths. Continue ignoring
legacy extras as before.

### DIFF
--- a/e2e/cases/venv-bin-scripts-423/BUILD.bazel
+++ b/e2e/cases/venv-bin-scripts-423/BUILD.bazel
@@ -8,6 +8,7 @@ py_test(
     python_version = "3.11",
     deps = [
         "@pypi_venv_bin_scripts_423//dice",
+        "@pypi_venv_bin_scripts_423//invoke",
     ],
 )
 
@@ -20,6 +21,7 @@ py_test(
     python_version = "3.11",
     deps = [
         "@pypi_venv_bin_scripts_423//dice",
+        "@pypi_venv_bin_scripts_423//invoke",
     ],
 )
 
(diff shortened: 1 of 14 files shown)


jbedard:
What is a real package that reproduces this? Can that be added to a test in `e2e/`?

tamird:
@jbedard Added in 6e05db6e. `invoke==2.2.0` publishes `invoke.main:program.run` ([source](https://github.com/pyinvoke/invoke/blob/2.2.0/setup.py#L46-L50)). The existing venv-bin-scripts e2e now installs it and runs `invoke --version`; the old wrapper rendered that object path as invalid Python syntax.

jbedard (approved):

tamird:
Extracted from #1151. Console wrappers in the parent encountered dotted
entry-point objects, which cannot be rendered as `from module import
object`; installed script parsing also changed case-sensitive names.

jbedard:
Needs rebase, which should also fix the buildifier error

tamird:
@codex review

tamird:
Rebased in 3711718e. Buildifier and all functional checks are green. The remaining red benchmark measured +2.2% with unchanged `sys.path` quality, then failed when the fork token could not post its result (403).

xangcastle (approved):

PR #1151, opened 2026-06-23
fix(py): align PBS runtime and build tools

tamird:
Generate runtime, Python C, and exec-tools registrations from the same
Python Build Standalone release, version, build configuration, and target
constraints. Resolve exec tools for the execution platform while keeping
the runtime and headers on the requested target platform.

The previous extension used separate local interpreter and C-toolchain
paths, and matched exec tools by only major/minor plus build mode. That
could combine different PBS releases or Python patch versions; CPython can
change bytecode magic without changing its major/minor cache tag.

Use rules_python's public exec-tools provider, normalize version defaults
once, and register exact target and executor companions from one release
cohort. Omit an exec registration when that cohort has no companion for
the executor platform.


tamird:
Closing this stale umbrella. The runtime selection, exec-runtime binding, C toolchain, bytecode identity, and source-build tool changes now have independent review units. This combined draft also carries obsolete copies of merged prerequisites and is no longer a branch that should land.

PR #1153, opened 2026-06-23
feat(uv): declare source-built scripts

tamird:
The source-script change is 02067ff6. Earlier commits on this branch
are independently submitted prerequisites and will be removed as they
reach `main`.

Source-built wheels do not exist until execution, but venv assembly
must create console-script wrappers during analysis. Downloaded-wheel
scripts can be inspected during repository evaluation; source-built
scripts therefore need an explicit analysis-time declaration.

Allow `override_package` to declare a complete name-to-entry-point map.
The generated install repository marks every source-build fallback
before wheel selection and attaches declarations only to that arm.
Prebuilt arms retain inspected metadata, custom source producers remain
opaque, and source topology stays unknown even if a producer reuses a
locked wheel filename.

The map is trusted analysis input and must be kept in sync with the
source-built wheel.

### DIFF
--- a/docs/uv-patching.md
+++ b/docs/uv-patching.md
@@ -10,7 +10,10 @@ There are three kinds of overrides:
 - **Pre-build patches** (`pre_build_patches`): Patch the extracted source distribution before building a wheel. Useful for fixing build scripts or source code.
 - **Post-install patches** (`post_install_patches`): Patch the installed package tree after wheel unpacking. Useful for fixing installed library code.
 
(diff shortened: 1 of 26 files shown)


tamird:
Extracted from #1151. The parent assembles virtual environments during
analysis, but source-built wheels do not exist until execution, so their
console scripts need an explicit analysis-time declaration.

xangcastle (approved):

PR #1154, opened 2026-06-23
fix(py): preserve permissive collision order

tamird:
rules_py 1.11.2 resolved permissive venv collisions by keeping the
last command for each destination. The analysis-time assembler selects
the first metadata claimant and leaves wheel depset order implicit,
reversing that precedence for known wheel layouts.

Aggregate wheel metadata in left-to-right postorder and select the last
distinct claimant for ordinary top-level paths, exact namespace entries,
and console scripts. Build regular-package merge groups from the same
sequence and overlay their inputs in order.

This is sequence-based precedence, not direct-requirement priority.
Depsets remove duplicate elements, so another dependency edge does not
refresh its position. PEP 420 top-levels continue to merge. Namespace
prefix overlaps retain the shallower path, and wheels outside
`PyWheelsInfo` remain on the `.pth` fallback. Error mode continues to
reject collisions.

### DIFF
--- a/e2e/cases/uv-include-group/BUILD.bazel
+++ b/e2e/cases/uv-include-group/BUILD.bazel
@@ -4,8 +4,8 @@ load("//tools:pth_snapshot.bzl", "extract_venv_pth")
 # setuptools vendors `packaging` (among others), which collides at
 # the top-level with the real packaging wheel pulled in via pytest.
 # Not a PEP 420 namespace on either side, so our analysis-time check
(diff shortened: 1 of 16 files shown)


tamird:
Extracted from #1151. That parent replaced the permissive last-command
selection from version 1.11.2 with selection of the first metadata
claimant while leaving wheel depset order implicit.

jbedard (approved):

xangcastle (approved):

PR #1155, opened 2026-06-23
fix(py): merge ambiguous namespace layouts

tamird:
The branch carries separately submitted prerequisites through 013ce7dd;
the namespace-layout change is 826d8550. Those prerequisite commits will
be removed as they reach `main`.

Project RECORD-derived namespace entries only when every claimant is an
unpatched downloaded wheel and the entries are pairwise disjoint. Merge
the complete top-level otherwise so nested regular packages, source-built
wheels, and hand-written wheels match a flat install.

Treat ordinary top-levels atomically and remove every known-layout wheel
from the `.pth` fallback after collision resolution. This prevents a
permissive loser from reappearing at runtime while preserving the
action-free path for disjoint downloaded namespaces.

Detect overlap by checking each slash-delimited ancestor rather than only
adjacent sorted entries, so an intervening sibling cannot hide an
ancestor/descendant collision.


tamird:
Extracted from #1151. The parent's wheel-layout work exposed namespace
collisions where nested symlinks were unsafe or incomplete, requiring
atomic ownership unless complete metadata proves a disjoint PEP 420
layout.

tamird:
Closing this stale umbrella. Collision ordering is already upstream, while directory topology, patched-layout ownership, and directory overlay are now separate review units. This draft carries drifted copies of merged prerequisites and no longer represents a patch that should land.

PR #1157, opened 2026-06-23
refactor(py): use rules_python exec tools

tamird:
rules_py copied rules_python's exec-tools rule so wheel actions could
obtain the exec-configured `PyRuntimeInfo` while supporting version
1.0.0. rules_python 1.9.0 now exposes `exec_runtime` and the rule and
toolchain type through its public API:
https://github.com/bazel-contrib/rules_python/blob/1.9.0/python/py_exec_tools_toolchain.bzl
https://github.com/bazel-contrib/rules_python/blob/1.9.0/python/private/py_exec_tools_info.bzl#L48-L66

Raise the dependency floor to 1.9.0 and register PBS exec tools with the
public type. Remove the copied provider, rule, executable wrapper, and
toolchain type so rules_py and rules_python share one exec-tools
contract.


jbedard:
Should we bump to rules_python 2.0.0? What would the advantages vs drawbacks be?

EDIT: I think it should be 1.9.0 like this PR currently does

jbedard (approved):

tamird:
Extracted from #1151. Exact target and execution runtime pairing in the
parent needs exec-configured `PyRuntimeInfo`; rules_python 1.9.0 now
provides that contract publicly, making rules_py's copied rule and
provider redundant.

tamird:
Rebased onto 910b228c.

xangcastle (approved):

PR #1159, opened 2026-06-24
fix(py): validate interpreter inputs

tamird:
Require each toolchain python_version tag to be a decimal major.minor,
the actual selection granularity, and reject malformed root-owned PBS
release identifiers before fetching a release manifest.

The exact-runtime cohort work uses both values as structured selection keys.
Previously, 3.14.3garbage silently selected 3.14, making exact-looking input
misleading, while 2026-03-03 became a failed network request. Keep
pre-release policy on the explicit pre_release attribute. Validate every
module's toolchain requests, but continue ignoring release configuration from
dependencies because the root module owns the release search space.

### DIFF
--- a/py/private/interpreter/extension.bzl
+++ b/py/private/interpreter/extension.bzl
@@ -1,7 +1,7 @@
 """Module extension for provisioning Python interpreters from python-build-standalone."""
 
 load(":repository.bzl", "local_python_interpreter", "python_interpreter", "python_toolchains")
-load(":version_util.bzl", "is_pre_release", "is_valid_python_tag", "version_gt")
+load(":version_util.bzl", "is_decimal", "is_pre_release", "version_gt")
 load(":versions.bzl", "BUILD_CONFIGS", "DEFAULT_RELEASE_BASE_URL", "DEFAULT_RELEASE_DATES", "PLATFORMS")
 
 # The GitHub API endpoint for resolving "latest" releases.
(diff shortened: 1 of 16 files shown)


tamird:
Extracted from #1151. That PR needs Python version tags and PBS release identifiers to be valid selection keys before it can pair exact target and execution runtimes.

py/private/interpreter/extension.bzl
```diff
@@ -185,6 +185,14 @@ def _python_interpreters_impl(module_ctx):
                     is_reproducible = False
                     date = _resolve_latest(module_ctx, base_url)
                     resolved_latest = date
+                invalid_release = len(date) != 8
+                for char in date.elems():
```
jbedard: isn't this the same as `_is_decimal`?

py/private/interpreter/version_util_test.bzl
```diff
@@ -1,7 +1,32 @@
...
+        "3.14.3",
+        "3.15.0a6",
+        "3.14.0b1",
+        "3.14.0rc2",
+    ]:
+        asserts.true(env, is_valid_python_tag(version), version)
+
+    for version in [
+        "3",
+        "3.14.3.1",
```
jbedard: Why would we care if a version is tagged as something odd like this? Don't we just need the `3.14` for the main python version and otherwise don't care about `-foobar` or `.3.4`?

py/private/interpreter/extension.bzl
tamird: Yes. I made the decimal predicate public and reuse it for both release identifiers and the two components of python_version.

py/private/interpreter/version_util_test.bzl
tamird: The tag only controls major/minor and never pinned a patch, so accepting suffixes was misleading. I changed it to require major.minor; pre-release selection now uses the existing pre_release flag. Exact PBS-version validation stays with the later manifest parser.

jbedard (approved):

xangcastle (approved):

PR #1162, opened 2026-06-24
refactor(py): retain PBS release versions

tamird:
Retain every full Python version for each platform and build configuration
in a PBS release index. The current parser collapses each entry to its newest
full version. When platform versions differ, that loses the older artifact
needed to describe an exact target/execution interpreter cohort.

Keep runtime selection unchanged: choose the first configured release that
contains the key, then its newest full version. Version the module-extension
facts key so cached indices are refetched in the new shape.

Malformed and duplicate manifest lines keep their existing permissive
behavior. Rejecting them would be a separate behavior change for mirrors.

### DIFF
--- a/py/private/interpreter/BUILD.bazel
+++ b/py/private/interpreter/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@bazel_lib//:bzl_library.bzl", "bzl_library")
 load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "string_flag")
 load(":exclude_feature.bzl", "exclude_feature_flag")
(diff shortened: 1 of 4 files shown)


tamird:
Extracted from #1151. Exact target/execution cohorts need older platform
companions that the current release index discards when another patch version
is newer.

tamird:
Closing because exact cohort generation is no longer part of the upgrade. The OpenAI consumer selects one PBS release and one exact final Python version, so retaining every artifact in each release index has no independent consumer. A separate bytecode-compatibility guard can reject mismatched target and execution interpreter metadata without expanding the index.

PR #1163, opened 2026-06-24
fix(uv): retain anyarch validation

tamird:
Source-build configuration depends on `sdist_build`'s pure/native
classification, but its generated target sets `args = []` for either
result. This overrides the `--validate-anyarch` default on `pep517_whl`,
so an archive classified as pure can emit a platform-specific wheel
without failing the fallback check.

Let generated targets inherit their rule-owned defaults. Pure builds
keep `--validate-anyarch`, while native builds keep an empty argument
list.


tamird:
Extracted from #1151. Source-build configuration in the parent exposed
that generated pure-wheel targets override `pep517_whl`'s
`--validate-anyarch` default with `args = []`, allowing a misclassified
source distribution to emit a platform wheel without failing.

jbedard (approved):

PR #1165, opened 2026-06-24
fix(py): detect execute-bit collisions

tamird:
Physical package merges consider byte-identical files non-conflicting
and adopt the later file mode. That silently accepts a collision when one
wheel marks a file executable and another does not, so strict collision
policy can produce order-dependent runtime behavior.

Treat executability changes as collisions while preserving the
permissive later-file overlay. Compare executable status rather than
exact owner, group, and other execute masks: the wheel extractor already
normalizes any source execute bit to executable, while read and write
differences can vary with extraction permissions.

### DIFF
--- a/py/tools/site_merge/site_merge.py
+++ b/py/tools/site_merge/site_merge.py
@@ -82,10 +82,14 @@ def merge(into, sources):
                     _remove(dest)
                 prior = owners.get(rel)
                 if dest.exists():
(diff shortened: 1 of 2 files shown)


tamird:
Extracted from #1154. That parent made byte-identical package files
adopt the later wheel's mode, exposing that strict collision handling
treated a change between executable and non-executable files as benign.

xangcastle (approved):

PR #1167, opened 2026-06-24
fix(uv): align PEP 517 frontend platform

tamird:
PEP 517 native-wheel actions run in the named `target` execution
group, whose native build toolchain constrains execution to the target
platform. The frontend executable instead uses the default exec
transition.

For example, a Linux/aarch64 sdist target on an x86_64 host can configure
the frontend for the host while scheduling the wheel action on the
Linux/aarch64 execution platform. The action then receives a launcher or
runtime that cannot execute there.

Use `config.exec("target")` so the frontend and wheel action are
configured for the same named execution platform.


tamird:
Extracted from #1152. The source-build tools patch configured its new
build-tool targets for the wheel action's named execution group, exposing
that the existing PEP 517 frontend still used the default exec transition.

jbedard:
What's an example scenario this fixes? Do we need some tests with different target vs exec configs to test this better?

tamird:
A concrete failure is a Linux/aarch64 native-sdist target on an x86_64 host: `NATIVE_BUILD_TOOLCHAIN` forces the wheel action onto the target platform, while `cfg = "exec"` can configure the frontend for the host. The target action then receives an incompatible launcher or runtime.

Added a production-action regression in bf33e1c6. It chooses a target CPU opposite the host, runs the real `pep517_native_whl` action, and has the configured frontend identify its platform in the declared output. It fails with `cfg = "exec"` and passes with `config.exec("target")`.

e2e/cases/pep517-frontend-exec-group/test.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
jbedard: Can this be tested with a `build_test` instead of `test.sh`?

tamird: Not without broadening the fixture scope. A transition wrapper can set the target platform for a `build_test`, but it cannot add an execution platform only for that target. This regression requires Bazel's default exec group to select the host while the named `target` group selects a synthetic opposite-architecture platform, so the fixture must pass both `--platforms` and `--extra_execution_platforms` and inspect the frontend's emitted output. `--extra_execution_platforms` is invocation-wide; registering that synthetic platform in the e2e module would make it eligible for unrelated actions. I added this rationale and the Bazel option citation to the script.

jbedard: Can a `build_test` not test a target that goes through transitions? See rules_js for examples

tamird: Yep, you were right. I treated `extra_execution_platforms` as invocation-global, but platform flags are scoped to the transitioned configuration.

https://github.com/aspect-build/rules_py/commit/e497c2011ca7cd77371cbed1a3858e0f54e1b4d0 replaces `test.sh` with one `build_test` over two `platform_transition_filegroup` targets. Each selected platform registers the host first and itself second. The default exec group therefore selects the host, while the native wheel's constrained `target` group selects the synthetic target. Building both CPU variants guarantees one differs from the host, with no `uname` branch or recursive Bazel invocation.

e2e/cases/pep517-frontend-exec-group/BUILD.bazel
```diff
@@ -0,0 +1,113 @@
...
+# https://bazel.build/reference/be/platforms-and-toolchains#platform
+platform(
+    name = "linux_aarch64",
+    constraint_values = [
+        "@platforms//cpu:aarch64",
+        "@platforms//os:linux",
+    ],
+    flags = [
+        "--extra_execution_platforms=@platforms//host,//cases/pep517-frontend-exec-group:linux_aarch64",
+    ],
```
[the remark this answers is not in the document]
tamird: Fixed in [6dbf8f3f](https://github.com/aspect-build/rules_py/commit/6dbf8f3fbd7a5ffad2d33d9bf80919feb1bcf7de). Both transitioned Linux platforms now set `platform_libc=glibc` in their scoped flags, matching the PBS Linux interpreter toolchain target settings.

jbedard (approved):

jbedard:
rebase

xangcastle (approved):

xangcastle (approved):

PR #1168, opened 2026-06-24
fix(uv): isolate PEP 517 runfiles

tamird:
The PEP 517 helper is a Bazel executable. Its launcher exports
RUNFILES_DIR, RUNFILES_MANIFEST_FILE, and JAVA_RUNFILES for the helper
runfiles tree. Bazel also sets RUNFILES_MANIFEST_ONLY when directory
runfiles are disabled. build_helper.py copies that identity into the
backend environment.

A Bazel executable invoked by the backend then prefers the inherited
source to its own adjacent runfiles. It resolves its interpreter in the
helper tree and fails with execve/ENOENT.

Remove all four variables from the copied backend environment. The
helper retains its own environment, while tools invoked by the backend
discover their own materialized runfiles.

### DIFF
--- a/uv/private/pep517_whl/tests/python_env_backend/backend.py
+++ b/uv/private/pep517_whl/tests/python_env_backend/backend.py
@@ -13,9 +13,19 @@ _DIST_INFO = "python_env_backend-0.0.1.dist-info"
 
 
 def _check_environment() -> None:
-    for name in ("PYTHONHOME", "PYTHONPLATLIBDIR"):
+    # JAVA_RUNFILES is Bazel's legacy runfiles locator, not a Java toolchain
+    # variable. Compiler and Java tool variables remain available to backends;
+    # only the parent launcher's runfiles identity must be removed.
+    for name in (
+        "JAVA_RUNFILES",
+        "PYTHONHOME",
+        "PYTHONPLATLIBDIR",
+        "RUNFILES_DIR",
+        "RUNFILES_MANIFEST_FILE",
+        "RUNFILES_MANIFEST_ONLY",
+    ):
         if name in os.environ:
-            raise RuntimeError(f"host {name} reached the PEP 517 backend")
+            raise RuntimeError(f"inherited {name} reached the PEP 517 backend")
 
     expected_pythonpath = os.environ.get("EXPECTED_PYTHONPATH")
     if expected_pythonpath is None:
--- a/uv/private/pep517_whl/build_helper.py
+++ b/uv/private/pep517_whl/build_helper.py
@@ -100,6 +100,20 @@ def _override_tool(env, key, wrapper):
 
 def _compiler_env(tmpdir):
     env = dict(os.environ)
+    # The helper's launcher exports RUNFILES_DIR, RUNFILES_MANIFEST_FILE, and
+    # JAVA_RUNFILES:
+    # https://github.com/hermeticbuild/hermetic-launcher/blob/381814d0818af0573263323dc0dd0e4e208fc3fa/README.md#runfiles-discovery
+    # Bazel adds RUNFILES_MANIFEST_ONLY when runfiles trees are disabled:
+    # https://github.com/bazelbuild/bazel/blob/9.1.1/src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java#L192-L201
+    # Nested Bazel executables check that inherited state before adjacent
+    # runfiles, so remove the parent's identity before package code runs.
+    for key in (
+        "JAVA_RUNFILES",
+        "RUNFILES_DIR",
+        "RUNFILES_MANIFEST_FILE",
+        "RUNFILES_MANIFEST_ONLY",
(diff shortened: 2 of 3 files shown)


tamird:
Extracted from #1152. Its cowsay source-build fixture invoked a declared
py_binary, which inherited the PEP 517 helper runfiles tree and failed
to execve its interpreter.

tamird:
The `RUNFILES_MANIFEST_ONLY` finding is valid. 29df9746 now removes
it, and the hostile action-environment fixture sets it so the real
helper -> `python -m build` -> backend path proves the isolation.

The other two findings were against the pre-rebase #1148 commits and do
not apply to this patch:

- Bazel explicitly forces `ctx.attr` to a list for outgoing transitions,
  including 1:1 transitions: [Bazel documentation](https://bazel.build/extending/config#accessing-attributes-with-transitions).
  `ctx.attr.target[0]` is intentional, and the merged tests pass.
- `@pypi//build` belongs only to the private, `testonly`
  `__build_helper`. The e2e workspace consumes rules_py without its dev
  `pypi` hub and passes generated sdist builds; loading `rule.bzl` does
  not analyze that unrelated fixture target.

tamird:
@codex review

uv/private/pep517_whl/tests/python_env_backend/backend.py
```diff
@@ -13,9 +13,16 @@
 
 
 def _check_environment() -> None:
-    for name in ("PYTHONHOME", "PYTHONPLATLIBDIR"):
+    for name in (
+        "JAVA_RUNFILES",
+        "PYTHONHOME",
+        "PYTHONPLATLIBDIR",
+        "RUNFILES_DIR",
```
jbedard, line 23: Why `RUNFILES_*`?

uv/private/pep517_whl/tests/python_env_backend/backend.py
```diff
@@ -13,9 +13,16 @@
 
 
 def _check_environment() -> None:
-    for name in ("PYTHONHOME", "PYTHONPLATLIBDIR"):
+    for name in (
+        "JAVA_RUNFILES",
```
jbedard, line 20: Why JAVA but not CC?

uv/private/pep517_whl/tests/python_env_backend/backend.py
tamird: The helper is itself a Bazel `py_binary`, so hermetic-launcher exports `RUNFILES_DIR`, `RUNFILES_MANIFEST_FILE`, and `JAVA_RUNFILES` for the helper's runfiles tree. The helper then starts package-controlled backend code, which may invoke a different Bazel-built tool with adjacent runfiles. If the parent's runfiles identity survives, that child launcher consults the wrong tree before its adjacent runfiles and fails lookup. We therefore scrub only the subprocess environment after the helper has started. `RUNFILES_MANIFEST_ONLY` covers Bazel's no-runfiles-tree mode; the production comments cite both sources.

uv/private/pep517_whl/tests/python_env_backend/backend.py
tamird: `JAVA_RUNFILES` is Bazel's unfortunately named legacy runfiles locator; it is not `JAVA`, `JAVA_HOME`, or another Java toolchain variable. `CC`/`CXX` and the Java tool variables are intentionally preserved because native backends consume them. Only the parent launcher's runfiles identity is removed. I added that distinction beside this assertion.

uv/private/pep517_whl/tests/python_env_backend/backend.py
jbedard: What if it invokes a tool that was not bazel-built but depends on these exports?

tamird: It would fail if those variables were its only file locator, and that is intentional here. By the time `build_helper.py` starts, hermetic-launcher has replaced the three locators with the helper's own runfiles identity; they are not generic backend tool configuration. Forwarding that identity is what prevents an independently materialized Bazel tool from discovering its adjacent tree, because inherited locators are checked first ([launcher discovery](https://github.com/hermeticbuild/hermetic-launcher/blob/381814d0818af0573263323dc0dd0e4e208fc3fa/README.md#L147-L154)). `RUNFILES_MANIFEST_ONLY` is the mode bit for the same identity and must be cleared with the locators.

A non-Bazel tool that genuinely needs files should receive declared paths or a tool-specific wrapper through the build-tool interface. PEP 517 guarantees build requirements and their commands through the build environment and `PATH`; it does not require inheriting the frontend's Bazel runfiles namespace ([PEP 517](https://github.com/python/peps/blob/c7fb67920695b1d915fbaf9cfec4619002d400ac/peps/pep-0517.rst#L440-L478)). Conditionally preserving the helper's globals for the whole backend would reintroduce the wrong-tree bug.

jbedard (approved):

PR #1169, opened 2026-06-24
fix(py): retain unknown wheel roots

tamird:
Source-built wheels need to expose console scripts during analysis,
before the wheel exists to reveal its immediate layout. That valid state
exposed a pre-existing bug in `py_unpacked_wheel`: it may advertise
console scripts without declaring `top_levels`, but venv assembly treats
the empty layout as fully covered. The wheel can disappear from
`sys.path`; if its install tree keeps it reachable, a plain path still
skips root `.pth` processing.

Treat a nonempty `top_levels` list as the existing complete-layout
signal. Preserve unknown layouts through whole-wheel fallback and call
`site.addsitedir` on a venv-local wheel path. Known layouts retain plain
path handling because their root entries have already been projected or
suppressed by collision resolution.

### DIFF
--- a/e2e/cases/pth-install-tree-fallback/BUILD.bazel
+++ b/e2e/cases/pth-install-tree-fallback/BUILD.bazel
@@ -65,6 +65,7 @@ py_unpacked_wheel(
     # site-packages exactly as uv's whl_install would.
     top_levels = [
         "apkg",
+        "itfa-1.0.dist-info",
         "shared",
         "marker.pth",
     ],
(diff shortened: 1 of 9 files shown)


tamird:
Extracted from #1153. That patch needs source-built wheels to advertise console scripts while leaving their immediate layout unknown; current venv assembly drops the wheel or skips its root `.pth` files in that state.

tamird:
Redundant after https://github.com/aspect-build/rules_py/commit/518e1164e3407a977477479d9449995b1331c56b and https://github.com/aspect-build/rules_py/commit/873b578f1d26dec150afba4335974ec1ece8af4f. Source-built wheels now advertise scripts without claiming topology, and unknown layouts remain on the whole-wheel `site.addsitedir` fallback after the runfiles indirection is removed. Rebasing this patch over those changes produces an empty diff.

PR #1170, opened 2026-06-24
fix(py): retain metadata-free wheel trees

tamird:
This depends on
https://github.com/tamird/rules_py/commit/b5252ca9379f72d66642d0b28e85e0f633c2d852.
The metadata-free wheel change is
https://github.com/tamird/rules_py/commit/c9512079ec266f62c65fef26b1d308765a3bbcf6.
The prerequisite is independently submitted and will be removed when it
reaches `main`.

Repository analysis cannot inspect source-built wheels, so
`whl_install` currently omits `PyWheelsInfo` for them. `py_image_layer`
uses that provider to classify wheel package leaves and read
`install_tree`; its generic `PyInfo` path deliberately skips wheel
files. Metadata-free source wheels can therefore disappear from
dependency layers.

Make installed-tree identity unconditional for both wheel producers
while keeping layout and console-script metadata optional. Venv
assembly localizes only records with actionable metadata, preserving
the existing runfiles fallback for metadata-free wheels. Image
layering can now retain every wheel tree as a package leaf.

### DIFF
--- a/e2e/cases/pth-install-tree-fallback/BUILD.bazel
+++ b/e2e/cases/pth-install-tree-fallback/BUILD.bazel
@@ -65,6 +65,7 @@ py_unpacked_wheel(
     # site-packages exactly as uv's whl_install would.
     top_levels = [
         "apkg",
+        "itfa-1.0.dist-info",
         "shared",
         "marker.pth",
     ],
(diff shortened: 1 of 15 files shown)


tamird:
Extracted from #1153. Auditing source-built wheel metadata exposed that metadata-free wheel targets omit `PyWheelsInfo`, so `py_image_layer` can drop their install trees from dependency layers.

tamird:
Obsolete after https://github.com/aspect-build/rules_py/commit/518e1164e3407a977477479d9449995b1331c56b made `PyWheelsInfo` unconditional for source-built and metadata-free wheels. Rebasing leaves only an analysis test that restates the producer fields without exercising the original image-layer loss, so there is no remaining behavior patch to land.

PR #1171, opened 2026-06-24
fix(uv): classify custom .data categories

tamird:
Wheel unpacking preserves unknown `.data/<category>/...` entries
under site-packages with the category prefix. Repository metadata drops
every category except `purelib` and `platlib`, so it can describe an
installed layout that omits files the unpacker retained.

Route only the standard `data`, `headers`, and `scripts` categories
away from site-packages. Preserve unknown categories exactly as the
unpacker does so RECORD-derived metadata matches the installed tree.

### DIFF
--- a/uv/private/whl_install/repository.bzl
+++ b/uv/private/whl_install/repository.bzl
@@ -75,9 +75,17 @@ def site_packages_segments(path, data_directory):
     segments = path.split("/")
     if segments[0] != data_directory:
         return segments
-    if len(segments) < 3 or segments[1] not in ("purelib", "platlib"):
+    if len(segments) < 3:
         return []
-    return segments[2:]
+    category = segments[1]
+    if category in ("purelib", "platlib"):
+        return segments[2:]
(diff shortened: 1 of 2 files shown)


tamird:
Extracted from #1142. Its installed-wheel topology metadata becomes
incomplete when unpacking preserves an unknown `.data` category that
repository classification drops.

jbedard (approved):

jbedard:
Could this be reproduced with any real packages in an e2e case?

tamird:
There isn't a real downstream reproducer. The concrete failure was
`tritonclient`'s `.data/purelib`, which is already handled on
`main`.

A scan of 535 locally available wheels from the downstream lock, plus
744 unique wheels across the local Bazel caches, found only `data`,
`headers`, and `scripts` beneath `.data`; none used an unknown
category.

This is therefore defensive support for a hypothetical nonstandard
wheel, not a required downstream fix. The patch is coming out of the
downstream integration, and this PR is closing rather than adding a
synthetic e2e case.

PR #1173, opened 2026-06-24
fix(py): union colliding package directories

tamird:
The atomic top-level venv layout selects one wheel when ordinary package
directories collide. Apache Airflow splits its regular `airflow` package
across core and provider distributions, so unique provider modules
disappear.

In permissive collision modes, union every ordinary directory claimant
at the final `site-packages` path. Sort inputs by runfiles path and
validate all claims before materializing the output. Disjoint entries
and identical files with matching executable bits are compatible.
File-valued top-levels, differing contents or executable bits, and
file/directory conflicts fail instead of selecting code by depset
traversal order.

Apply the same validation to regular packages spanning namespace wheels.
This restores flat-install directory behavior without analysis-time
entry-type metadata or an arbitrary permissive winner.


tamird:
Extracted from #1155.

tamird:
Redone directly against `main` as [#1194](https://github.com/aspect-build/rules_py/pull/1194).

PR #1174, opened 2026-06-24
feat(uv): monitor wheel build memory

tamird:
Large source-built wheels can exhaust memory before their action logs show
enough evidence to size a scheduler reservation.

Add an explicit `monitor_memory` package override that samples approximate
Linux procfs RSS for the PEP 517 frontend and its descendants. Emit the
first sample, 256 MiB high-water crossings, and a final peak. Preserve
earlier samples across procfs races and identify a direct SIGKILL as a
possible OOM.

Keep diagnostics independent from `resource_set`. The monitor inherits the
Bazel process group and never installs signal handlers or kills child
processes, so Bazel retains cancellation ownership. Only opted-in generated
build tools carry the monitor dependency.

Apply monitoring only to generated sdist actions. Selecting a compatible
prebuilt wheel produces no report; prebuilt-only packages and complete
custom BUILD replacements reject the override.

### DIFF
--- a/docs/uv-patching.md
+++ b/docs/uv-patching.md
@@ -126,6 +126,40 @@ package that resolves to a prebuilt wheel (no source build) fails the build
 rather than silently dropping the reservation — force a source build with
 `[tool.uv] no-binary-package` if you need the reservation to apply.
 
+### Monitoring wheel build memory
+
+Set `monitor_memory` to report the memory observed while building a wheel from
+an sdist:
+
+```starlark
+uv.override_package(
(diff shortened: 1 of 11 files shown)


tamird:
Extracted from #1140. That patch coupled memory diagnostics to resource
reservations and took ownership of the build process group; this replacement
makes monitoring explicit and leaves cancellation to Bazel.

jbedard (approved):

xangcastle (approved):

PR #1175, opened 2026-06-24
fix(py): select one PBS archive per mode

tamird:
PBS archive filenames encode physical build variants, but runtime
selection exposes only normal and free-threaded modes. The extension
currently registers normal and stripped archives with identical
constraints, and optimized and debug free-threaded archives with
identical constraints. Bazel then chooses by registration order: GNU
free-threaded targets use the debug archive, musl has only a debug
archive, and Windows has none.

Represent those two logical modes directly and map each supported
platform to one physical archive suffix. Normal mode uses
`install_only`; free-threaded mode uses the optimized archive that PBS
publishes for the platform. Version the cached manifest index and
replace documentation for the nonexistent `build_config` toolchain
attribute with the actual free-threaded build setting.

The generated `*_install_only_stripped` and
`*_freethreaded_debug` labels are removed, and
`*_freethreaded_pgo_lto` becomes `*_freethreaded`. Keeping aliases
would preserve the ambiguous registration surface.

### DIFF
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -176,25 +176,22 @@ versions:
 
 The target attribute takes precedence over the build-wide flag.
 
-## Build configurations
+## Runtime modes
 
-PBS provides several build configurations. The default is `install_only`, which
-is PGO+LTO optimized on platforms that support it. You can select a different
-configuration per toolchain:
+The extension registers one normal `install_only` PBS archive for each
+available Python version and platform. It also registers an optimized
+free-threaded archive where PBS publishes one, currently for Python 3.13 and
+newer. Normal mode is selected by default. Select free-threaded mode with the
+build setting:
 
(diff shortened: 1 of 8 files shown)


tamird:
Extracted from #1151. Aligning PBS runtimes and build tools exposed
normal/stripped and optimized/debug runtime registrations with identical
constraints; Bazel selected a physical build by registration order.

tamird:
Rebased onto 9986235c.

jbedard:
please rebase

tamird:
This already landed on `main` in
https://github.com/aspect-build/rules_py/commit/2ab36be017ca1f221400d437948fd54de18acf64.
That merge included the archive-selection prerequisite, so rebasing this
would recreate an existing change. Closing the duplicate.

PR #1177, opened 2026-06-24
fix(py): avoid ambiguous Linux exec tools

tamird:
PBS publishes GNU and musl Linux exec-tools registrations with identical
execution-platform constraints because libc is a target setting. The
default `@python_interpreters//:all` target-pattern registration
currently chooses GNU only because its label sorts first. Binding each
registration to its own PBS runtime would therefore make Linux libc
selection depend on order.

Emit PBS exec-tools registrations only for GNU Linux while retaining
both GNU and musl target runtimes. Supported Linux execution hosts use
glibc, so the generated graph has one Linux exec candidate without
restricting cross-compiled musl targets.

The generated musl exec-toolchain labels disappear rather than becoming
aliases; aliases would preserve the ambiguous registrations. Their
physical repositories still expose `exec_tools_toolchain` targets.

### DIFF
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -215,6 +215,11 @@ The following platforms are registered by default:
 Not all Python versions are available on all platforms. Unavailable combinations
 are silently skipped during toolchain resolution.
 
+All listed PBS platforms emit Python runtime registrations. PBS exec-tools
+registrations are emitted only for supported execution platforms: macOS,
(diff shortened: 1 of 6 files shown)


tamird:
Extracted from #1151. Binding each PBS exec-tools target to its own
runtime exposed GNU and musl Linux registrations with identical
execution-platform constraints; Bazel selected GNU only by
target-pattern registration order.

jbedard (approved):

PR #1179, opened 2026-06-24
fix(py): parse PBS prerelease metadata

tamird:
The interpreter extension can provision prerelease PBS builds such as
3.15.0a6, but its generated runtime currently places `0a6` in
`interpreter_version_info["micro"]`. rules_python converts that field
to an integer, so selecting the runtime fails during analysis.

Use one PBS version model for archive ordering, prerelease filtering,
and the five `sys.version_info` fields. Final-release BUILD output
remains unchanged; prerelease runtimes also declare `releaselevel` and
`serial`.


tamird:
Extracted from openai/openai#1009324. The upgrade provisions the PBS
Python 3.15 prerelease while validating native-extension toolchains; the
generated py_runtime fails analysis before those toolchains can be tested
because its integer micro field contains 0a6.

tamird:
Superseded by https://github.com/aspect-build/rules_py/commit/86d31bf1fad758c34efe77d21c8bc9445fa5dcda. That change emits complete prerelease runtime metadata from the pinned PBS release and covers the actual Python 3.15a6 form, so this parser refactor no longer fixes an unfixed behavior.

PR #1180, opened 2026-06-24
test(py): repair toolchain settings e2e

tamird:
The toolchain-settings e2e still uses a `3.11.14` tag to exercise
duplicate declarations. Toolchain tags now intentionally require
`major.minor`, so every fork PR fails this standalone test before it
reaches the settings invariant.

Use duplicate `3.11` tags instead. Keep the conflict case and the
order-insensitive duplicate case intact, and describe them as duplicate
declarations rather than normalization.


tamird:
Extracted from openai/openai#1009324. Rebasing its upstream rules_py
patch stack onto 910b228c exposed that the merged major.minor validation
and the toolchain-settings e2e disagreed before any patch-specific tests
ran.

jbedard (approved):

PR #1166, opened 2026-06-24
fix(uv): reject ignored build overrides

tamird:
uv.override_package accepted declarations whose lock, version, package
kind, or selected build path could not consume their values. Those
settings were then silently ignored.

Validate each override against a project lock and matching package
version. Reject modification-only overrides for virtual and editable
workspace packages, and require patch strip values to accompany patch
lists.

Classify source-build attributes by the selected path. Wheel-only
records reject every build attribute; complete custom BUILD content
owns pre-build patch context but rejects generated-action settings;
generated pure builds reject native environment and toolchain settings.

PR #1182, opened 2026-06-24
fix(uv): preserve explicit compiler env

tamird:
`pep517_native_whl` expands its `env` attribute and then replaces `CC`
and `CXX` with the resolved C++ toolchain compiler. This silently
discards matching `uv.override_package(env = ...)` values, including
custom wrappers and flags.

Establish the resolved compiler as a hermetic default before applying
the user environment. Matching keys now replace defaults, while omitted
keys still use the resolved toolchain and its declared inputs. Clarify
that `env` is a key-wise override rather than additions-only.

### DIFF
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -782,19 +782,18 @@ _override_package_tag = tag_class(
                   "bucket.",
         ),
 
-        # Per-package toolchain plumbing for native sdist builds. Both
-        # attributes AUGMENT the defaults baked into sdist_build's
-        # generated `pep517_native_whl(...)` call (the CC toolchain +
-        # CC/CXX/AR/LD/STRIP env) — they don't replace them. Use these
-        # to layer extra toolchains (Java runtime, Rust, …) and extra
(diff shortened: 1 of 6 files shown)


tamird:
Extracted from #1166. Auditing whether source-build override fields were consumed exposed that `pep517_native_whl` expands `env` and then overwrites explicit `CC`/`CXX`, contradicting the extension's key-wise merge.

tamird:
Closing this rather than expanding the isolated ordering fix. CI showed that generated `$(CC)` resolves to a toolchain launcher absent from the action inputs. A follow-up audit found that `AR`, `LD`, and `STRIP` share the same defect, and the current rule reads the default C++ toolchain while the build action runs in the `target` execution group.

Repairing those together would recreate the compiler-projection machinery that https://github.com/aspect-build/rules_py/commit/48fe82afa60cab52785ee84d8fd724267ede929e deliberately removes in favor of declared source-build tools. The integration that exposed the inconsistency has no `CC`/`CXX` override consumer, so this patch is not independently required once that change is used.

PR #1183, opened 2026-06-24
fix(py): bind selected PBS exec runtime

tamird:
PBS exec-tools registrations select a runtime by execution OS and CPU,
but rules_python's default `exec_interpreter` resolves Python again
under `cfg = "exec"`. A registration can therefore advertise one PBS
archive while invoking a different interpreter.

Bind each physical PBS runtime pair to an executable adapter that
implements rules_python's `exec_interpreter` contract. Constrain each
registration by canonical Python version, free-threaded mode, root
configuration settings, and root target compatibility. Keep target-only
libc settings out of execution selection and apply PBS OS and CPU
constraints to execution platforms.

The end-to-end fixture builds for Linux arm64 while executing on Linux
x86_64, then checks the selected interpreter's file, Python version, and
GIL mode. Its negative cases establish that incompatible root target or
execution constraints cannot select the PBS runtime.

### DIFF
--- a/.github/workflows/ci-workflows.yaml
+++ b/.github/workflows/ci-workflows.yaml
@@ -60,6 +60,22 @@ jobs:
             - name: Test
               working-directory: ${{ matrix.workspace.path }}
               run: aspect test --task-key=test-${{ matrix.workspace.slug }}-${{ matrix.bazel.id }} ${{ matrix.bazel.flags }} -- //...
+            # This fixture's root toolchain declarations live in its own module,
+            # not the parent e2e module used by the matrix entry above.
+            - name: Interpreter toolchain settings
+              if: matrix.workspace.slug == 'e2e'
+              working-directory: e2e/cases/interpreter-toolchain-settings
+              run: |
+                aspect test \
+                  --task-key=test-interpreter-toolchain-settings-${{ matrix.bazel.id }} \
+                  ${{ matrix.bazel.flags }} \
+                  --bazel-flag=--lockfile_mode=off \
+                  -- \
+                  //:resolved_311_test \
+                  //:resolved_312_test \
+                  //:resolved_313_freethreaded_test \
+                  //:resolved_313_glibc_test \
+                  //:resolved_313_musl_test
             # Standalone scripts that drive bazel directly (e.g. asserting a
             # build *fails*), so they can't be an `sh_test` under `//...` above.
             - name: e2e standalone test.sh scripts
--- a/e2e/cases/interpreter-toolchain-settings/BUILD.bazel
+++ b/e2e/cases/interpreter-toolchain-settings/BUILD.bazel
@@ -1,30 +1,110 @@
-load(":toolchain_test.bzl", "interpreter_toolchain_check")
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+load(":toolchain_test.bzl", "exec_toolchain_check", "interpreter_toolchain_check", "toolchain_resolution_test")
+
+string_flag(
+    name = "interpreter_setting",
+    build_setting_default = "",
+)
+
+string_flag(
+    name = "interpreter_setting_secondary",
+    build_setting_default = "",
+)
 
 config_setting(
     name = "config_311",
-    define_values = {"interpreter_setting": "311"},
+    flag_values = {":interpreter_setting": "311"},
 )
 
 config_setting(
     name = "config_311_secondary",
-    define_values = {"interpreter_setting_secondary": "311"},
+    flag_values = {":interpreter_setting_secondary": "311"},
 )
 
 config_setting(
     name = "config_312",
-    define_values = {"interpreter_setting": "312"},
+    flag_values = {":interpreter_setting": "312"},
+)
+
+constraint_setting(name = "python_target")
+
+constraint_setting(name = "python_exec")
+
+constraint_value(
+    name = "python_exec_supported",
+    constraint_setting = ":python_exec",
+    visibility = ["//visibility:public"],
+)
+
+constraint_value(
+    name = "python_target_supported",
+    constraint_setting = ":python_target",
+    visibility = ["//visibility:public"],
+)
+
+constraint_value(
+    name = "python_target_unsupported",
+    constraint_setting = ":python_target",
+    visibility = ["//visibility:public"],
 )
 
 platform(
     name = "linux_x86_64",
+    constraint_values = [
+        ":python_target_supported",
+        "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
+    ],
+)
+
+platform(
+    name = "linux_x86_64_exec",
     constraint_values = [
         "@platforms//cpu:x86_64",
         "@platforms//os:linux",
     ],
 )
 
+platform(
+    name = "linux_x86_64_exec_supported",
+    constraint_values = [
+        ":python_exec_supported",
+        "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
+    ],
+)
+
+platform(
+    name = "linux_aarch64",
+    constraint_values = [
+        ":python_target_supported",
+        "@platforms//cpu:aarch64",
+        "@platforms//os:linux",
+    ],
+)
+
+platform(
+    name = "linux_x86_64_unsupported",
+    constraint_values = [
+        ":python_target_unsupported",
+        "@platforms//cpu:x86_64",
+        "@platforms//os:linux",
+    ],
+)
+
+# This package is also visible from the parent e2e workspace, whose module does
+# not declare these root toolchain settings. CI names the tests explicitly from
+# this fixture's module; `manual` keeps parent-workspace wildcards from selecting
+# their checks or wrappers under the wrong module graph.
 interpreter_toolchain_check(
     name = "resolved_311",
+    expected_interpreter = "@python_3_11_x86_64_unknown_linux_gnu//:bin/python3",
+    python_version = "3.11",
+    tags = ["manual"],
+)
+
+exec_toolchain_check(
+    name = "resolved_311_exec",
     expected_interpreter = "@python_3_11_x86_64_unknown_linux_gnu//:bin/python3",
     python_version = "3.11",
     tags = ["manual"],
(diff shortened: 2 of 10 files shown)


tamird:
Extracted from #1151. Splitting its runtime/toolchain cohort patch exposed that `py_exec_tools_toolchain` re-resolves its interpreter under `cfg = "exec"`, so constraining registrations alone does not bind the PBS runtime they advertise.

py/private/interpreter/extension.bzl
```diff
@@ -547,59 +501,10 @@ Only honored from the root module.
...
-""",
-)
-
 python_interpreters = module_extension(
     implementation = _python_interpreters_impl,
     tag_classes = {
         "configure": _configure_tag,
-        "local": _local_tag,
         "toolchain": _toolchain_tag,
     },
```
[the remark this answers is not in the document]
tamird: Obsolete after the rebase. [25a0707b](https://github.com/aspect-build/rules_py/commit/25a0707b) removed the broken `local` tag upstream, so this branch no longer changes that API.

py/private/interpreter/BUILD.bazel
```diff
@@ -24,9 +22,12 @@ bzl_library(
...
 )
 
-string_flag(
+# Preserve the released rules_py label without maintaining separate state.
+# Bazel follows alias chains when parsing command-line build settings:
+# https://github.com/bazelbuild/bazel/blob/c8217fdd2f20e4a061122c0af0417380d09e9480/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java#L304-L376
+alias(
     name = "python_version",
-    build_setting_default = "",
+    actual = "@rules_python//python/config_settings:python_version",
```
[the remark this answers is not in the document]
tamird: Obsolete after the rebase. The branch no longer replaces the released Python build-setting label with an alias; its current diff only binds each generated exec-tools toolchain to its selected PBS runtime.

jbedard:
needs rebase

jbedard:
List PRs that are prefactors to this

tamird:
Rebuilt as 4fadd48e, directly on current `main`. It has no precursor PRs.

e2e/cases/interpreter-toolchain-settings/MODULE.bazel
```diff
@@ -2,6 +2,7 @@ module(name = "interpreter_toolchain_settings")
 
 bazel_dep(name = "aspect_rules_py")
 bazel_dep(name = "platforms", version = "1.1.0")
+bazel_dep(name = "rules_python", version = "1.9.0")
```
jbedard, line 9: Remove the need for direct deps on rules_python if possible

e2e/cases/interpreter-toolchain-settings/test.sh
```diff
@@ -17,23 +22,47 @@ check_toolchain() {
...
+        -- "//:resolved_${setting}" "//:resolved_${setting}_exec" \
         || fail "Python ${version} did not resolve with its root config_setting"
 }
 
+check_exec_toolchain() {
+    local target="$1"
+    local libc="$2"
+    shift 2
+
+    "$BAZEL" build \
```
jbedard: Unless checking for errors do not use `test.sh` for testing and use bazel tests within BUILDs

e2e/cases/interpreter-toolchain-settings/MODULE.bazel
tamird: Kept, with a comment explaining the boundary. The fixture directly requests `@rules_python//python:exec_tools_toolchain_type` and reads its `exec_tools` provider. `aspect_rules_py` exposes neither an alias nor a public helper for that contract, so removing the direct dependency would require adding API solely for this fixture.

e2e/cases/interpreter-toolchain-settings/test.sh
tamird: Done. Successful cases are now five discovered `build_test` targets. A host-configured test depends on each probe through an outgoing transition, so the synthetic Linux target platform does not break Bazel 9 test-toolchain resolution; the probe outputs also force all five `ExecInterpreterSmoke` actions into the build. CI runs the nested module with Bazel 8 and 9. `test.sh` now contains only expected analysis and module-evaluation failures.

.github/workflows/ci-workflows.yaml
```diff
@@ -60,6 +60,10 @@ jobs:
             - name: Test
               working-directory: ${{ matrix.workspace.path }}
               run: aspect test --task-key=test-${{ matrix.workspace.slug }}-${{ matrix.bazel.id }} ${{ matrix.bazel.flags }} -- //...
+            - name: Interpreter toolchain settings
```
jbedard, line 65: Why is this needed?

e2e/cases/interpreter-toolchain-settings/BUILD.bazel
```diff
@@ -30,9 +99,89 @@ interpreter_toolchain_check(
     tags = ["manual"],
 )
 
+exec_toolchain_check(
+    name = "resolved_311_exec",
+    expected_interpreter = "@python_3_11_x86_64_unknown_linux_gnu//:bin/python3",
+    python_version = "3.11",
+    tags = ["manual"],
```
jbedard, line 110: Why are these tagged `manual`?

e2e/cases/interpreter-toolchain-settings/toolchain_test.bzl
```diff
@@ -1,6 +1,37 @@
-"""Analysis check for the selected Python runtime toolchain."""
+"""Checks selected Python runtime and exec-tools toolchains."""
+
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
 
 _RUNTIME_TOOLCHAIN = "@bazel_tools//tools/python:toolchain_type"
+_EXEC_TOOLS_TOOLCHAIN = "@rules_python//python:exec_tools_toolchain_type"
+
+_EXEC_INTERPRETER_CHECK = """\
```
jbedard: why write python code in a string instead of .py file?

.github/workflows/ci-workflows.yaml
tamird: This package is a standalone Bzlmod root. The parent e2e module does not declare these root config settings or toolchain constraints, so its normal `//...` run cannot exercise the fixture correctly. This step names the five BUILD tests from the fixture module.

e2e/cases/interpreter-toolchain-settings/BUILD.bazel
tamird: The package is also traversable from the parent e2e workspace, whose module graph lacks the fixture root toolchain declarations. `manual` keeps parent-workspace wildcards from selecting these probes and wrappers under the wrong module graph; the nested workflow explicitly names all five tests.

e2e/cases/interpreter-toolchain-settings/toolchain_test.bzl
tamird: Moved it to `exec_interpreter_check.py` and declared the file as an action input.

.github/workflows/ci-workflows.yaml
jbedard: @xangcastle we should do this better - some tests must be fully isolated with their own MODULE files and maybe those should be in `e2e/{case}` instead of `e2e/cases/{case}`?

Today I always assume `e2e/cases/{case}/include.MODULE` and `test //...` will cover it. But sometimes a test needs its own isolated root MODULE like this `e2e/cases/interpreter-toolchain-settings` or `examples/protobuf` etc...

e2e/cases/interpreter-toolchain-settings/test.sh
```diff
@@ -1,48 +1,83 @@
...
+# Successful resolution is covered by BUILD tests. This script keeps cases
+# whose expected result is an analysis or module-evaluation failure.
 
 set -uo pipefail
 
 cd "$(dirname "$0")" || exit 1
 
 BAZEL="${BAZEL:-bazel}"
+BUILD_ARGS=()
+if [[ "$(uname -s)" != "Linux" || "$(uname -m)" != "x86_64" ]] ||
```
jbedard: Why was this added? Is this required?

tamird: It was needed while this script still executed the positive Linux cases. Those are BUILD tests now, so the conditional is obsolete. I will remove it and pass `--nobuild` unconditionally to the remaining expected-analysis-failure commands.

jbedard:
reduce the diff

e2e/cases/interpreter-toolchain-settings/toolchain_test.bzl
```diff
@@ -1,6 +1,9 @@
-"""Analysis check for the selected Python runtime toolchain."""
+"""Checks selected Python runtime and exec-tools toolchains."""
+
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
 
 _RUNTIME_TOOLCHAIN = "@bazel_tools//tools/python:toolchain_type"
+_EXEC_TOOLS_TOOLCHAIN = "@rules_python//python:exec_tools_toolchain_type"
```
jbedard, line 8: Why is exec from `rules_python` while runtime is from `bazel_tools`?

py/private/interpreter/repository.bzl
```diff
@@ -419,40 +444,43 @@ config_setting(
...
-            target_settings = target_settings,
+            runtime_target_settings = runtime_target_settings,
+            runtime_target_compatible_with = runtime_target_compatible_with,
         ))
 
         if info["register_exec_tools"]:
-            content.append("""# Exec tools toolchain: selected by exec platform (not target platform) so
-# that build actions using the interpreter (e.g. compileall) get a runnable
-# binary on the build host regardless of the target platform being built for.
+            content.append("""# Exec tools run on the selected execution platform for this target's
```
jbedard: While editing the generated code... remove the comment from the generated code and put it into the `repository.bzl` where the code is being generated.

jbedard:
Which e2e tests failed before this change?

e2e/cases/interpreter-toolchain-settings/test.sh
jbedard: Done yet?

e2e/cases/interpreter-toolchain-settings/toolchain_test.bzl
tamird: The runtime is Bazel's standard Python toolchain contract; exec tools are a separate public contract supplied by rules_python. I added that ownership explanation above the two constants.

py/private/interpreter/repository.bzl
tamird: Done. The explanation now lives beside the generator control flow; generated BUILD files contain only the toolchain declaration.

e2e/cases/interpreter-toolchain-settings/test.sh
tamird: Done.

e2e/cases/interpreter-toolchain-settings/BUILD.bazel
jbedard: Why is this traversable from the parent? Shouldn't `e2e/cases/interpreter-toolchain-settings` be in the parent `.bazelignore`?

e2e/cases/interpreter-toolchain-settings/toolchain_test.bzl
```diff
@@ -31,3 +36,124 @@ interpreter_toolchain_check = rule(
...
+            output.path,
+        ],
+        executable = exec_tools.exec_interpreter[DefaultInfo].files_to_run,
+        inputs = [ctx.file._exec_interpreter_check, ctx.file.expected_interpreter],
+        mnemonic = "ExecInterpreterSmoke",
+        outputs = [output],
+    )
+    return [DefaultInfo(files = depset([output]))]
+
+exec_toolchain_check = rule(
```
jbedard, line 70: Can this test coverage be accomplished without a custom rule?

tamird:
No end-to-end test failed before this change, and a consumer with one registered runtime resolves the same interpreter in both configurations. Closing until there is a concrete reproducer with independently selected target and execution runtimes.

Issue #959, opened 2026-04-30
[Bug]: incorrect docs about build_config


mboulton-fathom:
### What happened?

The documentation says that using the new 'unstable' toolchains the `build_config` attribute can be set when registering a python toolchain: https://github.com/aspect-build/rules_py/blob/b2feee5b81c51c3a4ed24ca2df3cb68f57a78f57/docs/interpreter.md#L186-L204

But this doesn't work and looking at py/private/interpreter/extension.bzl it doesn't actually appear to be implemented?

```
ERROR: Analysis of target '//my:rule' failed; build aborted: Traceback (most recent call last):
	File "/home/michael/code/myproject/MODULE.bazel", line 262, column 23, in <toplevel>
		interpreters.toolchain(
Error: in 'toolchain' tag, unknown attribute 'build_config' provided
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 9.0.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

```
bazel_dep(name = "rules_python", version = "2.0.0")
bazel_dep(name = "aspect_rules_py", version = "1.11.2")
```

Language(s) and/or frameworks involved:

### How to reproduce

```shell
Try passing build_config to `interpreters.toolchain`
```

### Any other information?

Previously I was doing something like this to select the 'stripped' version of Python:

```
python.single_version_override(
    python_version = "3.12.11",
    sha256 = {
        "x86_64-unknown-linux-gnu": "8592d071a3636e1af1bfcb473b31b1f058a092acd35a0a360238d6779ba48f28",
    },
    strip_prefix = "python",
    urls = ["https://github.com/astral-sh/python-build-standalone/releases/download/20250814/cpython-3.12.11+20250814-{platform}-install_only_stripped.tar.gz"],
)
```

This works fine

chpock:
The documentation mentioned in this ticket no longer exists.

So it looks like this issue is no longer relevant.

I created a new FR with the same intent: https://github.com/aspect-build/rules_py/issues/1218

mboulton-fathom:
I think that the PR https://github.com/aspect-build/rules_py/pull/827 that added the code in my initial comment was either unfinished or just fully AI-generated because it doesn't appear to have been reviewed or had any tests for the 'stripped' path, as it's been removed I'll close this now

PR #1187, opened 2026-06-25
fix(py): register PBS C toolchains

### LINKED ISSUE #1095: [Bug]: v2 python_interpreters extension doesn't register a py_cc (python/cc:toolchain_type) toolchain, breaking native extension builds

tamird:
The `python_interpreters` extension registers PBS runtimes but not the
matching `rules_python` C toolchains. Native-extension targets can
therefore fail to resolve `current_py_cc_headers` or select headers from
a separately registered interpreter.

Register a C toolchain beside every PBS runtime and give both
registrations identical version, free-threading, platform, libc, and
root target predicates. POSIX toolchains expose headers only because
extension modules resolve Python symbols from the loading interpreter.
Windows toolchains additionally expose versioned import libraries and
`python3.lib` for the regular stable ABI.

Regular runtimes expose stable-ABI headers. Free-threaded CPython 3.13
rejects `Py_LIMITED_API`, and `rules_python` does not model Python 3.15's
`abi3t` split, so free-threaded toolchains expose only the full ABI.

Windows target analysis verifies the selected headers and import-library
labels. The current CI matrix does not execute Windows binaries, so this
does not claim an end-to-end Windows extension build.

Fixes #1095.

The prerequisite archive-selection change is
https://github.com/tamird/rules_py/commit/dc43cdd1bd03ab04c57e3d92b52b4614ead26199.
GitHub cannot base this cross-repository pull request on a fork branch,
so it appears as the first commit until that change lands. The
C-toolchain change is
https://github.com/tamird/rules_py/commit/054fdb71acb62973ef44e0605bd295f59a3ad1f9.


chpock:
This PR completely removes stripped python distribution. Yes, it is broken as for now (https://github.com/aspect-build/rules_py/issues/959). But it is possible to use it by registering specific toolchains rather than `:all`.

xangcastle (approved):

PR #1188, opened 2026-06-25
fix(uv): validate bytecode runtime identity

tamird:
`whl_install` compiles wheel bytecode with an execution-platform
interpreter while the target runtime consumes the resulting tree. PBS
toolchains can resolve those runtimes independently, and Python version
metadata is not an exact bytecode identity.

When either runtime pair declares `pyc_magic_number`, require both sides
to declare the same value before enabling compilation. Preserve existing
behavior when neither runtime pair opts in, and bypass the check when
precompilation is disabled.

Read exec identity from `exec_interpreter` only when available, and
require its forwarded `py3_runtime` to match the RBE-safe `exec_runtime`
used by the action. This prevents validating one interpreter while
executing another.

The guard [f6af2403] uses the bytecode-identity producer [e59cd1cc] and
the selected-exec-runtime binding [45619404]; both follow archive
selection [c1cd9499]. GitHub cannot base this cross-repository pull
request on another fork branch, so those prerequisites remain in the
branch until they land upstream.

[f6af2403]:
  https://github.com/tamird/rules_py/commit/f6af2403
[e59cd1cc]:
  https://github.com/tamird/rules_py/commit/e59cd1cc
[45619404]:
  https://github.com/tamird/rules_py/commit/45619404
[c1cd9499]:
  https://github.com/tamird/rules_py/commit/c1cd9499

### DIFF
--- a/docs/interpreter.md
+++ b/docs/interpreter.md
@@ -176,25 +176,22 @@ versions:
 
 The target attribute takes precedence over the build-wide flag.
 
-## Build configurations
+## Runtime modes
 
-PBS provides several build configurations. The default is `install_only`, which
-is PGO+LTO optimized on platforms that support it. You can select a different
-configuration per toolchain:
+The extension registers one normal `install_only` PBS archive for each
+available Python version and platform. It also registers an optimized
+free-threaded archive where PBS publishes one, currently for Python 3.13 and
+newer. Normal mode is selected by default. Select free-threaded mode with the
+build setting:
 
-```starlark
-interpreters.toolchain(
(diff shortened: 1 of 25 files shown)


jbedard:
needs rebase

jbedard:
List PRs that are prefactors to this

tamird:
Closing this stacked version. The bytecode guard depends on the runtime metadata in #1178 and the selected-exec-runtime binding in #1183. It will be recreated directly on `main` after both land rather than asking reviewers to reason about a four-commit stack.

PR #1190, opened 2026-06-25
feat(uv): expose group dependency labels

tamird:
`group_deps()` suits configurable dependency attributes, but a macro
cannot iterate its `select()` result. Waiting until analysis also loses
the normalized package name after a hub alias resolves to a project SCC
target.

Expose `group_dep_labels(name)` from native `defs.bzl`. It returns sorted
`Label` values from the existing per-group membership table, so macros
can inspect package names during loading without publishing Bzlmod
canonical repository spellings.

The labels describe membership, not a pinned resolution. Package aliases
still resolve through the active `dep_group` and must be analyzed under
the matching group. Prefer `group_deps()` for ordinary rule attributes,
and keep `requirements.bzl` aligned with the `rules_python` surface.

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -153,6 +153,14 @@ py_binary(
 future options (such as PEP 508 package extras) can be added as arguments
 without breaking callers.
 
+Macros that must inspect dependency labels while loading a BUILD file can use
+`group_dep_labels(name)` instead. It returns the frozen membership list for one
+explicit group as sorted `Label` values and fails if the group is unknown.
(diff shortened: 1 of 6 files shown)


tamird:
Extracted from #1009324. Updating the downstream rules_py snapshot
removed the provisional `all_requirements_by_dep_group` symbol. Its
pytypes macro still needs canonical per-group labels during BUILD loading
to generate one mapping target per dependency group without analyzing
every package.

xangcastle (approved):

PR #1193, opened 2026-06-25
fix(uv): validate patched wheel topology

tamird:
Post-install patches run after repository analysis records a wheel's
layout. A downstream rules_py 2 upgrade patched a wheel that also
participates in package merges. Treating the pre-patch snapshot as
complete hides added paths; discarding it loses collision checks and
package merges for unchanged paths.

Mark patched layouts incomplete and retain whole-wheel fallback for
patch-added paths. Preserve observed topology for collision and merge
planning. Reject patches that change an observed path's file or
directory kind or package classification.

Keep original root `.pth` files and distribution metadata under one
runtime owner. Reject only collisions that leave unsuppressible fallback
copies. Console-script declarations remain pre-patch metadata.

### DIFF
--- a/e2e/cases/azure-core-tracing-overlap/BUILD.bazel
+++ b/e2e/cases/azure-core-tracing-overlap/BUILD.bazel
@@ -18,12 +18,15 @@ load("@aspect_rules_py//py:defs.bzl", "py_test")
 # Variants:
 #   :test       — py_test default (.pth strategy)
 #   :test_tool  — py_venv_test variant (symlink-forest venv)
+# Both use Python 3.11 because post_install.patch names the interpreter-specific
+# lib/python3.11/site-packages install path.
 
 py_test(
     name = "test",
     srcs = ["test.py"],
     dep_group = "azure-core-tracing-import",
     main = "test.py",
+    python_version = "3.11",
     deps = [
         "@pypi_azure_core_tracing//azure_core",
         "@pypi_azure_core_tracing//azure_core_tracing_opentelemetry",
--- a/py/private/py_venv/venv.bzl
+++ b/py/private/py_venv/venv.bzl
@@ -132,7 +132,7 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
 
     # Pass 1: bucket claimants per import root, distribution metadata entry,
     # and console-script name.
-    tl_claimants = {}  # tl -> list of struct(site_packages, is_ns, ns_entries)
+    tl_claimants = {}  # tl -> list of struct(site_packages, layout_complete, is_ns, ns_entries)
     metadata_claimants = {}  # metadata entry -> ordered site_packages paths
     cs_claimants = {}  # name -> list of struct(site_packages, module, func)
     wheel_by_sp = {}  # site_packages_rfpath -> wheel struct
@@ -148,6 +148,7 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
                 continue
             tl_claimants.setdefault(tl, []).append(struct(
                 site_packages = w.site_packages_rfpath,
+                layout_complete = w.layout_complete,
                 is_ns = tl in ns_set,
                 ns_entries = tuple(ns_entries_by_tl.get(tl, [])),
             ))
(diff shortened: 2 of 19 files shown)


tamird:
Extracted from #1166.

Auditing whether each `uv.override_package` setting was consumed exposed
that post-install patches run after wheel topology is extracted, leaving
venv assembly to trust stale package paths.

jbedard:
Why is the new `layout_complete` field required?

e2e/cases/azure-core-tracing-overlap/BUILD.bazel
```diff
@@ -24,6 +24,7 @@ py_test(
     srcs = ["test.py"],
     dep_group = "azure-core-tracing-import",
     main = "test.py",
+    python_version = "3.11",
```
jbedard, line 29: Why is this required?

tamird:
`top_levels` previously encoded two states: empty meant unknown, and
nonempty meant complete. Post-install patches introduce a third state:
the pre-patch entries are known, but the patch may add entries.

Venv assembly must retain the known entries for collision detection and
package-merge planning, while also retaining whole-wheel fallback so
patch-added entries remain reachable. Emptying `top_levels` loses the
former; treating it as complete loses the latter. `layout_complete`
carries that distinction.

py/private/py_venv/venv.bzl
```diff
@@ -181,8 +182,11 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
     ns_claimant_sps = {}  # sp -> True, wheels in any all-namespace collision
     for tl, claimants in tl_claimants.items():
         distinct_sp = {c.site_packages: c for c in claimants}
+        processed_pth = tl.endswith(".pth") and not tl.startswith(".")
```
jbedard: Why? What about `.gitignore` or silly other filenames?

py/tools/unpack/unpack.py
```diff
@@ -189,6 +190,27 @@ def main():
         args.wheel,
     )
 
+    site_packages = (
```
jbedard, line 193: Can any of the unpack logic or validation be in a prefactor?

tamird: Yes. I extracted the generic validation into #1200, directly on `main`. It rejects removal, file/directory changes, and import-package reclassification of repository-inspected paths while allowing ordinary content edits. After that lands, I will rebase this PR so it retains only the `layout_complete` and whole-wheel-fallback behavior needed for patch-added paths.

e2e/cases/azure-core-tracing-overlap/BUILD.bazel
tamird: The post-install patch is applied relative to the install root and names `lib/python3.11/site-packages/azure/core/patched_marker.py`, so both variants must resolve Python 3.11. I added that explanation above the two targets.

py/private/py_venv/venv.bzl
tamird: This now mirrors and cites CPython `site.addsitedir`: only non-hidden names ending in `.pth` are processed. I renamed the variable accordingly. `.gitignore`, `.hidden.pth`, and other names follow the ordinary top-level collision path.

py/private/py_venv/venv.bzl
```diff
@@ -148,6 +148,7 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
                 continue
             tl_claimants.setdefault(tl, []).append(struct(
                 site_packages = w.site_packages_rfpath,
+                layout_complete = w.layout_complete,
```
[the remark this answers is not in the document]
tamird: I am not adding this fallback. `PyWheelsInfo` is a private rules_py provider, and this change updates every in-repository wheel-record producer: `whl_install`, `py_unpacked_wheel`, and the two test producers. `py_library` only forwards their depset. Accepting pre-change ad hoc structs would create a compatibility contract for private internals and would make the new known-but-incomplete layout state ambiguous.

tamird:
I no longer have a consumer whose post-install patch adds a new top-level entry. The incomplete-layout state complicates analysis without repairing a demonstrated failure. #1200 separately rejects destructive topology changes and stands on its own.

PR #1194, opened 2026-06-25
fix(py): union colliding package directories

tamird:
The atomic top-level venv layout selects one wheel when ordinary package
directories collide. Apache Airflow splits its regular `airflow` package
across core and provider distributions, so unique provider modules
disappear.

In permissive collision modes, union every ordinary directory claimant
at the final `site-packages` path. Sort inputs by runfiles path and
validate all claims before materializing the output. Disjoint entries
and identical files with matching executable bits are compatible.
File-valued top-levels, differing contents or executable bits, and
file/directory conflicts fail instead of selecting code by depset
traversal order.

Apply the same validation to regular packages spanning namespace wheels.
This restores flat-install directory behavior without analysis-time
entry-type metadata or an arbitrary permissive winner.


tamird:
Extracted from #1155.

Reviewing ambiguous namespace-layout merges exposed that permissive
ordinary-directory collisions still selected one wheel and dropped
distinct files from regular packages split across distributions.

tamird:
Closing this rather than stacking another repair on it.

CI exposed a deterministic analysis error: `top_levels` includes both
files and directories, so this patch declared top-level files such as
`CHANGELOG.md` as tree artifacts and sent them to the directory merge
action.

The repair needs analysis-time directory classification. I will rebuild
#1172 directly on `main`, land that metadata contract first, and then
recreate this directory-union change directly on the resulting `main`.
The replacement will also retain the existing postorder overlay semantics
for `warning` and `ignore`; unconditional order-independent rejection
was a separate collision-policy change.

PR #1196, opened 2026-06-25
fix(uv): describe installed wheel topology

tamird:
Wheel repository analysis records every immediate `site-packages`
entry, but `PyWheelsInfo` does not preserve which regular packages are
directories. Venv analysis must choose between a symlink and a tree
artifact before an action runs, so it cannot safely merge colliding
package directories from names alone.

PEP 420 namespaces already have dedicated directory metadata. Carry only
regular top-level directories as positive pre-patch evidence for the
selected wheel. Source-built and patched wheels, absent fields, and
unlisted names remain unknown.

This lets consumers distinguish mergeable directory collisions without
classifying modules such as `CHANGELOG.md` as tree artifacts or
expanding the public `py_unpacked_wheel` API.

### DIFF
--- a/e2e/snapshots/abi3_compat.whl_install.cffi.BUILD.bazel
+++ b/e2e/snapshots/abi3_compat.whl_install.cffi.BUILD.bazel
@@ -115,6 +115,44 @@ whl_install(
             "cffi-2.0.0.dist-info",
         ],
     },
+    regular_directory_top_levels = {
+        "cffi-2.0.0-cp312-cp312-macosx_10_13_x86_64.whl": [
+            "cffi",
+        ],
+        "cffi-2.0.0-cp312-cp312-macosx_11_0_arm64.whl": [
+            "cffi",
+        ],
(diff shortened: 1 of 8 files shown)


tamird:
Extracted from #1142.

Reworking that wheel-topology draft exposed that `RECORD` parsing knew
which regular top-level packages were directories, but `PyWheelsInfo`
erased the distinction. Venv analysis therefore had no safe way to
choose between a symlink and a tree artifact for a collision.

This replaces #1172 because GitHub will not reopen a pull request after
its fork branch is recreated. eeaa3f39 is based directly on current
`main` and has no precursor PRs. It must land before I recreate the
consumer from closed #1194; I will not stack that consumer on this
branch.

jbedard:
This adds an API and information without showing how it will be used. What is this for?

tamird:
Agreed. The metadata is not useful by itself. I am folding the directory-union consumer into this PR, broadening the evidence to every positively known directory top level, and preserving the existing postorder collision policy. I will update it after #1200 and #1193 settle because patched-wheel directory evidence is only sound once the unpack action validates the inspected topology.

tamird:
This metadata existed to support a directory-union consumer that is no longer justified. No current consumer uses the added API, so I am closing it rather than publishing information without a demonstrated need.

PR #1200, opened 2026-06-25
fix(uv): reject patched wheel topology changes

tamird:
Repository evaluation inspects each selectable prebuilt wheel before
post-install patches run. It records the paths used for top-level
projection, collision handling, and regular-package merges. `WhlInstall`
then exports that metadata beside the patched install tree. Deleting or
reclassifying an observed path therefore leaves venv analysis planning
against a tree the action no longer produces.

Patched-wheel assembly needs the pre-patch metadata as a lower bound:
unchanged paths still participate in collisions and package merges,
while a whole-wheel fallback can retain additions. That lower bound is
sound only if every path consumed by analysis retains its
file-or-directory kind and package classification.

Pass those paths into `WhlInstall`, snapshot their shape after
unpacking, and reject incompatible changes after applying patches. This
check belongs in the install action because Bazel analysis cannot
inspect the generated tree. Content edits and ordinary additions remain
supported, although new prebuilt-wheel paths may not appear in generated
venv projections. Source-built topology remains unvalidated because
it is unavailable during repository evaluation. Use full package
replacement when the installed topology itself must change.


tamird:
Extracted from #1193. The need arose because #1193 retains repository-inspected topology for patched wheels; that is only sound if the unpack action proves every advertised path still has the same file/directory and package-classification shape after patching.

jbedard:
Add an e2e test.sh showing that builds failure if a patch breaks the rules

tamird:
Will do. I will add a standalone e2e module whose post-install patch removes a regular package's `__init__.py`, then have `test.sh` build the real wheel target and require the unpack action's package-classification diagnostic.

tamird:
Done.

e2e/cases/uv-patched-topology-change/MODULE.bazel
```diff
@@ -0,0 +1,35 @@
+module(name = "uv_patched_topology_change")
```
jbedard: Can we avoid a new MODULE?

tamird: Done.

jbedard:
can I merge this?

tamird:
Ready from my side: the final matrix is green and all review threads are resolved.

xangcastle (approved):

PR #1201, opened 2026-06-26
fix(py_venv): preserve PBS base for nested venvs

### LINKED ISSUE #1048: [Bug]: v2.0.0-alpha.1 breaks sys.base_prefix for Python 3.11/3.12 with native extensions

tamird:
CPython 3.11 and 3.12 interpret a relative pyvenv.cfg home from the process working directory. rules_py 2 writes a runfiles-relative value. When a py_binary runs as a Bazel action tool, that directory is the action sandbox, so prefix discovery can fall back to the PBS build-time /install prefix before Python can import encodings.

py_venv owns both the generated pyvenv.cfg and the relocatable bin/python link. For non-Windows in-build CPython 3.11 and 3.12 runtimes that declare supports_build_time_venv, write home with an empty value. getpath sees the key and resolves bin/python to the real PBS base executable, while the empty value does not force cwd-relative prefix discovery. That keeps outer startup relocatable and lets stdlib-created child venvs inherit PBS rather than the outer venv bin directory. System interpreters, unsupported runtimes, Windows, and other Python versions retain their current home.

The e2e matrix runs PBS 3.9 through 3.13 from an unrelated action cwd. It exercises the launcher, child interpreters with and without site initialization, the exposed venv, and a nested stdlib venv. The nested child starts from that unrelated cwd, imports encodings, and rejects /install as its base prefix. Two checked-in pyvenv.cfg snapshots make the generated metadata delta reviewable: 3.11 keeps an empty home line, while 3.13 retains a normalized nonempty home line.

Fix #1048.

### DIFF
--- a/e2e/cases/interpreter-local-pyvenv-home/BUILD.bazel
+++ b/e2e/cases/interpreter-local-pyvenv-home/BUILD.bazel
@@ -18,7 +18,8 @@ load("@rules_python//python:py_runtime_pair.bzl", "py_runtime_pair")
 
 py_runtime(
     name = "system_py3_runtime",
-    interpreter_path = "/opt/fake-python/bin/python3",
+    # Braces make sure the generated home line is not formatted a second time.
+    interpreter_path = "/opt/{fake-python}/bin/python3",
     interpreter_version_info = {
         "major": "3",
         "minor": "94",
--- a/e2e/cases/pbs-symlink-prefix-1048/action_probe.bzl
+++ b/e2e/cases/pbs-symlink-prefix-1048/action_probe.bzl
@@ -1,0 +1,65 @@
+"""Action-level PBS venv regression target."""
+
+load("@aspect_rules_py//py:defs.bzl", "py_binary")
+
+def pbs_action_probe(name, python_version):
+    """Run a PBS-backed binary from a non-runfiles action cwd."""
+    py_binary(
+        name = name,
+        srcs = ["test_pbs_prefix.py"],
+        expose_venv = True,
+        main = "test_pbs_prefix.py",
+        python_version = python_version,
+    )
+
+    commands = [
+        "root=$$(pwd)",
+        "mkdir -p $(@D)/cwd",
+        "cd $(@D)/cwd",
+        (
+            "\"$$root/$(execpath :{name})\" " +
+            "--expected-cwd \"$$PWD\" --test-children"
+        ).format(name = name),
+        (
+            "\"$$root/$(execpath :{name}.venv)\" " +
+            "\"$$root/$(location test_pbs_prefix.py)\" " +
+            "--expected-cwd \"$$PWD\""
+        ).format(name = name),
+        "touch \"$$root/$@\"",
+    ]
+
+    native.genrule(
+        name = name + "_output",
+        srcs = ["test_pbs_prefix.py"],
+        outs = [name + ".stamp"],
+        cmd = "\n".join(commands),
+        tools = [
+            ":" + name,
+            ":" + name + ".venv",
+        ],
+    )
+
+def pbs_pyvenv_cfg_snapshot(name, venv):
+    """Copy a PBS-backed py_venv's generated pyvenv.cfg from its runfiles."""
+    native.genrule(
+        name = name,
+        testonly = True,
+        outs = [name],
+        cmd = """
+            launcher=$(execpath {venv})
+            runfiles="$$launcher".runfiles
+            pkg=$$(dirname "$$launcher" | sed 's|^bazel-out/[^/]*/bin/||')
+            vname=$$(basename "$$launcher")
+            cfg="$$runfiles/_main/$$pkg/.$$vname/pyvenv.cfg"
+            if [ ! -f "$$cfg" ]; then
+                echo "expected pyvenv.cfg at $$cfg, not found" >&2
+                ls -la "$$runfiles/_main/$$pkg/" 2>&1 >&2
+                exit 1
+            fi
+            # PBS interpreter repository names include the host OS and CPU.
+            # Keep the `home` line while removing only that volatile value.
+            sed -E 's|^home = .+|home = <PBS_INTERPRETER_HOME>|' "$$cfg" > $@
+        """.format(venv = venv),
+        tools = [venv],
+        visibility = ["//:__pkg__"],
+    )
--- a/e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
+++ b/e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
@@ -1,48 +1,121 @@
-"""Regression test for issue #1048: PBS symlink chain breaks sys.base_prefix.
+"""Exercise a PBS-backed venv from a Bazel action working directory."""
 
-Python 3.11/3.12 reimplemented prefix discovery in pure Python (getpath.py).
-Its resolvedpath() has a bug with multi-hop relative symlinks that traverse
-.. components across repo boundaries — the exact shape of the venv's
-bin/python → ../../../../interpreter_repo/bin/python3.11 chain.
-
-When resolution fails, Python falls back to the compiled-in /install prefix
-(absent at runtime), causing:
-    ModuleNotFoundError: No module named 'encodings'
-
-The fix: pyvenv.cfg's home= key now points directly to the PBS bin/ directory,
-so Python only resolves the local python → python3.11 symlink (one hop).
-"""
-
+import argparse
 import os
+import subprocess
 import sys
+import tempfile
 
-from verify_venv import verify_all, verify_base_prefix, verify_in_venv
 
+def _verify_prefixes(expected_cwd: str) -> None:
+    assert os.path.samefile(os.getcwd(), expected_cwd), (os.getcwd(), expected_cwd)
+    assert os.path.isabs(sys.base_prefix), sys.base_prefix
+    assert os.path.isdir(sys.base_prefix), sys.base_prefix
+    assert sys.base_prefix != "/install", sys.path
 
-def test_base_prefix_not_install():
-    assert sys.base_prefix != "/install", (
-        f"sys.base_prefix is '/install' (the PBS compile-time prefix). "
-        f"Python {sys.version_info.major}.{sys.version_info.minor} failed to "
-        f"resolve the pyvenv.cfg home= symlink chain."
+    pyvenv_cfg = os.path.join(os.path.dirname(sys.executable), "..", "pyvenv.cfg")
+    with open(pyvenv_cfg, encoding="utf-8") as cfg:
+        config = cfg.read().splitlines()
+    assert "relocatable = true" in config, config
+    home = next(
+        (line.partition("=")[2].strip() for line in config if line.startswith("home =")),
+        None,
     )
 
+    expect_empty_home = os.name != "nt" and sys.version_info[:2] in {
+        (3, 11),
+        (3, 12),
+    }
+    assert (home == "") == expect_empty_home, (home, sys.version_info)
+    if expect_empty_home:
+        assert sys._base_executable != sys.executable, sys.executable
+        assert os.path.samefile(
+            os.path.dirname(os.path.dirname(sys._base_executable)),
+            sys.base_prefix,
+        ), (
+            sys._base_executable,
+            sys.base_prefix,
+        )
 
-def test_base_prefix_has_stdlib():
+    if not sys.flags.no_site:
+        assert os.path.isabs(sys.prefix), sys.prefix
+        assert os.path.isdir(sys.prefix), sys.prefix
+        assert sys.prefix != sys.base_prefix, (sys.prefix, sys.base_prefix)
+
     stdlib = os.path.join(
         sys.base_prefix,
         "lib",
         f"python{sys.version_info.major}.{sys.version_info.minor}",
     )
-    assert os.path.isdir(stdlib), (
-        f"stdlib missing: {stdlib!r} (sys.base_prefix={sys.base_prefix!r})"
+    assert os.path.isdir(stdlib), (stdlib, sys.path)
+
+
+def _verify_child(options: list[str], expected_cwd: str) -> None:
+    subprocess.run(
+        [
+            sys.executable,
+            *options,
+            __file__,
+            "--expected-cwd",
+            expected_cwd,
+        ],
+        check=True,
     )
+
+
+def _verify_nested_venv(expected_cwd: str) -> None:
+    with tempfile.TemporaryDirectory(dir=expected_cwd) as root:
+        child_venv = os.path.join(root, "child")
+        subprocess.run(
+            [sys.executable, "-m", "venv", "--without-pip", child_venv],
+            cwd=expected_cwd,
+            check=True,
+        )
+
+        with open(os.path.join(child_venv, "pyvenv.cfg"), encoding="utf-8") as cfg:
+            child_config = cfg.read().splitlines()
+        child_home = next(
+            (
+                line.partition("=")[2].strip()
+                for line in child_config
+                if line.startswith("home =")
+            ),
+            None,
+        )
+        assert child_home == os.path.dirname(os.path.abspath(sys._base_executable)), (
+            child_home,
+            sys._base_executable,
+        )
+
+        child_python = os.path.join(
+            child_venv,
+            "Scripts" if os.name == "nt" else "bin",
+            "python.exe" if os.name == "nt" else "python",
+        )
+        subprocess.run(
+            [
+                child_python,
+                "-c",
+                "import encodings, os, sys; "
+                "assert sys.base_prefix != '/install', sys.path; "
+                "assert os.path.isdir(sys.base_prefix), sys.base_prefix; "
+                "assert sys.prefix != sys.base_prefix, "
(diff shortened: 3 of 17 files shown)


e2e/cases/pbs-symlink-prefix-1048/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# The root module's Python 3.12 toolchain comes from rules_python, while the
+# e2e module uses rules_py's PBS extension. Exercise the original #1048
+# producer explicitly as well as the rules_py action probes in this package.
+cd "$(dirname "$0")/../../.."
+bazel test --//py/private/interpreter:python_version=3.12 -- \
```
jbedard: only use `test.sh` when testing for expected bazel failures, not for succesful builds

e2e/cases/pbs-symlink-prefix-1048/action_probe.bzl
```diff
@@ -0,0 +1,45 @@
+"""Action-level PBS venv regression target."""
+
+load("@aspect_rules_py//py:defs.bzl", "py_binary")
+
+def pbs_action_probe(name, python_version, test_children, test_venv):
```
jbedard: Why ever set `test_children` or `test_venv` to False? Why not always test?

e2e/cases/pbs-symlink-prefix-1048/test.sh
tamird: Done.

e2e/cases/pbs-symlink-prefix-1048/action_probe.bzl
tamird: Done.

tamird:
I found a contract gap in my own review, so I moved this back to draft. Omitting `home` is enough for current CPython 3.11/3.12 startup, but it is outside the PEP 405 venv contract: the PEP defines `home` as the venv marker and says that without it `sys.prefix == sys.base_prefix` ([PEP 405](https://peps.python.org/pep-0405/#specification)). A strict metadata consumer such as Ty rejects this shape, so the current test proves only the current CPython implementation, not the public `py_venv` contract. The latest test rewrite also removed `test.sh`, which was the explicit rules_python source-wheel producer check matching #1048. I am checking whether rules_py should define `relocatable = true` as an intentional extension and test it as such, or keep a parseable `home` through a different launcher/runtime design. Until that ownership boundary is explicit and the producer coverage is back, this should not merge.

tamird:
Correction to my previous note: I overstated the `test.sh` point. The root CI already runs `aspect test -- //...`, which includes `//:sdist_fallback_with_anyarch_wheel_build_test`; deleting the nested script did not remove that producer coverage. The open issue is the metadata contract: no-`home` is a CPython-specific relocatable extension, while PEP 405 and strict consumers treat it as outside a standard venv.

tamird:
Pushed `edf92694` and rewrote the body around the explicit rules_py relocatable-metadata contract. I am keeping this draft until the refreshed matrix is green and the local Ty adapter is proven against the real targets.

tamird:
Ready again: the refreshed upstream matrix is green, the contract review is clean, and the real Ty adapter targets pass with the strict parser kept outside this relocatable metadata path.

tamird:
@xangcastle could you review this when you have a chance? It is the same py_venv ownership boundary as PR 1147, and the full matrix is green.

e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
```diff
@@ -1,48 +1,80 @@
...
+    config_values = {}
+    for line in config.splitlines():
+        key, separator, value = line.partition("=")
+        assert separator and key.strip() and value.strip(), line
+        config_values[key.strip().lower()] = value.strip()
+    assert config_values["relocatable"] == "true", config_values
+    home = config_values.get("home")
 
-def test_base_prefix_not_install():
-    assert sys.base_prefix != "/install", (
```
jbedard, line 23: Did any of these assertions change or only move? The diff is unclear

jbedard:
Add a snapshot test that shows the diff of the generated code

tamird:
Done.

e2e/cases/pbs-symlink-prefix-1048/test_pbs_prefix.py
tamird: They changed. The old test called verify_all(), then repeated the /install and stdlib checks. The action-cwd probe keeps the #1048 checks and adds cwd/PWD, base_exec_prefix and executable existence, relocatable/home metadata, _ctypes, child -S/-BS, exposed venv, and 3.9-3.13 controls. It drops the incidental verify_all() symlink-walk and sys.path checks because this fixture now targets startup from an arbitrary action cwd rather than general venv layout.

py/private/py_venv/venv.bzl
```diff
@@ -705,31 +712,42 @@ def assemble_venv(
...
+    omit_venv_home = (
+        py_toolchain.runfiles_interpreter and
+        not is_windows and
+        getattr(py_toolchain.toolchain, "implementation_name", None) == "cpython" and
+        getattr(py_toolchain.toolchain, "supports_build_time_venv", False) and
+        py_toolchain.interpreter_version_info.major == 3 and
+        py_toolchain.interpreter_version_info.minor in [11, 12]
+    )
+    if omit_venv_home:
+        pyvenv_home = None
```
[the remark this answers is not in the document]
tamird: False positive. The action probe on this head asserts sys.prefix != sys.base_prefix for 3.11 and 3.12. I also ran the exposed 3.11 venv directly: sys.prefix was the generated .action_probe_311.venv, sys.base_prefix was the PBS interpreter, and sys.path included the generated venv site-packages directory. The runtime test executes both the launcher and exposed venv from an unrelated action cwd; omitting home does not disable venv activation in current CPython.

py/private/py_venv/venv.bzl
```diff
@@ -705,31 +712,42 @@ def assemble_venv(
...
+    # so getpath derives the base from the relocatable bin/python symlink.
+    # A relative `home` instead resolves from the startup cwd:
+    # https://github.com/python/cpython/blob/3bb231a6/Modules/getpath.py#L362-L365
+    # https://github.com/python/cpython/blob/3bb231a6/Modules/getpath.py#L431-L432
+    # Direct runtimes use that symlink. The capability also permits wrappers
+    # that set PYTHONEXECUTABLE to preserve the underlying base executable:
+    # https://github.com/bazel-contrib/rules_python/blob/bac54949/python/private/py_runtime_info.bzl#L316-L337
+    omit_venv_home = (
+        py_toolchain.runfiles_interpreter and
+        not is_windows and
```
jbedard, line 759: Why the exception for windows?

py/private/py_venv/venv.bzl
```diff
@@ -705,31 +712,42 @@ def assemble_venv(
...
+        getattr(py_toolchain.toolchain, "implementation_name", None) == "cpython" and
+        getattr(py_toolchain.toolchain, "supports_build_time_venv", False) and
+        py_toolchain.interpreter_version_info.major == 3 and
+        py_toolchain.interpreter_version_info.minor in [11, 12]
```
jbedard, line 763: Does 13 and higher explicitly not need this?

jbedard: Can this be a <13 instead?

jbedard:
Rebase

py/private/interpreter/repository.bzl
```diff
@@ -229,6 +229,7 @@ py_runtime(
     name = "py3_runtime",
     abi_flags = "{abi_flags}",
     files = [":files"],
+    implementation_name = "cpython",
```
jbedard: Why is this required?

jbedard: Ensure this diff is shown in a snapshot test

py/private/py_venv/venv.bzl
tamird: Windows is excluded deliberately because this workaround relies on the POSIX relocatable bin/python symlink path, and this fixture has no Windows repro. I am leaving Windows unchanged until there is a failing Windows case rather than broadening the extension without evidence.

py/private/py_venv/venv.bzl
tamird: 3.13 is explicitly the retained-home negative control here: action_probe_313 runs from the same unrelated action cwd and its snapshot keeps home. 3.9 and 3.10 are retained-home controls too, so <13 would broaden the behavior to known-passing versions. The allowlist is intentionally the reproduced 3.11/3.12 pair; this does not claim untested later versions need or do not need it.

py/private/interpreter/repository.bzl
tamird: Not required. rules_python 1.9.0 py_runtime already defaults implementation_name to cpython, and the existing interpreter-runtime-metadata test asserted that generated PBS runtime field before this PR. I removed the redundant line in the rebase.

tamird: I removed the redundant generated line, so there is no repository BUILD diff left to snapshot. The checked-in pyvenv.cfg snapshots continue to show the actual metadata behavior: 3.11 omits home and 3.13 retains it.

tamird:
Done.

py/private/py_venv/venv.bzl
jbedard: What wil happen when 3.13 is released?

jbedard:
squash, reduce unnecessary diffs

py/private/py_venv/venv.bzl
tamird: 3.13 is already covered here: action_probe_313 runs from the same unrelated action cwd with home retained, and its checked-in snapshot pins that line. If a later CPython minor reproduces the cwd bug, it should enter this allowlist with that failing case; the current change does not preemptively change unreproduced versions.

tamird:
Done.

tamird:
Correction: my earlier “Not required” reply was wrong. Exact downstream CI with the explicit line removed restored the /install encodings failure in four nested-venv tests. The gate reads the selected PyRuntimeInfo after toolchain plumbing, where the implicit default is not reliable. I restored the explicit producer-owned field in fdc79eab and updated the body.

tamird:
Second correction: the explicit field alone did not fix the exact downstream subset locally. The remaining failure is a nested stdlib venv whose home points at the outer relocatable venv bin directory, then falls back to /install. I moved this back to draft while I add that missing ownership case instead of asking you to review an incomplete fix.

jbedard:
Create an e2e test reproducing the same issue as downstream

tamird:
Done. The action probe now creates a nested stdlib venv from the unrelated action cwd, starts its child interpreter, imports encodings, and rejects /install. Empty home fixes the three downstream nested-venv failures. The fourth downstream failure was our monorepo test's detached venv alias, not this metadata path; I narrowed that test back to its shell-trampoline contract. This reroll also removes the unproven implementation_name line. The exact four-test subset passes locally.

py/private/py_venv/venv.bzl
```diff
@@ -705,31 +707,46 @@ def assemble_venv(
...
+        pyvenv_home = ""
+    elif py_toolchain.runfiles_interpreter:
         pbs_rlocation = to_rlocation_path(ctx, py_toolchain.python)
         pbs_bin_dir = "/".join(pbs_rlocation.split("/")[:-1])
         pyvenv_home = "{}/{}".format(venv_to_runfiles_escape, pbs_bin_dir)
     else:
         pyvenv_home = py_toolchain.python.path.rsplit("/", 1)[0]
 
     pyvenv_cfg = ctx.actions.declare_file("{}/pyvenv.cfg".format(venv_name))
+    home_line = "home =\n" if pyvenv_home == "" else "home = {}\n".format(pyvenv_home)
```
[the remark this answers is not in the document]
tamird: Done.

tamird:
@jbedard Could you review the current exact remaining downstream set: #1201, #1205, #1206, and #1212? #1202 is merged. All four heads are green; #1205 is rerolled at 718f03a1 with the google/protobuf e2e, and #1212 is the separate dirname-free console-script fix exposed by the downstream manifest-only check.

tamird:
Update to the exact remaining downstream set: #1201, #1205, #1206, #1212, #1213, and #1214. #1202 is merged. #1213 is the downstream fasttext C++ runtime failure; #1214 is the OpenAPI binary-as-dep PyWheelsInfo loss. All six public heads are green.

e2e/cases/interpreter-local-pyvenv-home/BUILD.bazel
```diff
@@ -18,7 +18,8 @@ load("@rules_python//python:py_runtime_pair.bzl", "py_runtime_pair")
 
 py_runtime(
     name = "system_py3_runtime",
-    interpreter_path = "/opt/fake-python/bin/python3",
+    # Braces make sure the generated home line is not formatted a second time.
```
jbedard, line 21: Why is this required?

tamird: This is the regression input for the brace-preservation fix in venv.bzl. interpreter_path is a host path string and may contain literal braces; before the change we formatted the home line once and then fed it through the outer pyvenv.cfg format, so {fake-python} was reinterpreted and either rewrote the path or failed analysis. Keeping braces here makes test_pyvenv_home fail on that bug; a plain path only covers the older home-directory behavior.

jbedard:
Review these changes

tamird:
I rechecked the current public stack against the pinned integration and reran the PBS action-cwd regression uncached. The three live threads above already have the evidence I have; no code change is pending from them. When you have a chance, could you confirm whether those answers are sufficient or point to a remaining change?

e2e/cases/snapshots/pbs_symlink_prefix_1048.python311.pyvenv.cfg
```diff
@@ -0,0 +1,6 @@
+home =
```
jbedard, line 1: Why is home blank in this scenario? Should it be omitted instead?

jbedard (approved):

xangcastle (approved):

PR #1204, opened 2026-06-26
fix(uv): align C++ toolchain platform

tamird:
Extracted from PR 1167. That fix moved the PEP 517 frontend onto the native wheel action target execution group, but cumulative review found that native wheel C++ lookup still read the default group. When the platforms differ, CC and CXX and their files can come from the host toolchain while the action executes on the target-matching platform.

Resolve the C++ toolchain from the named target execution group, unwrap Bazel current-toolchain providers there, and remove the unused default-group C++ registration. Keep frontend configuration, action execution, and toolchain lookup on one shared group constant.

The e2e fixture registers a platform-selected synthetic C++ toolchain, transitions two target platforms, and executes CC inside the real frontend. It fails under the old default lookup and passes only when compiler identity matches the action group.

### DIFF
--- a/e2e/cases/pep517-frontend-exec-group/BUILD.bazel
+++ b/e2e/cases/pep517-frontend-exec-group/BUILD.bazel
@@ -3,6 +3,9 @@ load("@bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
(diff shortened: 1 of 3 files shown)


tamird:
Extracted from #1167. That PR correctly aligned the PEP 517 frontend with the named target execution group, but the later cumulative review found one remaining consumer of the default group: native wheel C++ toolchain lookup. The synthetic CC check in this draft is the regression proof for that exact split-platform failure.

tamird:
Ready for review: the cumulative review is clean, the old default-exec-group lookup fails the new synthetic C++ fixture, and the full Bazel 8/9 matrix is green.

xangcastle (approved):

PR #1208, opened 2026-06-27
fix(uv): deduplicate whl_install wheel file labels

jimmyt857:
## Summary

`uv.lock` can contain multiple wheel entries that resolve to the same fetched wheel repository label. One way this happens is when duplicate package records have wheel URLs with different basenames but the same hash, so `rules_py` fetches them through the same `http_file` repository.

Before this change, the uv extension forwarded those duplicate labels directly to the `whl_install.whl_files` `label_list` attr. Bazel rejects duplicate labels during repository rule instantiation, so the install fails before wheel metadata can be extracted.

This change deduplicates the non-empty wheel file labels while preserving order, then changes `whl_install` metadata extraction to map the deduplicated labels back to the serialized prebuild targets by label string instead of relying on positional pairing. That keeps the label_list valid without losing top-level metadata for selectable wheels.

## Regression Test

Adds `e2e/cases/uv-duplicate-whl-labels`, whose lockfile contains two `six==1.16.0` wheel entries with the same hash and different wheel basenames. Without the fix, the test fails during repository rule setup with a duplicate `whl_files` label error.

## Testing

- `bazel run //:gazelle`
- `bazel test //uv/private/extension:all //uv/private/whl_install:all`
- `(cd e2e && bazel test //cases/uv-duplicate-whl-labels:test_import //cases/uv-dup-lock-records:all)`
- `pre-commit run buildifier --files e2e/MODULE.bazel e2e/cases/uv-duplicate-whl-labels/BUILD.bazel e2e/cases/uv-duplicate-whl-labels/setup.MODULE.bazel uv/private/extension/defs.bzl uv/private/whl_install/repository.bzl`

Note: `pre-commit run buildifier-lint --files ...` still reports existing warnings in touched upstream files, so I kept this patch focused on the regression rather than sweeping unrelated lint cleanup.


uv/private/whl_install/repository.bzl
```diff
@@ -118,6 +118,25 @@ def _find_whl_file(repository_ctx, whl_label):
             return entry
     return None
 
+def _whl_file_label_keys(whl_file_label):
+    label_str = str(whl_file_label)
+    keys = {label_str: True}
+    canonical_prefix = "@" + "@"
```
jbedard, line 124: What create this var via concatenation? lol
Why not just hardcode "@@" like we do with "//"?

uv/private/whl_install/repository.bzl
```diff
@@ -118,6 +118,25 @@ def _find_whl_file(repository_ctx, whl_label):
             return entry
     return None
 
+def _whl_file_label_keys(whl_file_label):
```
acozzette, line 121: I think this function could use a short docstring describing what it does. I gather it returns the original label string along with some alternative strings representing the same label?

uv/private/whl_install/repository.bzl
```diff
@@ -669,30 +688,26 @@ filegroup(
...
-    # of production and does not depend on runtime dict iteration order.
+    # JSON-encoded string of labels) because only the former adds the wheel
+    # repos to our visibility so `rctx.path(Label(...))` can resolve. Multiple
+    # lockfile wheel entries can resolve to the same target, but Bazel label_list
+    # attrs reject duplicates, so map the deduplicated labels back by canonical
+    # label string instead of pairing by list index.
     whl_file_labels = {}
-    whl_file_index = 0
-    for target in prebuilds.values():
-        if not target:
```
jbedard, line 682: This was dropped and not moved to anywhere new?

uv/private/whl_install/repository.bzl
```diff
@@ -669,30 +688,26 @@ filegroup(
...
     namespace_entries_by_whl = {}
     namespace_dirs_by_whl = {}
     regular_roots_by_whl = {}
     console_scripts_by_whl = {}
-    for target, whl_file_label in whl_file_labels.items():
+    for target in prebuilds.values():
         if target not in arm_targets:
             continue
+        whl_file_label = whl_file_labels[target]
         whl_name, tls, regular, css, ns_entries, dirs_set, init_dirs = _extract_wheel_metadata(
```
jbedard, line 711: This is now invoked once per `prebuilds` instead of once per `whl_file_labels` - is that ok? Should that be changed?

xangcastle (approved):

PR #1214, opened 2026-06-28
fix(py): preserve wheel metadata through binaries

tamird:
OpenAI Buildkite 5146797 exposed this through nested Python targets: an outer venv received wheel import paths from a `py_binary` dependency but lost the wheel collision records, so it could not reproduce the inner target's package overlay and `azure.core.tracing.ext.opentelemetry_span` disappeared before OpenAPI generators ran.

`py_binary` and `py_test` split dependencies onto a sibling venv, then return a launcher that previously surfaced only `PyInfo`. Forward that sibling venv's existing `PyWheelsInfo` from the launcher as well. The launcher does not rediscover or own wheel records; it preserves the metadata already owned by the sibling venv so an outer executable can make the same collision and namespace decisions.

The existing `pth-namespace-547` e2e now nests a `py_binary` that owns the jaraco wheels under a `py_test` that depends only on that binary. Without this change, the outer venv keeps only the fallback `.pth` and lacks the concrete `site-packages/jaraco` namespace asserted by the fixture. With the forwarded records, the outer projection matches the inner binary.

### DIFF
--- a/e2e/cases/pth-namespace-547/BUILD.bazel
+++ b/e2e/cases/pth-namespace-547/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@aspect_rules_py//py:defs.bzl", "py_test", "py_unpacked_wheel")
+load("@aspect_rules_py//py:defs.bzl", "py_binary", "py_test", "py_unpacked_wheel")
 load("//tools:pth_snapshot.bzl", "extract_venv_pth")
 
 # Test namespace package resolution with the symlink-forest venv.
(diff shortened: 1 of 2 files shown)


tamird:
The full matrix is green at 10b29bce. @jbedard, this is the downstream OpenAPI generator failure: a py_binary used as a dep now forwards its sibling venv wheel closure so the outer venv can preserve concrete namespace and regular-overlay planning. Ready for independent review.

tamird:
The full matrix is green at `10b29bce`. @jbedard, this is the nested-binary wheel-metadata failure from OpenAI Buildkite 5146797: the outer venv lost collision records and dropped `azure.core.tracing.ext.opentelemetry_span` before OpenAPI generation. Ready for independent review.

jbedard:
Nothing should depend on the `py_binary`, instead use `py_library` for shared things or multiple `py_venv_exec()` sharing a single `py_venv()`

tamird:
Agreed. The downstream OpenAPI macro is using the service executable as a dependency only to inherit its import environment; that is the wrong ownership boundary. I will drop this patch from the integration and fix the macro or callers downstream instead. The existing PyInfo forwarding for specialized consumers such as py_pex_binary does not justify extending ordinary deps with the venv wheel closure.

PR #1215, opened 2026-06-28
fix(py_venv): keep one fallback metadata owner

tamird:
A native-root collision can leave one wheel on whole-wheel fallback
while another claimant owns the direct package projection. When both
wheels publish the same dist-info entry, the metadata collision path
rejected the venv even though only one fallback copy remains visible.

Apply the metadata collision policy first. Keep the normal projected
winner when no claimant needs fallback. Let a sole fallback copy own
metadata discovery without another projection. Retain the hard failure
when multiple fallback copies cannot be suppressed.

This preserves one importlib.metadata result for a permissive
duplicate-lock case while keeping error policy and the ambiguous case
explicit.

### DIFF
--- a/py/private/py_venv/venv.bzl
+++ b/py/private/py_venv/venv.bzl
@@ -91,10 +91,10 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
 
     * **Declared distribution metadata in site-packages.** Metadata discovery
       scans every `sys.path` entry instead of stopping at the first match.
-      Preserve the existing exact-entry collision precedence, but project the
-      selected `*.dist-info` and `*.egg-info` entry only when that wheel needs
-      no whole-wheel fallback. Reject an exact-entry collision when a losing
-      claimant remains on fallback and therefore cannot be suppressed.
+      Apply exact-entry collision policy first. When no claimant remains on
+      whole-wheel fallback, project the normal winner. When exactly one
+      remains, project nothing and let that sole fallback copy own discovery.
+      Reject multiple fallback copies because they cannot be suppressed.
 
     * **Console-script name in bin/.** Apply `package_collisions` directly
       — no namespace equivalent.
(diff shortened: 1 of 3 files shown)


tamird:
Extracted from #1205. Downstream exact-head CI found two
pydantic_core-2.41.5.dist-info trees from the Stripe and default lock
universes after a native-root collision left one claimant on whole-wheel
fallback. The focused follow-up keeps exactly one visible metadata owner
instead of making that #1205/#1191 interaction fail analysis.

tamird:
@jbedard This is the focused #1205/#1191 interaction from
downstream exact-head CI. Two same-version pydantic_core wheels from
Stripe and default lock universes leave one fallback metadata copy
visible after a native-root collision. The regression covers no
fallback, one fallback, multiple fallback, and package_collisions =
error. The full matrix is green at e5953046. Ready for independent
review.

jbedard:
add an e2e test case with real packages demonstrating the bug

py/private/py_venv/venv.bzl
```diff
@@ -445,7 +433,20 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
...
+        ]
+        if len(fallback_claimants) > 1:
+            fail(("{}: distribution metadata entry `{}` selects {}, but " +
+                  "multiple claimants remain on whole-wheel fallback: {}.").format(
+                ctx.label,
+                tl,
+                winner,
+                sorted(fallback_claimants),
+            ))
+        if not fallback_claimants:
```
jbedard, line 449: In the single-fallback branch, the sole visible `dist-info` belongs to the fallback wheel, which isn't necessarily the wheel that won the importable-package projection. The `single_fallback` fixture exercises exactly this: metadata for `collision-first` resolves to `_collision_first` (the fallback copy) while the `collision_namespace` code resolves to `_metadata_collision_second` (the projected namespace winner) — so `importlib.metadata` and `import` can report two different wheels.

For genuine duplicate locks (same name+version, identical metadata) this is harmless and clearly the intent. Flagging only to confirm it's a conscious tradeoff for the rarer case of two *distinct* distributions colliding on a `dist-info` name under `package_collisions = "ignore"`, where the metadata a consumer reads would then describe a different build than the code it imports.

^ robot

jbedard:
rebase and add an e2e test case with real packages demonstrating the bug

tamird:
Superseded by #1205 at 1b43f47c. The one-fallback metadata rule can let importlib.metadata report a different wheel from the imported package, so I am not keeping that model. #1205 now suppresses declared losing native roots while retaining fallback for metadata-unknown wheels, which removes duplicate metadata without splitting import and metadata ownership. Its focused collision suite and real jwt plus google/protobuf e2e cases pass uncached.

py/private/py_venv/venv.bzl
tamird: Yes. This is a conscious tradeoff, and the fixture uses different summaries so the mismatch stays visible rather than looking accidentally equivalent.

After permissive collision policy accepts the exact dist-info collision, one whole-wheel fallback means we cannot both leave that wheel on sys.path and project another metadata copy without importlib.metadata returning two distributions. Suppressing only the fallback metadata would need a partial-wheel projection model that this path does not have.

error still rejects the collision, and more than one fallback still fails. For two differing builds with the same normalized name and version under warning or ignore, permissive collision handling can already mix winners across import roots; this chooses one discoverable metadata owner rather than claiming code and metadata are guaranteed to come from one wheel.

PR #1219, opened 2026-06-30
feat(py_image_layer): configurable default layer compression

thirtyseven:
## Motivation

`py_image_layer` always compresses layers with gzip unless a `py_layer_tier` pins a per-group algorithm. There's no way to flip the *default* (e.g. to zstd) globally without editing every tier.

## Change

Adds a `//py:layer_compression` `string_flag` (`gzip`|`zstd`, default `gzip`). It sets the default compression for any layer that doesn't pin its own algorithm via `py_layer_tier.compression`:

```
bazel build //path:layers --@aspect_rules_py//py:layer_compression=zstd
```

- Read through a private `_layer_compression` attr on both layer aspects and the rule, mirroring how `_layer_tier` is plumbed.
- `_compression_for` and `_declare_group_tar` fall back to the flag value and derive the `.tar.gz` / `.tar.zst` extension from it, instead of hard-coding gzip.
- Explicit per-group `py_layer_tier.compression` still wins over the flag.

## Test

Built the `//cases/oci/py_image_layer` e2e case both ways:
- default → `*_default.tar.gz`, `*_squashed.tar.gz`
- `--@aspect_rules_py//py:layer_compression=zstd` → `*_default.tar.zst`, `*_squashed.tar.zst`

Default behavior is unchanged (still gzip), so existing tar-listing snapshots are unaffected.

### DIFF
--- a/py/BUILD.bazel
+++ b/py/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_lib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 
 # Users can set, e.g. --@aspect_rules_py//py:python_version=3.12
 alias(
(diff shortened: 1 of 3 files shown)


py/BUILD.bazel
```diff
@@ -15,6 +16,19 @@ label_flag(
     visibility = ["//visibility:public"],
 )
 
+# Users can set --@aspect_rules_py//py:layer_compression=zstd to change the
+# default compression py_image_layer uses for any layer that doesn't pin its
+# own algorithm via py_layer_tier.compression.
+string_flag(
```
thesayyn, line 22: Shouldn't this be read by layer_tier and set the default compression to this value instead of every py_image_layer deciding that.

PR #1231, opened 2026-07-06
refactor!: use simplified rules_py providers

jbedard:
Remove the direct usage of the `rules_python` `PyInfo` and create our own provider. Merge `PyVirtualInfo` into `PyInfo`

Outputs of `aspect_rules_py` may no longer be consumed by `rules_python`. All public entries into `aspect_rules_pi` such as `deps` of any field still support outputs of `rules_python` (including outputs of `protobuf` which outputs `rules_python` specific providers).

### Test plan

### DIFF
--- a/py/private/BUILD.bazel
+++ b/py/private/BUILD.bazel
@@ -75,6 +75,7 @@ bzl_library(
         ":transitions",
         "@bazel_skylib//lib:new_sets",
         "@bazel_skylib//lib:types",
+        "@rules_python//python:defs_bzl",
     ],
 )
 
--- a/py/private/BUILD.bazel
+++ b/py/private/BUILD.bazel
@@ -48,6 +48,7 @@ bzl_library(
     srcs = ["py_image_layer.bzl"],
     deps = [
         ":py_info",
+        ":py_info_interop",
         "@bazel_lib//lib:transitions",
         "@tar.bzl//tar:mtree",
         "@tar.bzl//tar:tar",
@@ -58,7 +59,7 @@ bzl_library(
     name = "pth",
     srcs = ["pth.bzl"],
     deps = [
-        ":py_info",
+        ":py_info_interop",
         "@bazel_skylib//lib:paths",
     ],
 )
@@ -70,12 +71,12 @@ bzl_library(
         ":providers",
         ":pth",
         ":py_info",
+        ":py_info_interop",
         ":py_semantics",
         ":py_wheel",
         ":transitions",
         "@bazel_skylib//lib:new_sets",
         "@bazel_skylib//lib:types",
-        "@rules_python//python:defs_bzl",
     ],
 )
 
@@ -125,13 +126,22 @@ bzl_library(
     ],
 )
 
+# keep
 bzl_library(
     name = "py_info",
     srcs = ["py_info.bzl"],
     # PyInfo is re-exported as public API from //py:defs.bzl, so its bzl_library
     # is public too, allowing sibling trees (e.g. //uv) to depend on it.
     visibility = ["//visibility:public"],
-    deps = ["@rules_python//python:defs_bzl"],
+)
+
+bzl_library(
+    name = "py_info_interop",
+    srcs = ["py_info_interop.bzl"],
+    deps = [
+        ":py_info",
+        "@rules_python//python:defs_bzl",
+    ],
 )
 
 bzl_library(
--- a/py/private/providers.bzl
+++ b/py/private/providers.bzl
@@ -144,11 +144,3 @@ def make_wheel_record(
         metadata_top_levels = tuple(metadata_top_levels),
         cs_claims = tuple(cs_claims),
     )
-
-PyVirtualInfo = provider(
-    doc = "FIXME",
-    fields = {
-        "dependencies": "Depset of required virtual dependencies, independent of their resolution status",
-        "resolutions": "FIXME",
-    },
-)
--- a/py/private/pth.bzl
+++ b/py/private/pth.bzl
@@ -1,7 +1,7 @@
 """Helper functions for building imports depsets."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
-load("//py/private:py_info.bzl", "PyInfo")
+load("//py/private:py_info_interop.bzl", "get_py_info", "has_py_info")
 
 def _make_import_path(label, workspace, imp):
     if imp.startswith("/"):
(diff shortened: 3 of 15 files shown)


gregmagolan (approved):

ctcjab:
Drive-by feedback from moving a codebase fully onto `python_interpreters` (as the rules_python `python.toolchain()` drop-in), relevant to this detachment work: version *selection* still requires `@rules_python` to be visible.

With the direct `rules_python` dep dropped, setting the interpreter fails, because the flag lives in rules_python:

```
ERROR: @rules_python//python/config_settings:python_version :: Error loading option
@rules_python//python/config_settings:python_version: No repository visible as
'@rules_python' from main repository
```

rules_py's own e2e selects the version the same way (`--@rules_python//python/config_settings:python_version=…`), so there doesn't appear to be a rules_py-native equivalent. Net effect: a consumer that "fully" moves to `python_interpreters` still has to keep a direct `bazel_dep(rules_python)` purely to pick a Python version. If the goal is for the detachment to go unnoticed by such consumers, `python_interpreters` probably wants to re-export (or provide its own) version-selection flag. Not blocking anything for us — just flagging while the surface is in flux.

tamird (commented):
One downstream seam: this centralizes mixed-provider handling in private py_info_interop.bzl, but custom rules that inspect mixed rules_py/rules_python deps while emitting rules_py's PyInfo still need the same boundary. We have a concrete typing-source aggregator that would otherwise have to keep a direct rules_python load or reimplement this private helper. Could has_py_info/get_py_info, and perhaps RulesPythonPyInfo for attr constraints, be re-exported from //py:defs.bzl before this lands?

tamird:
I checked the refreshed provider split against our downstream consumers. Forward interop inside rules_py looks good, but `has_py_info`/`get_py_info` and `RulesPythonPyInfo` are still private. Our typing-source aggregator can receive native rules_python/py-proto deps directly, so switching it to only rules_py `PyInfo` silently drops their transitive sources, imports, and provider-only stubs. Could these interop helpers/provider constraint be exposed from `//py:defs.bzl`? A fresh consumer audit confirms typing is the remaining mixed-provider seam; the tar and rules_py analysis fixtures already intentionally use the rules_py provider.

tamird:
The version-selection gap raised above is now addressed on main: #1309 and #1319 make `--@aspect_rules_py//py:python_version=3.12` select both the interpreter and UV constraints while retaining the rules_python fallback. I am exercising that path in our downstream compatibility migration and have removed the direct rules_python version flag there. The remaining blocker I see for this PR is the mixed-provider accessor surface described in my previous comment.

tamird:
The remaining typing seam needs a small upstream change: re-export the existing RulesPythonPyInfo, get_py_info, and has_py_info from //py:defs.bzl and add //py/private:py_info_interop to the defs bzl_library deps.

RulesPythonPyInfo is needed for both an either-provider attr constraint and direct access to native transitive_pyi_files: get_py_info intentionally prefers the rules_py provider when a target carries both, so reading the native-only stub field through that accessor would silently lose provider-only stubs. The common sources/imports can use get_py_info; downstream will continue forwarding rules_py's virtual fields and wheel provider explicitly.

A small public-API consumer between an existing native py/proto producer and a rules_py test, with providers=[[PyInfo], [RulesPythonPyInfo]], would cover the export and source/import-root propagation without duplicating the existing proto runtime tests. A focused precursor is being prepared.

tamird:
One correction to the version-selection note above: #1309/#1319 cover interpreter and UV wheel constraints, but generic PEP 508 markers still read the rules_python version settings. A non-transitioned consumer can therefore select 3.12 while evaluating a conditional dependency as 3.11. I restored the synchronized downstream flag and am preparing the small upstream marker follow-up with precedence, fallback, and full-version coverage. This is separate from the mixed-provider accessor blocker.

tamird:
Opened #1322 for the public PyInfo interop surface needed by the downstream typing aggregator. It covers native-only, rules_py-only, and dual-provider dependencies so mixed sources, imports, and native transitive stubs remain observable. I also split the generic UV marker version-source fix into #1323.

tamird:
I checked the refreshed head with #1322 integrated. The mixed-provider surface is now sufficient for our typing aggregator: it can retain native/proto sources, imports, and `transitive_pyi_files` while consuming rules_py providers.

One remaining provider load is still private: `PyWheelsInfo` is only available from `//py/private:providers.bzl`. The aggregator needs to forward it so PEP 561/stub-only wheels remain visible to ty and pyright. Could that provider also be re-exported from `//py:defs.bzl`? That would leave the downstream rule with no private rules_py loads.

tamird:
The rebase to d07c377869de8c92ba2448b720aec4219aa267f8 dropped the public provider-interop exports from #1322: RulesPythonPyInfo, get_py_info, and has_py_info are again private-only. The downstream typing aggregator still needs them, along with PyWheelsInfo, to preserve native/proto sources and PEP 561 wheels. I’m refreshing the provider follow-up against the new head so this seam is covered again.

tamird:
The refreshed head is green and approved. I’ll keep the provider-interop and PyWheelsInfo follow-up separate and rebase it onto main once this lands, so there is no stacked PR or branch coupling. Could you land this when ready?

jbedard:
@tamird I want to ensure going this direction will not be roadblock for you, and I was hoping you would never need those helpers.

What scenario is it that you are consuming providers and uncertain if it is rules_py vs legacy rules_python providers? Why can't that scenario be updated to only rules_py?

tamird:
I rechecked the downstream consumers: the mixed-provider case in my fixture was synthetic, so the typing aggregator can consume rules_py PyInfo only and I can drop the interop-helper request.

The source-only typing aggregation still needs to forward PyWheelsInfo so stub-only/PEP 561 wheels remain visible without pulling runtime/native build inputs back into type checks. That provider is the only public export I need, or an equivalent source-only-deps rule. This is not a blocker for landing #1231.

tamird:
Checks are green, this is approved, and nothing downstream blocks landing. Could a maintainer merge it so the independent `PyWheelsInfo` export and downstream typing follow-up can proceed?

tamird (approved):
Reviewed d07c3778. Forward rules_python interop, virtual-dependency propagation, and the downstream typing boundary look correct; the remaining PyWheelsInfo export can stay in the separate follow-up. No blocking findings.

xangcastle (approved):

PR #1233, opened 2026-07-06
fix(uv): add $(EXECROOT) make-var to anchor native sdist env paths across the backend chdir

sallustfire:
`pep517_native_whl` (via `uv.override_package(env = {...}, toolchains = [...])`) expands `$(VAR)` make-variables in the build action's `env` from the toolchains it depends on. Those values are workspace-relative paths: a dependency's include dir expands to something like `bazel-out/.../include`, which resolves from the execroot but not from the PEP 517 backend's cwd. `python -m build` runs the backend with its cwd inside the unpacked sdist worktree (`build_helper` invokes it with `cwd=<worktree>`), so a workspace-relative `-Ibazel-out/.../include` dangles and the build fails with `fatal error: '<header>' file not found`, or the link can't find the dep archive.

Add a `$(EXECROOT)` make-variable to anchor those paths, e.g. `-I$(EXECROOT)/$(DEP_INC)`. The execroot isn't known at analysis time, so `$(EXECROOT)` expands to a marker that `build_helper` replaces with the absolute execroot at build time, before the chdir. It's a plain text substitution rather than flag parsing, so it works for any env var and any flag form (`-I`, `-L`, `-Wl,-rpath`, `-framework`, `CC`, ...).

`EXECROOT` is reserved for this. If a toolchain in `toolchains` already exports it, `$(EXECROOT)` would expand to that value instead of the marker and silently skip anchoring, so analysis now fails with a message pointing at the collision rather than producing a confusing compile error later.

I went with an opt-in anchor rather than having `build_helper` detect and absolutize path-bearing flags itself: it can't reliably tell an execroot-relative path from a symbol/framework name or a worktree-relative path, can't cover every flag syntax, and would risk mangling values it doesn't need to touch.

More generally, what might be missing is some notion of a `deps` analog to `cargo_build_script` that can inspect a CcInfo provider and get the `CcInfo.linking_context`. This is a much bigger design question if were to be offered first-class since setuptools handles the linking, and this would all have to get flattened into LDFLAGS/extra_link_args. 

## Regression test

Adds `uv/private/pep517_whl/tests/native_dep`: a `cc_library` whose include dir and static archive are surfaced as make-vars (standing in for `override_package(toolchains=, env=)`), consumed by a small PEP 517 backend that compiles a C extension against them.

Split into a red/green pair. The test commit wires the env as plain `-I$(DEP_INC)` / `$(DEP_LIB_A)`, which fails today because the backend compiles with an execroot-relative `-I` that dangles after the chdir. The fix commit adds `$(EXECROOT)` and switches the test to `-I$(EXECROOT)/$(DEP_INC)`.

- Searched for relevant documentation and updated as needed: yes (`$(EXECROOT)` documented on the `override_package(env=)` tag and the `pep517_native_whl` `env` attr)

> `uv.override_package(env = {...})` now supports a `$(EXECROOT)` make-variable to anchor workspace-relative paths for native sdist builds (e.g. `"CPPFLAGS": "-I$(EXECROOT)/$(DEP_INC)"`), so they survive the PEP 517 backend's chdir into the sdist worktree.

## Testing

- `bazel test //uv/private/pep517_whl/tests/native_dep:native_dep_env_test`
- `buildifier --mode=check uv/private/pep517_whl/rule.bzl`


xangcastle:
Hi, could you resolve the conflicts?

tamird:
This is now superseded by #1328, which carries the EXECROOT anchor plus configured compiler/archiver/linker selection and the restored native-dependency regression on the current rules_py base. #1233 is conflicted and the two changes should not both land.

jbedard:
I believe #1328 solved this?

sallustfire:
@tamird Thanks for pulling the relevant pieces in.

Would you be open to entertaining a PR that builds on the relevant pieces of #1328 and extracts the linking context from a CcInfo automatically and passes it to setuptools?

tamird:
Yes. A follow-up that derives the include and link inputs from CcInfo and passes them through the native-sdist path would be welcome. Please keep the setuptools-specific behavior scoped to that backend, preserve the configured C++ driver/EXECROOT handling from #1328 and #1342, and include an executable extension regression that proves both the header and linked-library paths work after the backend changes directory.

PR #1234, opened 2026-07-06
fix(py): exclude Tcl/Tk runtime libraries when exclude_feature=tkinter is set

### LINKED ISSUE #1203: [Bug]: v2 exclude_feature=tkinter does not remove Tcl/Tk shared libraries from python-build-standalone interpreters
### What happened?

`--@aspect_rules_py//py/private/interpreter:exclude_feature=tkinter`
does not remove all Tcl/Tk-related files from recent python-build-standalone
interpreter layouts.

I reproduced this with Python 3.13 x86_64 linux gnu using the
`install_only_stripped` interpreter variant. After building an application image
with:

    --@aspect_rules_py//py/private/interpreter:exclude_feature=tkinter

the image no longer contains the Python `tkinter` package or `_tkinter` extension,
but it still contains Tcl/Tk shared libraries and support directories from the
interpreter repository, for example:

    lib/libtcl9.0.so
    lib/libtcl9tk9.0.so
    lib/thread3.0.4/libtcl9thread3.0.4.so
    lib/itcl4.3.5/libtcl9itcl4.3.5.so

This does not appear to be specific to `install_only_stripped`. According to the
python-build-standalone documentation, `install_only_stripped` is equivalent to
`install_only`, but with debug symbols removed. Since the remaining Tcl/Tk files
are normal runtime shared libraries and support files, the same layout issue is
likely to affect both `install_only` and `install_only_stripped`.

The current `tkinter` feature definition only matches:

    lib/python{major}.{minor}/tkinter/**
    lib/python{major}.{minor}/lib-dynload/_tkinter*

That removes the Python wrapper package and the `_tkinter` extension module, but
does not remove the Tcl/Tk runtime libraries that are shipped separately by newer
python-build-standalone releases.

Recent python-build-standalone releases split `_tkinter`, `libtcl`, and `libtk`
into separate dynamic libraries. As a result, excluding the `tkinter` feature
probably also needs to exclude the Tcl/Tk shared libraries and support
directories from the interpreter `lib` directory.

Suggested additional patterns for the `tkinter` feature:

    lib/libtcl*
    lib/libtk*
    lib/tcl*/**
    lib/tk*/**
    lib/thread*/**
    lib/itcl*/**

Expected behavior:

When `exclude_feature=tkinter` is enabled, all files required only for Tk/Tcl GUI
support should be excluded from the generated interpreter repository and from
application runfiles/images.

Actual behavior:

The Python `tkinter` package and `_tkinter` extension are excluded, but Tcl/Tk
runtime libraries remain in the output image.

Why this matters:

For server-side Python applications and minimal OCI images, Tk/Tcl GUI support is
not needed. Leaving these libraries in the interpreter layer increases the final
artifact size even though `tkinter` was explicitly excluded.

### Version

Development (host) and target OS/architectures: linux

Output of `bazel --version`: any

`WORKSPACE` or `MODULE.bazel` file: 2.0.0-alpha.3

### How to reproduce

```shell

```

### Any other information?

_No response_

xangcastle:
Recent python-build-standalone releases ship Tcl/Tk as separate runtime shared libraries and support directories, so excluding only the Python tkinter package and _tkinter extension module leaves libtcl*, libtk*, tcl*/, tk*/, itcl*/, and thread*/ behind in the interpreter image. This   change extends the tkinter feature definition in `py/private/interpreter/exclude_feature.bzl` to also remove those files, adds an  e2e/cases/interpreter-features-tkinter regression case that pins the aspect_rules_py interpreter and verifies the resulting OCI layer contents and regenerates the layer snapshot so that `--exclude_feature=tkinter` once again produces a Tcl/Tk-free output.
fix #1203 

---

### Test plan

- New test cases added


e2e/cases/interpreter-features-tkinter/BUILD.bazel
```diff
@@ -0,0 +1,78 @@
...
+load("@bazel_features//:features.bzl", "bazel_features")
+load("@bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
+load("@container_structure_test//:defs.bzl", "container_structure_test")
+load("@rules_oci//oci:defs.bzl", "oci_image")
+load("//tools:asserts.bzl", "assert_tar_listing")
+
+# See cases/oci/py_venv_image_layer/BUILD.bazel for the rationale: Bazel 9
+# stages canonical repo names so `runfiles.bash` bootstrap can't locate
+# `bazel_tools/tools/bash/runfiles/runfiles.bash` by apparent name.
+_RUNS_BINARY_INCOMPATIBLE_WITH_BAZEL9 = ["@platforms//:incompatible"] if bazel_features.rules.merkle_cache_v2 else []
```
jbedard: Can we just use the `skip-on-bazel9` tag instead?

jbedard (approved):

PR #1235, opened 2026-07-06
fix(py_image_layer): wire PY_TOOLCHAIN through py_venv_exec so interpreter layers are emitted

### LINKED ISSUE #1217: [Bug]: v2 `interpreter_group` doesn't work as needed
### What happened?

According to this https://github.com/aspect-build/rules_py/blob/22925c2f9532acfc759575262f0ee031ae30b75f/py/private/py_image_layer.bzl#L98 when `interpreter_group` is not empty, it should be extracted in dedicated layer. But this doesn't happen.

Bug is reproduced by existing e2e tests.

`my_app_layers_fp` uses layer `my_app_tier` with `interpreter_group = "interpreter"` (https://github.com/aspect-build/rules_py/blob/22925c2f9532acfc759575262f0ee031ae30b75f/e2e/cases/oci/py_image_layer/BUILD.bazel#L94)

However, according to test results, python interpreter in the 3rd (default) layer: https://github.com/aspect-build/rules_py/blob/22925c2f9532acfc759575262f0ee031ae30b75f/e2e/cases/oci/py_image_layer/snapshots/my_app_layers_fp_listing.yaml#L55

### Version

Development (host) and target OS/architectures: linux

Output of `bazel --version`: doesn't matter

`WORKSPACE` or `MODULE.bazel` file: 2.0.0-alpha.3

### How to reproduce

```shell
Bug is reproduced by existing e2e tests.
```

### Any other information?

_No response_

xangcastle:
`py_image_layer` builds the interpreter tar from the Python toolchain aspect, but it reads that tar back from ctx.rule.toolchains on the binary target. py_venv_exec is the rule that the py_binary macro ultimately attaches the layer aspect to, yet it only declared the launcher toolchains, not `PY_TOOLCHAIN`. Because the toolchain was invisible to the aspect, the interpreter was silently dropped into the default source layer instead of a dedicated layer. Adding `PY_TOOLCHAIN` to `py_venv_exec` and `py_venv_exec_test` exposes the toolchain info so the aspect can route interpreter files to their own layer as intended.

Fix #1217 

### Test plan

- New test cases added

### DIFF
--- a/e2e/cases/oci/py_image_layer/snapshots/my_app_layers_fp_listing.yaml
+++ b/e2e/cases/oci/py_image_layer/snapshots/my_app_layers_fp_listing.yaml
@@ -6,40 +6,6 @@ files:
 ---
 layer: 1
 files:
-  - -rwxr-xr-x  0 0      0         266 Jan  1  2023 ./app.runfiles/aspect_rules_py++uv+whl_install__images__colorama__0_4_6/actual_install.install/lib/python3.11/site-packages/colorama/__init__.py
-  - -rwxr-xr-x  0 0      0        2522 Jan  1  2023 ./app.runfiles/aspect_rules_py++uv+whl_install__images__colorama__0_4_6/actual_install.install/lib/python3.11/site-packages/colorama/ansi.py
-  - -rwxr-xr-x  0 0      0       11128 Jan  1  2023 ./app.runfiles/aspect_rules_py++uv+whl_install__images__colorama__0_4_6/actual_install.install/lib/python3.11/site-packages/colorama/ansitowin32.py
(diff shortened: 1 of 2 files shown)


py/private/py_venv/py_venv_exec.bzl
```diff
@@ -246,13 +247,13 @@ py_venv_exec = rule(
     implementation = _py_venv_exec_impl,
     attrs = _attrs,
     executable = True,
-    toolchains = [launcher.finalizer_toolchain_type, launcher.template_toolchain_type],
+    toolchains = [PY_TOOLCHAIN, launcher.finalizer_toolchain_type, launcher.template_toolchain_type],
```
thesayyn: while this is an okay fix, there will be some correctness issues assuming toolchain we received here was the one that got propagated. i think the real fix is to have the aspect propagate to the py_venv and pick the toolchain there.

thesayyn: have the aspect propagate up the relation between py_venv_exec and whatever creates the venv.

thesayyn (approved):

PR #1246, opened 2026-07-07
refactor(uv): make SCC computation linear and single-pass

jbedard:
### Test plan


jbedard:
This most likely isn't worth doing if we land https://github.com/aspect-build/rules_py/pull/1253

acozzette (approved):

PR #1253, opened 2026-07-07
perf(uv): derive SCC ids without hashing full SCC content

jbedard:
collect_sccs() computed each SCC's id as a pure-Starlark SHA-1 over the
repr of its members plus all external deps and markers — often
multi-KB per package, ~94% of the function's CPU (830ms -> 70ms on a
500-package graph).

A lone base package (the common case; dependency graphs are DAGs) now
gets the readable `<name>__<version>`; a real cycle hashes only its
member list. An id_state dict shared across a project's dependency
groups interns ids by full content, so identical SCCs keep one id and
SCCs differing only in deps/markers stay distinct — the property the
content hash previously provided for the cross-group aggregation.

SCC ids only name targets inside the generated project repos, so the
new spelling (visible in the regenerated snapshots) is not a breaking
change — and makes the generated BUILD files debuggable.

### Test plan

### DIFF
--- a/e2e/cases/snapshots/project_single.BUILD.bazel
+++ b/e2e/cases/snapshots/project_single.BUILD.bazel
@@ -11,7 +11,7 @@ filegroup(
 # cowsay
 # {
 #     "single_project_hub": {
-#         "aee49c8f5bedac06": {
+#         "cowsay__6_1": {
 #             "": 1,
 #         },
 #     },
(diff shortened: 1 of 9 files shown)


xangcastle:
Please wait on merging this one after I work in the analysis time benchmark

xangcastle (approved):

PR #1257, opened 2026-07-07
fix: resolve py_image_layer.platform relative to the calling repo

jbedard:
### Test plan

- New test cases added


py/private/py_image_layer.bzl
```diff
@@ -1041,6 +1041,11 @@ def py_image_layer(
     """
     tags = kwargs.pop("tags", []) + ["manual"]
 
+    # The transition consumes the platform as a raw string, which Bazel parses
+    # relative to aspect_rules_py — canonicalize against the calling package.
+    if platform:
```
jbedard: Shouldn't this just be a real label attribute instead of string?

thesayyn (approved):

PR #1265, opened 2026-07-07
refactor(uv): alias cp* python-version constraints to their py* twins

jbedard:
The generated settings check only the interpreter version, so cp312
and py312 were identical config_settings (each with a private flag
rule behind it). Generate real settings for the generic py prefix and
aliases for the rest. Also make is_python_version_at_least's version/
visibility params mandatory: the version-defaults-to-name fallback was
unreachable, and the visibility default was the .bzl builtin function.

### Test plan

### DIFF
--- a/uv/private/constraints/python/macro.bzl
+++ b/uv/private/constraints/python/macro.bzl
@@ -9,7 +9,6 @@ load(":defs.bzl", "is_python_version_at_least")
 # buildifier: disable=function-docstring
 def generate(
         visibility):
-    # FIXME: Needs to generate a cascade.
     for major in MAJORS:
         is_python_version_at_least(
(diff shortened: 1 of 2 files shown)


uv/private/constraints/python/macro.bzl
```diff
@@ -10,17 +10,35 @@ load(":defs.bzl", "is_python_version_at_least")
 def generate(
         visibility):
     # FIXME: Needs to generate a cascade.
```
acozzette: Can we remove this FIXME? I'm not even sure what it's referring to by "cascade" but the macro seems to be generating everything it needs to at this point, right?

acozzette (approved):

PR #1267, opened 2026-07-07
chore: config hygiene — drop no-op flags, dead ci.bazelrc stanzas, tracked generated/empty files

jbedard:
- --incompatible_enable_cc_toolchain_resolution is default-true since Bazel 7; the repo runs 8/9 only
- ci.bazelrc: its only consumer is startup-benchmark.yml's build commands (which pass --disk_cache=), so the test stanzas, build:release, and disk_cache lines never take effect
- benchmark/startup/MODULE.bazel is generated with a machine-local local_path_override and regenerated by CI; untrack + gitignore it
- .prettierignore was a tracked empty file
- fix stale comments in e2e/cases/.bazelrc

### Test plan


.gitignore
```diff
@@ -12,6 +12,10 @@
 .venv/
 node_modules/
 
+# Generated by benchmark/startup/generate_module.py with a machine-local
+# local_path_override; regenerated by CI before every use.
+benchmark/startup/MODULE.bazel
```
jbedard, line 17: @xangcastle is it correct that this is generated and should not be checked in?

acozzette (approved):

.gitignore
xangcastle: yes is correct

xangcastle (approved):

PR #1268, opened 2026-07-07
fix(venv): treat semantically equivalent namespace __init__.py stubs as non-colliding

xangcastle:
When multiple wheels contribute to the same namespace package, their __init__.py files often contain equivalent pkgutil/pkg_resources namespace stubs that differ only in comments, whitespace, or quote style. The previous byte-for-byte and SHA-256 comparisons treated these as collisions. This change introduces semantic equivalence     
   detection for __init__.py namespace stubs and applies it in both the dynamic symlink path (pth.rs) and the static venv population path (venv.rs), so equivalent namespace package declarations no longer fail the build under strict collision policies.

---

### Test plan

- New test cases added


py/tools/py/src/namespace.rs
```diff
@@ -0,0 +1,238 @@
...
+/// Files at the same path are trivially equivalent. If both files are
+/// `__init__.py` namespace package stubs, they are treated as equivalent even
+/// when their bytes differ only in comments, whitespace, or quote style.
+/// Otherwise the files are compared byte-for-byte.
+pub fn are_files_equivalent(p1: &Path, p2: &Path) -> Result<bool> {
+    if p1 == p2 {
+        return Ok(true);
+    }
+
+    if is_namespace_init(p1)? && is_namespace_init(p2)? {
```
jbedard: Should we avoid opening + reading the files multiple times? Currently `is_namespace_init` and `files_are_byte_equal` may both open+read the files?

Sounds expensive...

py/tools/py/src/namespace.rs
```diff
@@ -0,0 +1,249 @@
...
+    let content1 = std::fs::read_to_string(p1).unwrap_or_default();
+    let content2 = std::fs::read_to_string(p2).unwrap_or_default();
+
+    if is_namespace_init_content(p1.file_name(), &content1)
+        && is_namespace_init_content(p2.file_name(), &content2)
+    {
+        return Ok(true);
+    }
+
+    files_are_byte_equal(p1, p2)
```
jbedard, line 118: This still reads the files a second time

PR #1269, opened 2026-07-07
fix(pytest): expose main() from py_pytest_main-generated entrypoints

jbedard:
By de-duplicating `py/private/pytest.py.tmpl` and `py/private/pytest_main.py` which were already diverging (the added test case).

### Test plan

- New test cases added


e2e/cases/pytest-main-867/BUILD.bazel
```diff
@@ -44,6 +44,29 @@ py_test(
...
+    name = "wrap_main",
+    tags = ["manual"],
+)
+
+py_test(
+    name = "test_wrapped_main",
+    srcs = [
+        "test_wrapped.py",
+        "wrapped_main.py",
+        ":wrap_main",
```
acozzette: Does `:wrap_main` need to be in both srcs and deps?

acozzette (approved):

e2e/cases/pytest-main-867/BUILD.bazel
jbedard: Removed and still reproduces the original bug being fixed

PR #1270, opened 2026-07-07
fix: pkgutil namespace package __init__.py collisions when multiple wheels share a namespace stub

xangcastle:
This PR fixes false-positive package collisions for namespace packages like backports where each wheel ships an equivalent  `__init__.py` that only declares a `pkgutil` or `pkg_resources` namespace. The wheel metadata extractor now detects these stubs, marks the top-level as a namespace package, and the venv assembly merges the entries with individual symlinks instead of treating it as a collision. site_merge.py also learns to treat equivalent namespace stubs as the same file during physical merges as a defensive fallback.

---

### Test plan

- New test cases added


uv/private/whl_install/repository.bzl
```diff
@@ -762,6 +822,11 @@ filegroup(
     # Only emitted for wheels that contribute to a namespace — keeps the
     # generated BUILD files (and their e2e snapshots) unchanged for the
     # common regular-top-level-only case.
+    if pkgutil_namespace_top_levels_by_whl:
```
jbedard, line 825: Seems like we have no snapshot tests covering this case?

jbedard:
Do we want to merge #1205 first to avoid waiting for that bot to rebase?

uv/private/whl_install/repository.bzl
xangcastle: adding a new one

py/tools/site_merge/site_merge.py
```diff
@@ -32,6 +33,195 @@
...
+
+def _are_files_equivalent(src_file, dest):
+    """Compare two files, treating namespace stubs as equivalent.
+
+    Returns True if the files are byte-identical or if both are
+    __init__.py files containing only a recognized namespace package stub.
+    """
+    if filecmp.cmp(str(src_file), str(dest), shallow=False):
+        return True
+    if src_file.name != "__init__.py" or dest.name != "__init__.py":
```
jbedard, line 215: If both are `__init__.py` files should we only do the `_is_namespace_init_stub()` calls and not `filecmp.cmp()`?

Today (just like in the rust version) we are reading the file twice, so 2x the IO operations, when both files are `__init__.py`. I think when analyzing the AST there's no point doing it twice?

py/tools/site_merge/site_merge.py
```diff
@@ -32,6 +33,195 @@
...
+        args[0].id == "__name__"
+    )
+
+
+def _is_namespace_init_stub(content):
+    """Return True if the __init__.py content is a namespace package stub."""
+    try:
+        tree = ast.parse(content)
+    except SyntaxError:
+        return False
```
jbedard, line 203: What if the file was empty? Do we have a test for that?

py/tools/site_merge/site_merge.py
```diff
@@ -32,6 +33,195 @@
...
+            continue
+        # Track the canonical `from pkgutil import extend_path` import, then
+        # expect the assignment as the only other meaningful statement.
+        if isinstance(stmt, ast.ImportFrom) and stmt.module == "pkgutil":
+            if len(stmt.names) == 1 and stmt.names[0].name == "extend_path":
+                import_seen = True
+                continue
+        meaningful.append(stmt)
+
+    if len(meaningful) != 1:
```
jbedard, line 64: Is an empty body not an "init ast"?

jbedard:
Robot feedback, possibly all 7 cases should have unit tests and be fixed?

```
Bugs / uncovered stub forms (false negatives → spurious collision errors)

1. The try/except form — the function's own documented example fails. Line 90 requires stmt.handlers[0].type is None, i.e. a bare except:. The canonical setuptools fallback stub uses except ImportError: (handler type is ast.Name), so the entire ast.Try branch rejects the exact code shown in its own comment at lines 297–301 of the diff. This is the standard pkg_resources-with-pkgutil-fallback template (shipped verbatim in protobuf's google/__init__.py lineage — confirmed the pattern in protobuf-3.0.0's wheel):

try:
    __import__('pkg_resources').declare_namespace(__name__)
except ImportError:
    __path__ = __import__('pkgutil').extend_path(__path__, __name__)

2. Plain import pkgutil / import pkg_resources statements are never filtered out. The pre-filter only drops from pkgutil import extend_path, so these common two-liners leave len(meaningful) == 2 and fail:

import pkgutil
__path__ = pkgutil.extend_path(__path__, __name__)
Consequence: "pkgutil form 2 (module attribute)" in _is_pkgutil_extend_path_call is dead code — it can never be reached, because the import pkgutil that makes pkgutil.extend_path meaningful always fails the count check first. Same for import pkg_resources + pkg_resources.declare_namespace(__name__).

3. from pkg_resources import declare_namespace is unhandled twice over. The import filter only recognizes module == "pkgutil", and _is_pkg_resources_declare_namespace_call has no bare-Name form (asymmetric with extend_path, which accepts extend_path(...) as a plain name):

from pkg_resources import declare_namespace
declare_namespace(__name__)

4. Multi-statement try/handler bodies — the zope/paste/PasteDeploy-style stub with imports inside the branches fails the len(...) == 1 checks:

try:
    import pkg_resources
    pkg_resources.declare_namespace(__name__)
except ImportError:
    from pkgutil import extend_path
    __path__ = extend_path(__path__, __name__)

5. Aliased import: from pkgutil import extend_path as ep is dropped by the filter (name matches), but __path__ = ep(...) isn't recognized. Related: import_seen (lines 48/60) is assigned
and never read — dead variable, and it means a pkgutil import paired with a pkg" the filter step.

Crashes (worse than false negatives)

6. AttributeError on bare try/except with non-expression statements. Lines 95–9 handlers[0].body[0].value without checking node type. A colliding __init__.pylike a compat shim:

try:
    import simplejson
except:
    pass
raises AttributeError: 'Import' object has no attribute 'value' — confirmed. _is_namespace_init_stub (line 202) only catches SyntaxError, and merge() has no guard, so the whole site_merge action dies with a traceback instead of reporting the collision.

7. UnicodeDecodeError on non-UTF8 files (in the wrapper _are_files_equivalent, a __init__.py with a # -*- coding: latin-1 -*- cookie and non-ASCII commentcrashes read_text() — confirmed. Should be read_text(errors=...) or bytes-compare-first (byte-identical is already handled, but differing non-UTF8 files crash).
```

jbedard:
What does native uv/pip/python do though? Can we include references to those docs or implementations?

PR #1275, opened 2026-07-07
fix(uv): resolve extra-only markers during graph activation and treat them as venv-resolved in selects

xangcastle:
This fix handles PEP 508 markers that reference the extra variable by recognizing they are not Bazel execution-environment conditions: during dependency-graph activation, real extra markers such as `extra == 'build'` are partially evaluated against the active extra so they collapse to unconditional or platform-only expressions, while any marker that depends solely on extra—including uv's internal conflict-routing markers—is treated as already resolved by the active dependency-group/venv and rendered as a //conditions:default arm in the generated aliases instead of an un-evaluable decide_marker select, preventing analysis failures when the venv name does not match the extra.

---

### Test plan

- New test cases added

### DIFF
--- a/uv/private/extension/test_marker_simplify.bzl
+++ b/uv/private/extension/test_marker_simplify.bzl
@@ -1,0 +1,98 @@
+"""Tests for marker_simplify.bzl."""
+
+load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
+load(":marker_simplify.bzl", "is_extra_only_marker", "simplify_extra_marker", "simplify_markers_for_extras")
+
+def _simplify_extra_marker_test_impl(ctx):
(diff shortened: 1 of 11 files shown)


jbedard:
Same for `main`?

uv/private/extension/test_marker_simplify.bzl
```diff
@@ -0,0 +1,98 @@
+"""Tests for marker_simplify.bzl."""
```
jbedard, line 1: I added one commit with some claude recommendations, including this file. Feel free to edit or revert any of those changes

jbedard (approved):

xangcastle:
yes, same for main https://github.com/aspect-build/rules_py/pull/1279

PR #1279, opened 2026-07-08
fix: uv 'extra-marker' markers in dependency groups

xangcastle:
### Test plan

- New test cases added


jbedard:
@tamird does this effect you?

tamird:
Not currently. Our active uv locks include extras selected from dependency groups, but their resolved optional-dependency edges do not retain `extra == ...` markers. This does not block our rules_py 2 migration; the fixture covers a valid lock shape that could affect us if uv starts retaining those markers.

— tamirdex

jbedard (approved):

PR #1283, opened 2026-07-08
refactor: simplify _resolve_wheel_collisions

acozzette:
Behavior-preserving cleanup (all venv actions are bit-identical on the existing test fixtures):

* Compute duplicate-metadata losers with a _distinct_claimants comprehension per metadata entry, replacing the manual winner/seen loop.
* Add _under/_shallowest helpers, replacing five hand-rolled path-prefix / minimal-root loops.
* Merge ns_covered_per_wheel into covered_per_wheel; pass 3 consumed them identically.
* Derive the distinct-claimant chain from distinct_sp directly: dict keys already preserve first-claim order with last-claim values.
* Drop the console-script singleton special case; _complain_chain is a no-op on a single claimant.
* Replace the native-winner scan and pass 2a's manual dedup with comprehensions over precomputed distinct orders.

Also add a regression test pinning the per-entry scope of the duplicate-metadata dedup: a distinct-metadata native claimant that loses to a regular claimant must keep its .pth fallback even when a third wheel follows it in metadata-claimant order.

---

### Test plan

- New test cases added

### DIFF
--- a/py/private/py_venv/venv.bzl
+++ b/py/private/py_venv/venv.bzl
@@ -155,6 +155,25 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
         for i in range(1, len(distinct)):
             _complain(what, name, distinct[i - 1], distinct[i])
 
+    def _under(path, roots):
+        """True when path equals or sits below any of roots."""
+        for root in roots:
+            if path == root or path.startswith(root + "/"):
+                return True
+        return False
+
+    def _shallowest(roots):
+        """Minimal cover of roots: drop any root nested below another.
+
+        Lexicographic order puts an ancestor before its descendants, so a
+        single pass over the sorted roots suffices.
+        """
+        out = []
+        for root in sorted(roots):
+            if not _under(root, out):
+                out.append(root)
+        return out
+
     # Pass 1: bucket claimants per import root, distribution metadata entry,
     # and console-script name. The per-wheel claim structs are precomputed
     # by make_wheel_record, so this pass only merges them per closure.
@@ -178,17 +197,13 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
     # entry makes every losing fallback unsound because metadata discovery
     # scans every sys.path entry. Cover only prior duplicate-metadata
     # claimants; the selected claimant still keeps fallback when needed.
-    duplicate_metadata_loser_sps = {}
-    for claimants in metadata_claimants.values():
-        winner = None
-        seen = {}
-        for site_packages in claimants:
-            if site_packages in seen:
-                continue
-            if winner != None:
-                duplicate_metadata_loser_sps[winner] = True
-            winner = site_packages
-            seen[site_packages] = True
+    # Losers are computed per entry: wheels that never share a declared
+    # entry are not duplicate-metadata losers.
+    duplicate_metadata_loser_sps = {
+        loser: True
+        for claimants in metadata_claimants.values()
+        for loser in _distinct_claimants(claimants)[:-1]
+    }
 
     # Pass 2: resolve top-levels. Track which (site_packages, tl) pairs
     # we SKIPPED (left to the .pth fallback) and which claims were fully
@@ -196,7 +211,6 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
     # pass 3 can decide which wheels are fully covered.
     top_level_to_site_pkgs = {}
     skipped_per_wheel = {}
-    ns_covered_per_wheel = {}
     covered_per_wheel = {}
     merge_groups = []
     conflicted_roots = {}  # root path -> True (regular package spanning wheels)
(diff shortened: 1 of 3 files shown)


jbedard:
The analysis benchmark currently says +2%, lets watch that as we retrigger (with fixups or just ammend+push) it to see if that is real...

py/private/py_venv/venv.bzl
```diff
@@ -238,45 +250,26 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
...
-                        under_native_root = True
-                        break
-                if not under_native_root:
-                    native_conflicted_roots[root] = True
-
-            mergeable_conflicted_roots = {}
+            native_conflicted_roots = _shallowest([
+                root
+                for root in tl_conflicted_roots
+                if any([_under(nc, [root]) for nc in native_candidates])
```
jbedard: Robot feedback1 (covering this line and one more below?):

```
Kill the _under(x, [root]) single-element-list idiom with a converse helper.
_under asks "is path below any of these roots?", but two call sites use it backwards — wrapping a single root in a list to ask "does anything sit below this root?":

if any([_under(nc, [root]) for nc in native_candidates])      # line 267
any([_under(entry, [root]) for entry in c.ns_entries])        # line 313

A reader has to mentally unwrap any + list-literal + swapped roles. A converse helper makes both sites read as intended:

def _covers(root, paths):
    """True when any of paths equals or sits below root."""
    for p in paths:
        if p == root or p.startswith(root + "/"):
            return True
    return False

Then: if _covers(root, native_candidates) and ... or _covers(root, c.ns_entries). It also drops the per-call list allocations Starlark's any() forces.
```

py/private/py_venv/venv.bzl
jbedard: I'm uncertain of this robot feedback, so feel free to ignore:

```
Name the distinct_sp dict trick once, in a helper.
The invariant "dict keyed by sp = precedence chain in key order, last claim struct in values" is now load-bearing in two places — the top-level loop (line 217, with the 3-line comment at 425–427) and pass 2b (line 492, uncommented). A tiny helper carries the invariant in one docstring and makes both sites self-describing:

def _distinct_by_sp(claimants):
    """Last claim struct per claimant, keyed in first-claim (precedence) order."""
    return {c.site_packages: c for c in claimants}

That deletes the comment at 425 and documents the console-script site for free. It also gives distinct_sp.values()[-1] (line 494) a nicer home — _distinct_by_sp(claimants).values()[-1] still reads oddly to Python-trained eyes (in Python, dict.values() isn't indexable); naming it winner = ... with the helper makes the Starlark-ism less jarring.
```

py/private/py_venv/venv.bzl
```diff
@@ -238,45 +250,26 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
 
             unique_claimants = distinct_sp.values()
 
-            native_candidates = {}
-            for root in tl_conflicted_roots:
-                for c in unique_claimants:
-                    if root in wheel_by_sp[c.site_packages].native_roots:
-                        native_candidates[root] = True
-                        break
+            native_candidates = {
```
jbedard: Robot feedback:
```
Reshape native_candidates as a filter, not a cross-product.
The dict comprehension at 253–258 is a for root / for c / if cross-product where the dict silently dedups. The intent is "which conflicted roots are native for some claimant" — a filter says that directly:

native_candidates = [
    root
    for root in tl_conflicted_roots
    if any([root in wheel_by_sp[c.site_packages].native_roots for c in unique_claimants])
]

Same cost, and it stops the reader wondering whether duplicate keys matter. (Downstream only iterates it, so list vs. dict is free — same as the PR already did for native_conflicted_roots.)
```

py/private/py_venv/venv.bzl
```diff
@@ -296,34 +289,29 @@ def _resolve_wheel_collisions(ctx, wheels, package_collisions):
...
+                    regulars = [
+                        c
+                        for c in unique_claimants
+                        if root in wheel_by_sp[c.site_packages].regular_roots
+                    ]
+                    if not regulars:
                         fail("{}: native conflicted root {} has no regular claimant.".format(ctx.label, root))
-                    top_level_to_site_pkgs[root] = winner.site_packages
-                    native_winner_by_root[root] = winner.site_packages
+                    top_level_to_site_pkgs[root] = regulars[-1].site_packages
```
jbedard: Robot feedback:
```
Bind the winner once in native_winner_by_root (lines 304–305):

winner_sp = regulars[-1].site_packages
top_level_to_site_pkgs[root] = winner_sp
native_winner_by_root[root] = winner_sp

Two regulars[-1] reads make you check they're the same index.
```

jbedard:
If you feel like more cleanup/refactors, more robot feedback:
```
out of scope for this PR: the all_namespace arm is now ~120 lines nested four deep inside the tl loop with two continue-terminated sub-branches (regular-span vs. pure PEP 420). Extracting it into a local _resolve_all_namespace(tl, distinct_sp) function would flatten pass 2 into a readable dispatch. That's a structural change deserving its own diff, though — worth suggesting as a follow-up rather than blocking this one.
```

py/private/py_venv/venv.bzl
acozzette: Agreed — added `_covers` as the converse of `_under` in 7e3b9ab; both call sites now read in the intended direction and drop the per-call list allocations.

py/private/py_venv/venv.bzl
acozzette: Took it — added `_distinct_by_sp` in 7e3b9ab. The ordering invariant (insertion order preserved, overwrite keeps key position) now lives in one docstring and the inline comment in the top-level loop is gone.

py/private/py_venv/venv.bzl
acozzette: Agreed — reshaped as a filter over the conflicted roots in 7e3b9ab.

py/private/py_venv/venv.bzl
acozzette: Done in 7e3b9ab — bound once as `winner_sp`.

jbedard:
The `Bazel analysis benchmark` currently shows this hurting performance. It was +2% before, after addressing my PR comments it's +5% - so it might not just be a flake and needs more investigation...

acozzette:
I now think the benchmark result was probably just a flake, because I haven't been able to reproduce any major differences locally, and after re-triggering the benchmark the results for main and this PR are within a standard deviation of each other.

jbedard (approved):

jbedard:
Thanks for verifying 👍

PR #1284, opened 2026-07-09
fix: Don't remove `project.dependencies` when deps groups are present

konsti-openai:
Currently, `project.dependencies` are removed from consideration where the first dep group is added. This shouldn't happen, as dependency groups are additive, and `project.dependencies` are intended as the main production dependencies.

---

### Test plan

- New test cases added

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -103,9 +103,14 @@ uv.override_package(
 use_repo(uv, "pypi")
 ```
 
-We can configure a default dependency group by setting the `dep_group` flag on our hub as part of the `.bazelrc`.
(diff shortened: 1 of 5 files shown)


xangcastle:
This looks valuable, but I would like to see and e2e that demostrate the feature

EdNutting:
+1 Also facing this problem. The reproduction is trivial: setup a package that has both base dependencies and a dependency group, and add any script/tool that relies on dependencies from both the group and the base. At the moment, invoking the script/tool will fail because the base dependencies will (incorrectly) not be installed.

PR #1288, opened 2026-07-10
refactor: remove use of rules_python internals

jbedard:
### Test plan


uv/private/uv_hub/repository.bzl
```diff
@@ -255,8 +253,14 @@ all_whl_requirements = all_whl_requirements_by_package.values()
 
 all_data_requirements = all_requirements
 
+def _normalize_name(name):
```
acozzette: We have this same function in uv/private/normalize_name.bzl. Is there any reason we can't just use that one so that we can avoid generating this one in every requirements.bzl file?

acozzette (approved):

uv/private/uv_hub/repository.bzl
jbedard: Good point 👍

PR #1295, opened 2026-07-13
fix(py): Resolve Java toolchain paths before the PEP 517 backend chdirs into the sdist

xangcastle:
Bazel expands `$(JAVA)` and `$(JAVABASE)` as workspace-relative paths, but the PEP 517 backend chdirs into the unpacked sdist before building; leaving JAVA_HOME/JAVA pointing at locations that no longer resolve. 

The build helper now resolves them to absolute while it still runs from the action execroot, mirroring `_resolve_compiler_path` for CC/CXX. The initial path.exists guard is dropped (the JDK is a declared action input, so a missing path should fail loudly rather than silently ship a broken env); the shared absolutization is extracted into _absolutize_path so the two toolchain paths can't drift again.

---

### Test plan

### DIFF
--- a/uv/private/pep517_whl/tests/python_env_backend/backend.py
+++ b/uv/private/pep517_whl/tests/python_env_backend/backend.py
@@ -37,6 +37,12 @@ def _check_environment() -> None:
     if os.environ.get("PYTHONSAFEPATH") != "1":
         raise RuntimeError("unrelated PYTHONSAFEPATH was not preserved")
 
+    if os.environ.get("EXPECT_JAVA_TOOL_PATHS") == "1":
+        for name in ("JAVA_HOME", "JAVA"):
+            value = os.environ.get(name)
+            if not value or not os.path.isabs(value) or not os.path.exists(value):
+                raise RuntimeError(f"{name} is not an absolute existing path: {value!r}")
+
 
 def get_requires_for_build_wheel(
     config_settings: dict[str, object] | None = None,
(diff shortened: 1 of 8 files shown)


uv/private/pep517_whl/tests/python_env_backend/backend.py
```diff
@@ -37,6 +37,18 @@ def _check_environment() -> None:
...
+    if os.environ.get("EXPECT_JAVA_TOOL_PATHS") == "1":
+        # JAVA_HOME/JAVA are workspace-relative make-vars that the build
+        # helper must absolutize before the backend chdirs into the sdist.
+        for name in ("JAVA_HOME", "JAVA"):
+            value = os.environ.get(name)
+            if not value or not os.path.isabs(value) or not os.path.exists(value):
+                raise RuntimeError(f"{name} is not an absolute existing path: {value!r}")
+        # No $(JAR) make-var exists; $(JAVABASE)/bin/jar was a manual composite
+        # that broke on the cwd change. Guard against silent reintroduction.
+        if "JAR" in os.environ:
```
tamird: Please drop this assertion. `_common_env` forwards non-Python default shell env, and `build_helper` does not remove `JAR`, so `--action_env=JAR=/usr/bin/jar` makes this fixture fail even though the rule is behaving correctly. The fixture already stopped requesting the stale composite; the useful regression is the absolute/existing `JAVA_HOME` and `JAVA` check above.

tamird (commented):
The JAVA_HOME/JAVA absolutization looks sound, and the shared helper preserves the existing CC/CXX behavior. I found one test issue below. Please also replace the template PR body with the causal summary and test plan before marking this ready.

tamird (approved):
Re-reviewed the cumulative diff at a00a7015. The JAR assertion is gone; JAVA_HOME/JAVA still exercise the backend cwd change, and the shared helper preserves the existing CC/CXX behavior. Clean.

tellett (approved):

PR #1297, opened 2026-07-14
feat: Support `tool.uv.extra-build-dependencies` in annotation loading

konsti-openai:
Currently, `annotations.toml` supports declaring extra build dependencies in a format that runs parallel to [`tool.uv.extra-build-dependencies`](https://docs.astral.sh/uv/reference/settings/#extra-build-dependencies). This PR adds support for loading uv's `tool.uv.extra-build-dependencies` to `rules_py`'s `uv.project`. This moves rules_py closer to uv's interface.

Unlike `annotations.toml`, `tool.uv.extra-build-dependencies` supports the full requirements syntax.

---

### Test plan

Extended the existing e2e snapshot test on top of uv-sdist-fallback to cover both the old and the new syntax for two different packages.

### DIFF
--- a/annotations.toml
+++ b/annotations.toml
@@ -1,12 +1,13 @@
 ## Annotations.toml
 #
 # This is an aspect_rules_py//uv specific configuration file which allows
-# packages to be annotated with their build time dependencies. This fills a gap in
-# both the pylock and uv lockfile formats, neither of which allow for build
-# requirements to be specified.
+# packages to be annotated extra information such as entrypoint.
 #
-# For instance if a package requires cython be available, this is how you can
-# configure it to be delivered.
+# This file allows specifying additional build time dependencies, however
+# <https://docs.astral.sh/uv/reference/settings/#extra-build-dependencies>
+# should be preferred, which is loaded by `uv.project`. For instance if a
+# package requires setuptools be available, this is how you can configure#
+# it to be delivered.
 #
 # Takes the place of giant MODULE.bazel annotation tables.
 #
@@ -17,15 +18,15 @@
 # We version lockfiles and support semver semantics here
 version = "0.0.0"
 
-# Bravado doesn't have bdists, need to build it. Mark explicitly that we need
-# wheel setuptools and build in order to do so; build being the standard build
-# tool driver.
-[[package]]
-name = "bravado-core"
-build-dependencies = [
-   { name = "build" },
-   { name = "setuptools" },
-]
+# This format is supported, but
+# <https://docs.astral.sh/uv/reference/settings/#extra-build-dependencies>
+# is now recommended.
+#[[package]]
+#name = "bravado-core"
+#build-dependencies = [
+#   { name = "build" },
+#   { name = "setuptools" },
+#]
 
 # We can also annotate packages as providing console scripts we care about.
 # Declared console scripts will have Bazel targets generated for them.
--- a/annotations.toml
+++ b/annotations.toml
@@ -1,12 +1,12 @@
 ## Annotations.toml
 #
 # This is an aspect_rules_py//uv specific configuration file which allows
-# packages to be annotated extra information such as entrypoint.
+# packages to be annotated with extra information such as entrypoints.
 #
 # This file allows specifying additional build time dependencies, however
 # <https://docs.astral.sh/uv/reference/settings/#extra-build-dependencies>
 # should be preferred, which is loaded by `uv.project`. For instance if a
-# package requires setuptools be available, this is how you can configure#
+# package requires setuptools to be available, this is how you can configure
 # it to be delivered.
 #
 # Takes the place of giant MODULE.bazel annotation tables.
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -70,6 +70,7 @@ load("//uv/private/whl_install:dist_repository.bzl", "whl_dist")
 load("//uv/private/whl_install:metadata.bzl", "parse_console_script")
 load("//uv/private/whl_install:repository.bzl", "whl_install")
 load(":graph_utils.bzl", "activate_extras", "collect_sccs")
+load(":git_utils.bzl", "locked_git_requirement_urls")
 load(":lockfile.bzl", "build_marker_graph", "collect_bdists", "collect_configurations", "collect_sdists", "normalize_deps", "url_basename")
 load(":projectfile.bzl", "collate_versions_by_name", "collect_activated_extras", "extract_requirement_marker_pairs")
 
@@ -307,6 +308,10 @@ def _parse_projects(module_ctx, hub_specs):
                 for artifact in artifacts:
                     url = artifact.get("url")
                     if url:
+                        locked_urls[(locked_package["name"], url)] = dependency
+                git = locked_package.get("source", {}).get("git")
+                if git:
+                    for url in locked_git_requirement_urls(git):
                         locked_urls[(locked_package["name"], url)] = dependency
 
             def _resolve(name, version):
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -453,7 +453,7 @@ def _parse_projects(module_ctx, hub_specs):
 
             whl_configurations.update(collect_configurations(lock_data))
 
-            configuration_names, activated_extras = collect_activated_extras(project.lock, project_id, project_data, lock_data, default_versions, marker_graph, package_versions)
+            configuration_names, activated_extras = collect_activated_extras(project.lock, project_id, project_data, lock_data, default_versions, marker_graph, package_versions, locked_urls = locked_urls)
             version_activations = collate_versions_by_name(activated_extras)
 
             # SCC graph shapes:
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -325,8 +325,8 @@ def _parse_projects(module_ctx, hub_specs):
                     )
                     if not resolved_deps:
                         fail((
-                            "Unable to resolve extra build dependency {} for package {} in {}. " +
-                            "`uv lock` does not include packages referenced only by " +
+                            "Unable to resolve extra build dependency `{}` for package {} in {}. " +
+                            "`uv.lock` does not include packages referenced only by " +
                             "`tool.uv.extra-build-dependencies`. Add the dependency as a dependency " +
                             "and regenerate the lock."
                         ).format(repr(dep), repr(package), project.pyproject))
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -454,6 +454,8 @@ def _parse_projects(module_ctx, hub_specs):
 
             whl_configurations.update(collect_configurations(lock_data))
 
+            # Build environments have their own exact-version closure. Keep
+            # their extras out of the application's dependency-group graph.
             configuration_names, activated_extras = collect_activated_extras(project.lock, project_id, project_data, lock_data, default_versions, marker_graph, package_versions, locked_urls = locked_urls)
             version_activations = collate_versions_by_name(activated_extras)
 
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -276,6 +276,7 @@ def _parse_projects(module_ctx, hub_specs):
 
         for project in mod.tags.project:
             project_data = toml.decode_file(module_ctx, project.pyproject)
+            tool_uv = project_data.get("tool", {}).get("uv", {})
             lock_data = toml.decode_file(module_ctx, project.lock)
 
             # The stamp derives from the project name: stable across reloads and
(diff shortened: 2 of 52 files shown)


konsti-openai:
I'm new to rules_py, so I'm interested in feedback in how to expose this best: Should we just load extra build dependencies by default? Should this be a separate command similar to `unstable_annotate_packages`?

uv/private/extension/defs.bzl
```diff
@@ -797,6 +816,13 @@ _project_tag = tag_class(
                   "for sdists that build a wheel. Only demanded when an sbuild is " +
                   "guaranteed to be selected (forced builds or sdist-only packages).",
         ),
+        "load_annotations": attr.bool(
```
jbedard: > Should we just load extra build dependencies by default?

I assume we should by default if that is what `uv` does.

Is there any known reason for disabling it? Can we just enable this 100% of the time and drop this attribute?

annotations.toml
```diff
@@ -17,15 +18,15 @@
...
-# Bravado doesn't have bdists, need to build it. Mark explicitly that we need
-# wheel setuptools and build in order to do so; build being the standard build
-# tool driver.
-[[package]]
-name = "bravado-core"
-build-dependencies = [
-   { name = "build" },
-   { name = "setuptools" },
-]
+# This format is supported, but
```
jbedard, line 21: Should we still be testing both though? Or is this old/deprecated enough we can just delete it (instead of commenting it out like this)?

uv/private/extension/defs.bzl
```diff
@@ -219,44 +220,62 @@ def _parse_projects(module_ctx, hub_specs):
...
+                extra_build_dependencies = tool_uv.get("extra-build-dependencies", {})
+                for package, extra_deps in extra_build_dependencies.items():
+                    target = _resolve(package, None)
+                    if target == None:
+                        # Allow a shared annotation file to include entries for other locks.
+                        continue
+                    deps = []
+                    skip = False
+                    for dep in extra_deps:
+                        resolved = _resolve(dep, None)
```
tamird: `extra-build-dependencies` entries are uv requirements, not necessarily bare distribution names. Valid forms include a PEP 508 string such as `cython>=3; sys_platform == 'linux'` and `{ requirement = "torch", match-runtime = true }`. The former cannot match `default_versions` and silently drops the whole annotation; the latter reaches `normalize_name` as a dict and fails evaluation. Please parse the requirement/object forms (including markers and match-runtime semantics), or explicitly reject unsupported forms instead of silently skipping them. A bare-name-only e2e case misses both failures.

uv/private/extension/defs.bzl
```diff
@@ -219,44 +220,62 @@ def _parse_projects(module_ctx, hub_specs):
...
                 elif name in default_versions:
                     return default_versions[name]
                 return None
 
             lock_build_dep_anns = {}
             lock_native_anns = {}
+            if project.load_annotations:
+                extra_build_dependencies = tool_uv.get("extra-build-dependencies", {})
+                for package, extra_deps in extra_build_dependencies.items():
+                    target = _resolve(package, None)
```
tamird: Resolving the annotated package only through `default_versions` silently skips every valid split-version lock: that map is populated only when a name has one version, while uv applies the annotation by package name across selected versions. The same issue affects build dependencies below. Please expand the target/dependency resolution over `package_versions` and add a conflicting-version or platform-fork regression.

uv/private/extension/defs.bzl
```diff
@@ -219,44 +220,62 @@ def _parse_projects(module_ctx, hub_specs):
...
+                    deps = []
+                    skip = False
+                    for dep in extra_deps:
+                        resolved = _resolve(dep, None)
+                        if resolved == None:
+                            skip = True
+                            break
+                        deps.append(resolved)
+                    if not skip:
+                        lock_build_dep_anns[target] = deps
```
tamird: This requires every extra build dependency to already be present in the runtime `uv.lock`; otherwise `skip` discards the complete annotation. A common valid case is `cchardet = ["cython"]` where Cython is build-only and absent from the project lock. The current e2e fixture masks this because its build helpers are direct/transitive runtime dependencies. How will build-only requirements be resolved? Please cover a backend that genuinely fails without a distinct extra dependency.

uv/private/extension/defs.bzl
```diff
@@ -219,44 +220,62 @@ def _parse_projects(module_ctx, hub_specs):
...
                         if "build-dependencies" in package:
                             deps = []
                             skip = False
                             for dep in package["build-dependencies"]:
-                                resolved = _resolve(dep)
+                                resolved = _resolve(dep["name"], dep.get("version"))
                                 if resolved == None:
                                     skip = True
                                     break
                                 deps.append(resolved)
```
tamird, line 401: When both uv and `unstable_annotate_packages` annotate the same package, this assignment replaces the uv-derived build dependencies rather than composing them. The e2e case deliberately annotates different packages, so the migration/precedence behavior is untested. Please define and test the intended union-or-override semantics; silently losing either set will produce incomplete builds.

tamird (changes requested):
This currently accepts only a narrow subset of uv's extra-build-dependencies syntax and silently drops several valid cases. Please either preserve uv's semantics for requirement objects, markers, split versions, and build-only dependencies or explicitly reject unsupported forms, and add an e2e case whose build really depends on the annotation.

uv/private/extension/defs.bzl
konsti-openai: I've added that, though it's important to note that `build-system.requires` extracts only the name (https://github.com/aspect-build/rules_py/blob/5204fc2ad524d2b6f2da2de9936de02882dbc5a2/uv/private/sdist_configure/detect_native.py#L124-L128), so we have a divergence between how `tool.uv.extra-build-dependencies` and `build-system.requires` are parsed.

uv/private/extension/defs.bzl
konsti-openai: I'm not aware of any, dropped!

annotations.toml
konsti-openai: I kept it there to not confuse user who come across the old format. imho it's fine to keep both format around and avoid breaking users, the comment should help users to understand what they should use and why they may see something different in existing code.

konsti-openai:
Codex also flagged two other problems. They are already present with `annotations.toml`, so I'm cutting scope here:

> - Dependency-group independence: build dependencies resolve through the active runtime group. If a required build package exists only in another group, Bazel analysis can fail.
> - Version preservation: although an annotation can resolve a specific build-dependency version, the generated package alias may select the version associated with the active runtime group instead.

uv/private/extension/defs.bzl
```diff
@@ -245,44 +246,89 @@ def _parse_projects(module_ctx, hub_specs):
...
+                        dep = dep["requirement"]
+                    resolved_deps = extract_requirement_marker_pairs(
+                        project.lock,
+                        project_id,
+                        dep,
+                        {},
+                        package_versions,
+                        fail_if_missing = False,
+                    )
+                    if not resolved_deps:
```
konsti-openai, line 364: It's arguable whether this should or shouldn't fail. Codex complains that this fails for wheel-only packages while `annotations.toml` skips, but OTOH annotating a wheel-only package with requirements missing in the lock sounds bogus.

konsti-openai:
I'm having a hard time scoping this against preexisting gaps and making this too large for review, but I think the current state is ready for a review.

uv/private/extension/defs.bzl
```diff
@@ -504,11 +586,37 @@ def _parse_projects(module_ctx, hub_specs):
...
+                                fail_if_missing = sbuild_required,
+                            )
                         ]
 
                     build_deps = sets.to_list(sets.make(build_deps + lock_build_deps))
 
+                    # A base requirement and its extras resolve through the same
+                    # project package label, so deduplicate after rendering labels.
+                    sbuild_deps = sets.to_list(sets.make([
+                        "@{0}//:{1}".format(*dep)
```
tamird: This discards the resolved build-dependency version and always emits the versionless project alias. With conflicting groups, a plain requirement such as torch==2.5 can therefore build against the active runtime torch==2.6; conversely the object form's match-runtime flag is explicitly ignored above, so the documented guarantee is not enforced. The current fixtures keep six/pyparsing single-version and cannot catch either case. Please preserve version-bearing install labels for pinned requirements, implement (or explicitly reject) match-runtime, and add a two-group/two-version source-build regression that asserts the configured build tool receives the intended version. uv documents the distinction here: https://docs.astral.sh/uv/concepts/projects/config/#augmenting-build-dependencies

— tamirdex

uv/private/extension/defs.bzl
```diff
@@ -245,44 +247,107 @@ def _parse_projects(module_ctx, hub_specs):
...
+                    resolved_deps = extract_requirement_marker_pairs(
+                        project.lock,
+                        project_id,
+                        dep,
+                        {},
+                        package_versions,
+                        locked_urls = locked_urls,
+                        fail_if_missing = False,
+                    )
+                    if not resolved_deps:
```
tamird, line 364: This resolves and fails every annotated package before checking whether a source build is needed. A wheel-only package with an otherwise harmless missing build annotation therefore breaks module evaluation, even though uv would never build it. More fundamentally, the new missing-extra-build-dep fixture is uv's canonical cchardet = ["cython"] example, so this API cannot support its primary build-only use case without polluting runtime dependencies. Please skip irrelevant wheel-only annotations and either resolve build-only requirements separately or clearly document this limitation in the public release/API documentation.

— tamirdex

uv/private/extension/defs.bzl
```diff
@@ -245,44 +247,107 @@ def _parse_projects(module_ctx, hub_specs):
...
+            for locked_package in lock_data.get("package", []):
+                dependency = (project_id, locked_package["name"], locked_package["version"], "__base__")
+                artifacts = locked_package.get("wheels", []) + [
+                    locked_package.get("sdist", {}),
+                    locked_package.get("source", {}),
+                ]
+                for artifact in artifacts:
+                    url = artifact.get("url")
+                    if url:
+                        locked_urls[(locked_package["name"], url)] = dependency
```
tamird: The direct-reference index includes wheel/sdist URLs and git remotes, but not locked source.path/directory/editable entries. A valid locked requirement such as name @ file:///... therefore reaches the missing-dependency error even though the source is in uv.lock. Since this change advertises full requirement syntax, please support the locked local-source forms or explicitly reject/document them, with a local-source regression.

— tamirdex

tamird (changes requested):
Re-reviewed because Jason requested another pass in #ext-aspect-build. The updated implementation now covers markers, extras, split target versions, URL/git references, and composition with legacy annotations, but two correctness gaps remain: selected build-dependency versions are erased (and match-runtime is accepted but ignored), and a marker on a requested extra is not propagated to that extra's transitive dependencies.

Pre-existing path amplified by this change: uv/private/extension/projectfile.bzl:261-270 traverses children using only each edge marker and drops the parent's reachability marker. Thus urllib3[socks]; sys_platform != 'win32' can activate its transitive PySocks on Windows. The newly added collect_activated_extras_build_extra_test codifies this loss by supplying a sys_platform root marker and asserting the child has only its CPython edge marker. Carry/conjoin parent and edge markers through the worklist and add a false-platform regression.

The primary uv use case is also still unsupported: the new negative fixture is the documented cchardet/cython example, so users must add a build-only dependency to runtime to use this API. If that limitation is intentional for this release, it needs to be explicit in the user-facing documentation/release note, and wheel-only targets should not fail an otherwise unnecessary annotation. The current fixtures do not falsify these cases.

— tamirdex

uv/private/extension/defs.bzl
konsti-openai: I'd leave match-runtime unsupported here and keep this for a follow-up, to limit the scope here. I don't think we need to explicitly reject; We're already pretty good as all build deps must come from locked runtime deps, and this feature was invented due to non-bazel uv doing independent resolutions for build deps and runtime deps.

uv/private/extension/defs.bzl
konsti-openai: Documented the limitation; I think it's reasonable to have this requirement if you're already using extra annotations.

konsti-openai:
> The primary uv use case is also still unsupported: the new negative fixture is the documented cchardet/cython example, so users must add a build-only dependency to runtime to use this API. If that limitation is intentional for this release, it needs to be explicit in the user-facing documentation/release note, and wheel-only targets should not fail an otherwise unnecessary annotation. The current fixtures do not falsify these cases.

Where would that go? I've looked for `annotate_packages` and `annotations.toml` but they don't seem to be mentioned in any markdown file. Note that the same limitation applies to `annotations.toml`, in both cases we have to lock separately (as uv does not support build dependency locking).

> Pre-existing path amplified by this change: uv/private/extension/projectfile.bzl:261-270 traverses children using only each edge marker and drops the parent's reachability marker. Thus urllib3[socks]; sys_platform != 'win32' can activate its transitive PySocks on Windows. The newly added collect_activated_extras_build_extra_test codifies this loss by supplying a sys_platform root marker and asserting the child has only its CPython edge marker. Carry/conjoin parent and edge markers through the worklist and add a false-platform regression.

Fixed those

konsti-openai:
This PR is growing quite a bit, if it's helpful I can split out some refactorings into a separate PR.

uv/private/extension/defs.bzl
tamird: Keeping runtime-version matching out of scope is reasonable. The remaining problem is that match-runtime = true explicitly requests a version guarantee, but this implementation silently discards it. Please reject true with an actionable unsupported error and cover that path with one negative fixture; ordinary requirement strings and match-runtime = false can continue to work. That keeps this change small without promising matching behavior it does not implement.

[tamirdex]

uv/private/extension/defs.bzl
tamird: The new paragraph in docs/uv.md is exactly the right place. It clearly explains that uv.lock must contain extra build requirements even when a wheel might be selected. That resolves my documentation concern, and I agree that annotations.toml has the same locking limitation.

[tamirdex]

jbedard:
This has grown quite a bit and smaller PRs would be ideal if you can. What is the priority of this?

jbedard:
@konsti-openai what's the status here? Is OpenAI patching this internally and hoping this will be supported directly by rules_py still?

konsti-openai:
I had to reprioritize, no update here

PR #1298, opened 2026-07-15
refactor: refactoring venv assembly system

xangcastle:
---

### Test plan

- New test cases added


acozzette:
## Code review findings

Ran a high-effort review (multi-angle finder pass + independent verification, including `bazel build //...` and the full test suite, which pass) over this refactor. 8 findings survived verification, ranked most severe first.

### Correctness

1. **Collision-policy error message ordering changed** — `virtuals_resolvers.bzl:487`
   `_resolve_metadata_collisions`'s unconditional `fail()` now always runs to completion before the deferred `enforce_collision_policy` (`assemble_venv.bzl:298`) gets a chance to raise the recorded top-level/console-script collision message. A target with `package_collisions = "error"` that has *both* an ordinary top-level collision *and* an unrelated metadata-collision loser will always fail with the internal metadata-invariant message instead of the intended, actionable collision message. Build still fails either way, but the reported root cause is misleading. Pre-refactor pass ordering meant the top-level collision fired first.

2. **Snapshot tool silently drops absolute-target symlinks** — `py/private/py_venv/snapshots/venv_tree_snap.bzl:29`
   The relative-target branch always emits a `LINK` line; the absolute-target branch only tries to inline the dereferenced file's contents (gated on `-f` + `wc -c < 8192`), with no unconditional `LINK` echo. For a non-hermetic/system Python toolchain, `bin/python` is an absolute-target symlink to a real interpreter binary (>8192 bytes) — the snapshot test silently omits it (no FILE, no LINK), so it wouldn't catch a regression in that symlink's target.

3. **`pythonMAJOR.MINOR` directory name duplicated 4x with no shared helper** — `toolchains_resolver.bzl:24`
   The same `"python{}.{}".format(major, minor)` format string is independently written in `toolchains_resolver.bzl`, `py_unpacked_wheel.bzl`, and twice in `unpack.py`. These must stay byte-identical for wheel resolution to find the right `site-packages` dir; a future edit to one copy (padding, free-threaded suffix, etc.) without the others fails silently at runtime with no build-time signal.

### Cleanup

4. **Redundant derivable parameter `venv_name`** — `assemble_venv.bzl:267`
   Always constructed as `".{}".format(safe_name)` at its one call site (`py_venv.bzl:133`) but threaded separately through ~6 helper functions instead of computed once internally.

5. **Dead parameter `any_namespace`** — `virtuals_resolvers.bzl:315`
   `_resolve_directory_collision` never references it in its body; the caller computes it for nothing.

6. **Triplicated path-containment check** — `virtuals_resolvers.bzl:15`
   `_is_within`, `_within_any`, and `_contains_any` each inline the same `path == root or path.startswith(root + "/")` check instead of the latter two calling `_is_within` in a loop. `_is_within` itself is never called anywhere.

7. **`_build_wheel_lookup_sets` rebuilt 3-4x for the same claimant set** — `virtuals_resolvers.bzl:199`
   Called independently in `_scan_namespace_conflicts`, `_classify_conflicted_roots`, and `_resolve_native_span` (plus an inline per-claimant rebuild), each discarding 2 of 3 returned dicts. Negligible perf impact at realistic wheel-collision counts, but pure waste.

8. **Stale docstring reference to a nonexistent provider** — `virtuals_resolvers.bzl:569`
   `compute_wheel_plan`'s docstring says the struct has "all fields needed by `PyWheelPlanInfo`" — no such provider exists anywhere in the repo.

Two other candidates were investigated and refuted: an apparent `-I` interpreter-flag inconsistency between `py_venv.bzl` and `py_venv_exec.bzl` (intentional — only `py_venv_exec` has an `isolated` attribute), and an apparent duplicate toolchain-lookup pattern between `assemble_venv.bzl` and `py_unpacked_wheel.bzl` (correctly tracks each rule's different mandatory-vs-optional toolchain registration).

PR #1301, opened 2026-07-16
refactor(venv): Consolidate py_venv public API into defs.bzl

xangcastle:
Move the public `py_venv` Starlark API into a single `defs.bzl` entry point so external callers load from one place instead of splitting imports `across py_venv.bzl` and `py_venv_exec.bzl`.

---

### Test plan


py/private/py_venv/defs.bzl
```diff
@@ -0,0 +1,30 @@
+"""Re-exports of the public API for the ``py_venv`` package.
+
+External consumers should load from ``//py/private/py_venv:defs.bzl``
```
jbedard, line 3: External consumers should load from `//py:defs.bzl`, this file is still private

jbedard (approved):

PR #1305, opened 2026-07-16
refactor!: own the interpreter runtime toolchain rules instead of rules_python's

jbedard:
Generated interpreter repos instantiate a single rules_py py_runtime_toolchain
target (@python_X//:runtime) in place of the rules_python py_runtime +
py_runtime_pair + py_exec_tools_toolchain stack, backing both toolchain
registrations. It carries rules_python's public PyRuntimeInfo (re-exported from
@aspect_rules_py//py:defs.bzl), so interop is preserved both ways: rules_py
consumes rules_python-provisioned runtimes, and rules_python executables and
their consumers (py_console_script_binary, py_zipapp_binary, py_interpreter)
run on rules_py-provisioned toolchains. Each repo's rules_python loads drop
from 31 files to the py_cc_toolchain chain plus py_runtime_info.bzl.

Build actions (whl_install, py_unpacked_wheel, py_venv site-merge) resolve a
rules_py-owned exec-tools toolchain type; rules_python's exec-tools type is
left untouched. Registrations are version-gated to follow the Python version
flags (#1309), with a highest-version fallback that resolves in every consumer
via the hub bundled in rules_py's MODULE. .pyc precompilation is skipped when
the exec and target interpreter_version_info differ, since mismatched bytecode
magic (including between same-minor prereleases) is unusable by the target.

Breaking: generated repo targets @python_X//:py3_runtime, :runtime_pair, and
:exec_tools_toolchain are now a single @python_X//:runtime. Direct label
references need updating; toolchain resolution and provider identity are
unaffected.

### Test plan

### DIFF
--- a/py/private/interpreter/repository.bzl
+++ b/py/private/interpreter/repository.bzl
@@ -201,9 +201,7 @@ cc_library(
     return """\
 load("@rules_cc//cc:cc_import.bzl", "cc_import")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
-load("@rules_python//python:py_runtime.bzl", "py_runtime")
-load("@rules_python//python:py_runtime_pair.bzl", "py_runtime_pair")
-load("@rules_python//python:py_exec_tools_toolchain.bzl", "py_exec_tools_toolchain")
+load("@aspect_rules_py//py/private/interpreter:runtime.bzl", "py_runtime_toolchain")
 load("@rules_python//python/cc:py_cc_toolchain.bzl", "py_cc_toolchain")
 
 package(default_visibility = ["//visibility:public"])
@@ -244,8 +242,8 @@ filegroup(
     deps = {header_interface_deps},
 )
 
-py_runtime(
-    name = "py3_runtime",
+py_runtime_toolchain(
+    name = "runtime",
     abi_flags = "{abi_flags}",
     files = [":files"],
     interpreter = "{python_bin}",
@@ -256,17 +254,6 @@ py_runtime(
         "releaselevel": "{releaselevel}",
         "serial": "{serial}",
     }},
-    python_version = "PY3",
-)
-
-py_runtime_pair(
-    name = "runtime_pair",
-    py2_runtime = None,
-    py3_runtime = ":py3_runtime",
-)
-
-py_exec_tools_toolchain(
-    name = "exec_tools_toolchain",
 )
 
 py_cc_toolchain(
@@ -433,6 +420,7 @@ config_setting(
     # offers a matching pair: https://github.com/aspect-build/rules_py/issues/1095
     #
     # Second pass: emit toolchain() registrations.
+    exec_tools_fallbacks = {}  # repr(exec_compatible_with) -> (version tuple, name, repo, constraints)
     for info, platform_setting_names in toolchain_infos:
         extra_config_settings = info.get("config_settings", [])
         extra_target_compatible = info.get("target_compatible_with", [])
@@ -461,7 +449,7 @@ toolchain(
     name = "{name}",
     target_compatible_with = {target_compatible_with},
     target_settings = {target_settings},
-    toolchain = "@{repo}//:runtime_pair",
+    toolchain = "@{repo}//:runtime",
     toolchain_type = "@bazel_tools//tools/python:toolchain_type",
 )
 
@@ -483,17 +471,46 @@ toolchain(
             content.append("""# Exec tools toolchain: selected by exec platform (not target platform) so
 # that build actions using the interpreter (e.g. compileall) get a runnable
 # binary on the build host regardless of the target platform being built for.
+# Version-gated so the exec interpreter follows the version flags.
 toolchain(
     name = "{name}_exec_tools",
     exec_compatible_with = {exec_compatible_with},
-    toolchain = "@{repo}//:exec_tools_toolchain",
-    toolchain_type = "@rules_python//python:exec_tools_toolchain_type",
+    target_settings = ["{version_setting}"],
+    toolchain = "@{repo}//:runtime",
+    toolchain_type = "@aspect_rules_py//py/private/toolchain:exec_tools_toolchain_type",
 )
 """.format(
                 name = info["name"],
                 repo = info["repo"],
                 exec_compatible_with = exec_compatible_with,
+                version_setting = version_setting,
             ))
+
+            # Track the platform's highest provisioned version for the
+            # ungated fallback entry emitted after the loop.
+            version_key = tuple([int(part) for part in info["python_version"].split(".")])
+            prior = exec_tools_fallbacks.get(repr(exec_compatible_with))
+            if prior == None or version_key > prior[0]:
+                exec_tools_fallbacks[repr(exec_compatible_with)] = (version_key, info["name"], info["repo"], exec_compatible_with)
+
+    # Ungated exec-tools fallbacks: build actions only need *a* runnable host
+    # interpreter, so configurations matching no version-gated entry (e.g. the
+    # version flags at defaults not provisioned by this hub) fall back to the
+    # platform's highest provisioned version. Registration is lexicographic by
+    # name, so the `zz_` prefix sorts these after every gated entry.
+    for _, name, repo, constraints in exec_tools_fallbacks.values():
+        content.append("""
+toolchain(
+    name = "zz_{name}_exec_tools_fallback",
+    exec_compatible_with = {exec_compatible_with},
+    toolchain = "@{repo}//:runtime",
+    toolchain_type = "@aspect_rules_py//py/private/toolchain:exec_tools_toolchain_type",
+)
+""".format(
+            name = name,
+            repo = repo,
+            exec_compatible_with = constraints,
+        ))
 
     content.append("""
 exports_files(
--- a/py/private/interpreter/runtime.bzl
+++ b/py/private/interpreter/runtime.bzl
@@ -1,0 +1,100 @@
+"""Python runtime toolchain rule for provisioned interpreter repositories.
+
+A minimal replacement for the rules_python `py_runtime` + `py_runtime_pair` +
+`py_exec_tools_toolchain` rule stack: one target provides the runtime and the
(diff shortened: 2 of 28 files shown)


py/private/interpreter/runtime.bzl
```diff
@@ -0,0 +1,129 @@
...
+        runtime,
+        platform_common.ToolchainInfo(
+            # The standard Python toolchain contract
+            # (@bazel_tools//tools/python:toolchain_type).
+            py2_runtime = None,
+            py3_runtime = runtime,
+            # The //py/private/toolchain:exec_tools_toolchain_type contract.
+            exec_runtime = runtime,
+        ),
+        DefaultInfo(files = depset([ctx.file.interpreter], transitive = [runtime.files])),
```
tamird, line 54: The runtime returned through the standard Python toolchain now carries rules_py's distinct PyRuntimeInfo. A rules_python py_binary/py_test forwards that runtime object, and rules_python py_zipapp_binary/py_zipapp_test (and explicit py_interpreter(binary=...)) require and index rules_python's PyRuntimeInfo, so those compositions fail analysis even though the binary itself can run. Could this retain/publish the rules_python runtime provider on the standard-toolchain path and add an e2e zipapp/interpreter consumer over a rules_python binary using the provisioned runtime?

py/private/toolchain/types.bzl
```diff
@@ -1,5 +1,5 @@
 """Constants for toolchain types"""
 
 PY_TOOLCHAIN = "@bazel_tools//tools/python:toolchain_type"
-EXEC_TOOLS_TOOLCHAIN = "@rules_python//python:exec_tools_toolchain_type"
+EXEC_TOOLS_TOOLCHAIN = "@aspect_rules_py//py/private/toolchain:exec_tools_toolchain_type"
```
tamird, line 4: Changing the exec-tools type to a new rules_py-private type breaks the supported mixed setup where a module provisions only rules_python's python.toolchain() and consumes rules_py uv dependencies: WhlInstall/py_unpacked_wheel/site-merge now have no matching exec toolchain. The rules-python-interop case provisions a rules_python runtime but only exercises pip.parse, so it misses this. Please keep/bridge the existing exec-tools type or register an adapter automatically, and extend that interop case with a rules_py uv wheel (and ideally a site-merge collision).

py/private/interpreter/repository.bzl
```diff
@@ -483,18 +471,47 @@ toolchain(
...
+    # version flags at defaults not provisioned by this hub) fall back to the
+    # platform's highest provisioned version. Registration is lexicographic by
+    # name, so the `zz_` prefix sorts these after every gated entry.
+    for _, name, repo, constraints in exec_tools_fallbacks.values():
+        content.append("""
+toolchain(
+    name = "zz_{name}_exec_tools_fallback",
+    exec_compatible_with = {exec_compatible_with},
+    toolchain = "@{repo}//:runtime",
+    toolchain_type = "@aspect_rules_py//py/private/toolchain:exec_tools_toolchain_type",
```
tamird, line 507: Falling back to the highest provisioned exec interpreter when the active target version is absent is unsafe for compile_pyc. WhlInstall takes the target layout/version from the standard toolchain but runs compileall with this exec runtime; a 3.12 target plus a 3.13-only rules_py hub therefore installs 3.13-magic pyc under python3.12. Require a matching exec version (or explicitly disable/error on pyc compilation when none exists) and add a mixed-version compile_pyc regression.

tamird (changes requested):
There are three compatibility/correctness gaps in the refreshed runtime boundary: the emitted runtime provider breaks rules_python zipapp/interpreter composition, the new private exec-toolchain type breaks modules that provision only a rules_python interpreter but consume rules_py uv wheels, and the highest-version fallback can emit unusable pyc for a different target Python version. The current interop coverage does not exercise these paths. The PR description also appears stale: it says the exec-toolchain type is unchanged while the code changes it, and says rules_python binaries cannot execute while the new coverage runs one.

py/private/interpreter/runtime.bzl
jbedard: Why would "A rules_python py_binary/py_test" ever receive "rules_py's distinct PyRuntimeInfo"?

tamird: Because both rule sets resolve the same standard toolchain type. In rules_python 1.9.0, `@rules_python//python:toolchain_type` aliases `@bazel_tools//tools/python:toolchain_type`; this PR registers the provisioned runtime against that type and puts its new provider in `ToolchainInfo.py3_runtime`. The existing `rules-python-consumers:pycowsay` fixture already exercises a rules_python binary landing on that runtime.

rules_python's executable rule field-reads `py3_runtime`, so the binary itself works, then forwards that same provider object. Adding `py_zipapp_binary(name = "app_zip", binary = ":pycowsay")` fails because zipapp requires and indexes rules_python's `PyRuntimeInfo`, not the structurally compatible rules_py provider. `py_interpreter(binary = ...)` has the same identity check. That is the composition the current fixture misses.

jbedard: Should we drop the use of `@bazel_tools//tools/python:toolchain_type` as well then?

tamird: I would keep the standard target type. `@bazel_tools//tools/python:toolchain_type` is the shared target-runtime contract, not a rules_python implementation detail: third-party/system/FIPS runtimes and both rule sets use it. Moving rules_py to a private target type would make mixed builds select two independent runtimes and would break the currently supported direction where rules_py consumes a rules_python-provisioned interpreter.

The smaller option is to keep the shared type and use/re-export rules_python's public `@rules_python//python:py_runtime_info.bzl` provider for `ToolchainInfo.py3_runtime`. That still removes the `py_runtime + py_runtime_pair + py_exec_tools_toolchain` rule stack while preserving provider identity for zipapp/interpreter composition; one generated runtime target can serve both registrations. A fully private target type is coherent only as an explicit full-detachment/major-version migration, with adapters for external runtimes, and is materially broader than this load-reduction refactor.

tamird: The new head correctly re-exports rules_python's public `PyRuntimeInfo` and places that exact provider in the shared toolchain, which fixes the composition bug. There is still no `py_zipapp_binary`/`py_zipapp_test` or explicit-interpreter consumer in the e2e suite: `pycowsay` only runs the underlying binary. Please add the small zipapp regression so provider identity cannot silently regress.

py/private/toolchain/types.bzl
tamird: The new rules-python-interop case now provisions rules_python's target runtime, consumes a real rules_py uv/cowsay wheel through the bundled exec fallback, and passes on Bazel 8/9. That covers the mixed setup from this thread; resolving. Small documentation correction: the new comments call the fallback an adapter bridging rules_python exec tools, but the implementation is the bundled PBS exec fallback.

py/private/interpreter/repository.bzl
tamird: The new major/minor guard in `whl_install/rule.bzl` correctly disables compileall when target and exec runtimes differ, and the matching/mismatching action-argv tests exercise both paths. That addresses the wrong-magic `.pyc` failure; resolving.

tamird (commented):
The new head fixes and meaningfully covers mixed provisioning and target/exec bytecode-version mismatch. The provider identity is now correct, but please retain the small zipapp regression requested in the remaining thread. The PR description/release notes also need a refresh: they still describe a distinct rules_py `PyRuntimeInfo`, no rules_python minimum/API dependency, and rules_python-only setups failing, while the new code and tests now deliberately do the opposite.

tamird:
All 88 checks are green. The mixed-provisioning and target/exec bytecode-mismatch threads are fixed and covered, but GitHub does not permit an external reviewer to resolve them; could a maintainer resolve those two? The remaining request is the small zipapp/provider-identity regression plus updating the now-stale release notes.

uv/private/whl_install/rule.bzl
```diff
@@ -105,10 +105,16 @@ def _whl_install(ctx):
...
+    # Use the exec-configured interpreter from the exec-tools toolchain so cross-arch
     # builds work (the target interpreter isn't runnable on the build host). This is
-    # safe because .pyc bytecode varies by Python version, not by architecture.
-    if ctx.attr.compile_pyc:
+    # safe because .pyc bytecode varies by Python version, not by architecture — so
+    # skip it when a fallback exec runtime doesn't match the target version: its
+    # bytecode magic would be wrong for the lib/python{version}/ layout.
+    exec_matches_target = (
+        exec_runtime.interpreter_version_info.major == py_toolchain.interpreter_version_info.major and
+        exec_runtime.interpreter_version_info.minor == py_toolchain.interpreter_version_info.minor
```
tamird: Comparing only major/minor is not sufficient for bytecode compatibility. CPython changes the magic number between prereleases of the same minor; this is a supported mixed-provisioning case here because rules_python 1.9.0 can select 3.15.0a2 while rules_py provisions 3.15.0a6. Both report 3.15, but their magic numbers are 3656 and 3660, so this guard enables compileall and produces unusable pyc files for the target runtime. Please require the full interpreter_version_info (including release level/serial and micro) to match, and add a same-minor prerelease-mismatch regression.

tamird (changes requested):
The zipapp regression and updated runtime documentation address the prior provider-identity request. One bytecode-compatibility gap remains in the precompile guard.

tamird (changes requested):
One bytecode-compatibility gap remains in the existing unresolved thread: exec_matches_target still accepts different CPython prereleases of the same minor. rules_python 1.9.0 resolves 3.15 to 3.15.0a2 (magic 3656), while rules_py provisions 3.15.0a6 (magic 3660); this enables compileall and installs unusable pyc for the target runtime. Please compare the complete interpreter_version_info, including micro/releaselevel/serial, and add a same-minor prerelease-mismatch analysis test. The provider-identity and mixed-provisioning fixes now look correct.

xangcastle (approved):

PR #1306, opened 2026-07-16
test(venv): Add venv tree snapshots and expose all_files in VirtualenvInfo

xangcastle:
This change adds an `all_files` field to `VirtualenvInfo` so snapshot consumers can depend on the complete venv output set, and introduces a new `venv_tree` snapshot rule under `py/private/py_venv/snapshots/` that renders the full venv tree as a reviewable text file. Generated launcher and tree snapshots are checked in to catch unintended changes to venv assembly. 

---

### Test plan

### DIFF
--- a/py/private/py_venv/tests/BUILD.bazel
+++ b/py/private/py_venv/tests/BUILD.bazel
@@ -1,0 +1,53 @@
+load("@bazel_lib//:bzl_library.bzl", "bzl_library")
+load("@bazel_lib//lib:write_source_files.bzl", "write_source_files")
+load("//py/private/py_venv:defs.bzl", "py_venv", "py_venv_exec")
+load(":venv_tree.bzl", "venv_tree")
+
+py_venv(
+    name = "venv",
+    testonly = True,
+    srcs = ["main.py"],
+    tags = ["manual"],
+    deps = ["@pypi//cowsay"],
+)
+
+py_venv_exec(
+    name = "exec",
+    testonly = True,
+    srcs = ["main.py"],
+    main = "main.py",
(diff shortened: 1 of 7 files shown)


py/private/py_venv/tests/snapshots/venv.launcher.snap
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
jbedard, line 1: This file is in a `snapshots/snapshots` directory, and not in a `/test/` directory...

py/private/py_venv/tests/BUILD.bazel
```diff
@@ -0,0 +1,44 @@
+load("@bazel_lib//:bzl_library.bzl", "bzl_library")
+load("@bazel_lib//lib:write_source_files.bzl", "write_source_files")
+load("//py/private/py_venv:defs.bzl", "py_venv", "py_venv_exec")
+load(":venv_tree.bzl", "venv_tree")
+
+py_venv(
```
jbedard, line 6: Should everything in this BUILD be `testonly = True`?

Is this the directory which should be renamed to `test`?

py/private/py_venv/tests/snapshots/venv.launcher.snap
xangcastle: Done

py/private/py_venv/tests/BUILD.bazel
xangcastle: testonly = True & some tags manual added

jbedard (approved):

PR #1309, opened 2026-07-16
fix(py): give rules_py python_version flag precedence over rules_python

jbedard:
### Test plan

- New test cases added

### DIFF
--- a/e2e/cases/current-py-toolchain-896/BUILD.bazel
+++ b/e2e/cases/current-py-toolchain-896/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load(":report_version.bzl", "report_version")
 
 # Verify that $(PYTHON3) Make variable is provided by
 # @python_interpreters//:current_py_toolchain.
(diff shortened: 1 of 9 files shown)


xangcastle (approved):

tamird:
I checked the refreshed coverage while aligning #1318. The new `rules-python-consumers/test.sh` assumes the version flag selects `exec_tools`, but generated `*_exec_tools` registrations have only `exec_compatible_with` and no version `target_settings`, so that toolchain cannot follow either version flag; the script is also not exercised by the e2e Bazel tests. The runtime-toolchain test is valid. Could we drop the exec-tools assertion (or separately define/test its intended selection semantics) before landing? #1318's launcher/C/ABI3 regression will still cover the newly exposed aliases.

tamird:
Correction: I checked the exact Bazel 8/9 e2e logs and both scripts did run and pass, including every exec-tools version assertion. Please disregard the exec-tools concern above. The only integration detail for #1318 is that both PRs now add `rules-python-consumers/test.sh` and touch its BUILD file; I will combine that coverage on rebase.

PR #1316, opened 2026-07-16
perf(benchmark): gate on Bazel analysis phase and report per-function Starlark cost

xangcastle:
Rewrite the benchmark to gate on Bazel's analysis phase (runAnalysisPhase extracted from --profile) instead of process wall time, excluding JVM/IO noise. The two separate workspaces are merged into a single parameterized one where --packages drive both the analysis load and a fan-in py_binary whose venv scales with it, and the PR comment now surfaces a per-function Starlark CPU breakdown (via --starlark_cpu_profile) so that when something regresses it tells you which functions got slower not just that it did.

---

### Test plan


benchmark/profile_benchmark.py
```diff
@@ -0,0 +1,294 @@
...
+    rows: list[dict[str, Any]] = []
+    for name in names:
+        series = [d.get(name, 0.0) for d in star_runs]
+        rows.append({"name": name, "mean_ms": statistics.mean(series)})
+    total = sum(r["mean_ms"] for r in rows) or 1.0
+    for r in rows:
+        r["pct"] = r["mean_ms"] / total * 100.0
+        r["runs"] = len(star_runs)
+    rows.sort(key=lambda r: r["mean_ms"], reverse=True)
+    return rows[:20]
```
tamird: Truncating each side to its own top 20 before serialization makes the comparator treat a PR function just below main's cutoff as absent (`main=0`) and report it as `new` with its full runtime as a regression. The current report already shows false-new entries such as `select`, `filegroup`, and `_py_venv_lib`. Please keep all aggregated rows (or at least a matched union) in the JSON and truncate only while rendering; a comparator regression covering a function below main's cutoff would catch this.

tamird (commented):
The analysis-phase gate is useful; I found one issue that makes the per-function comparison report false new regressions.

benchmark/profile_benchmark.py
tamird: The implementation and refreshed report now look correct: _py_venv_lib, select, and the other shared functions retain nonzero main times. Could you add the focused regression from the original comment—a function below main's former top-20 cutoff that must be reported with its real main time rather than new—before resolving this?

benchmark/compare.py
```diff
@@ -0,0 +1,400 @@
...
+        # ~sqrt(n)x and flag sampling jitter as significant.
+        sn = math.sqrt(n)
+        return math.sqrt((m_std / sn) ** 2 + (p_std / sn) ** 2)
+
+    total_se = combined_se(main_total.get("stddev_ms", 0.0), pr_total.get("stddev_ms", 0.0))
+    main_total_ms = main_total.get("mean_ms", 0.0)
+    pr_total_ms = pr_total.get("mean_ms", 0.0)
+    delta_total = pr_total_ms - main_total_ms
+    pct_total = pct(main_total_ms, pr_total_ms) if main_total_ms else 0.0
+    total_significant = total_se > 0 and delta_total > TOTAL_SIGMA * total_se
```
tamird: This total-level gate hides a real function regression whenever an unrelated speedup cancels it, even though the section is specifically meant to explain per-function changes. It also treats zero variance as no signal (`total_se > 0`), so a deterministic positive regression is reported as clean; the per-function flag at line 168 has the same issue. Please compute/render significant positive movers independently of the signed total, treat `stderr == 0 && delta > 0` as significant, and add focused cases for a cancelling speedup and zero-variance regression.

MODULE.bazel.lock
```diff
@@ -301,6 +301,197 @@
         ]
       }
     },
+    "@@llvm+//3rd_party/gcc/extension:gcc.bzl%gcc": {
```
tamird: This root lockfile extension state was introduced only by the latest benchmark commit. The benchmark runs its own generated workspace, so the added LLVM/GCC, pybind11, and rules_fuzzing state is unrelated and can make root `--lockfile_mode=error` depend on local evaluation history. Please restore the root lockfile to the merge-base version.

tamird (commented):
The refreshed aggregation fixes the false-new report, but the new significance filter can now hide real per-function regressions and still has no cutoff regression test. I also found unrelated root lockfile churn introduced by the latest commit.

benchmark/profile_benchmark.py
```diff
@@ -0,0 +1,312 @@
...
+    its stddev across runs and source file so the comparator can flag only
+    statistically significant deltas and point at where to read the code.
+    """
+    runs = len(star_runs)
+    names: set[str] = set()
+    for d in star_runs:
+        names.update(d.keys())
+    functions: list[dict[str, Any]] = []
+    for name in names:
+        series = [d[name][0] for d in star_runs if name in d]
```
tamird: A missing function in a sampled pprof run is a zero-time observation, not a reason to drop that run. Filtering it out computes a conditional mean, understates variance, and can inflate an intermittently sampled function (including above 100% of the run-level total), producing false movers/hotspots. Keep the full run count with `d[name][0] if name in d else 0.0` and add a regression where a function appears in only one of several runs.

tamird (commented):
The latest profile-path change reintroduces sampling bias for intermittently observed functions. The earlier significance and remaining root-lockfile findings still apply, and the requested aggregation/comparator regressions are still absent.

MODULE.bazel.lock
tamird: Correction: I compared the tip directly to current main instead of the PR merge base. At 750747b3 the root lockfile is fully restored relative to the merge base, so this finding is addressed.

tamird (commented):
One additional significance issue: main and PR profiles can contain different numbers of usable Starlark runs, so their standard errors cannot share one sample count.

benchmark/compare.py
```diff
@@ -0,0 +1,435 @@
...
+    pr_sf = pr_result.get("starlark_fn")
+    if not isinstance(main_sf, dict) or not isinstance(pr_sf, dict):
+        return ""  # missing or old (pre-stddev) schema -- can't do significance
+
+    main_total = main_sf.get("total", {})
+    pr_total = pr_sf.get("total", {})
+    main_fns = {r["name"]: r for r in main_sf.get("functions", [])}
+    pr_fns = {r["name"]: r for r in pr_sf.get("functions", [])}
+
+    runs = main_total.get("runs") or pr_total.get("runs") or 1
```
tamird: `main_total.runs` and `pr_total.runs` can differ because `profile_benchmark.py` only appends a non-empty decoded profile. Picking one count for both sides miscomputes the combined stderr and can hide or invent significance. Carry `main_n` and `pr_n` separately and compute `sqrt(main_std**2/main_n + pr_std**2/pr_n)`; include an unequal-run-count comparator regression.

benchmark/pprof_decode.py
```diff
@@ -0,0 +1,166 @@
...
+            loc_to_fn[lid] = line.get(1, [0])[0]
+
+    totals: dict[str, float] = defaultdict(float)
+    files: dict[str, str] = {}
+    for chunk in top.get(2, []):
+        loc_ids = _parse_packed_or_singles(chunk, 1)
+        values = _parse_packed_or_singles(chunk, 2)
+        if not loc_ids or val_idx >= len(values):
+            continue
+        leaf = loc_ids[0]
```
tamird, line 151: This attributes every sample to the leaf, including builtins, while the rendered report says builtins are attributed to the caller. In real profiles alias/config_setting/glob therefore dominate the movers table instead of identifying actionable Starlark code. Walk outward through loc_ids to the first non-builtin frame, and cover a builtin-leaf/caller sample.

benchmark/pprof_decode.py
```diff
@@ -0,0 +1,166 @@
...
+        fid = loc_to_fn.get(leaf)
+        if fid is None:
+            continue
+        name = s(fn_name.get(fid))
+        totals[name] += values[val_idx] / divisor
```
tamird, line 156: Keying totals/files only by the short function name merges unrelated functions such as _impl from different .bzl files and keeps whichever filename was seen first. That can create false movers and point reviewers at the wrong source. Preserve normalized (file, function) identity through aggregation/comparison and add a same-name/different-file regression.

benchmark/profile_benchmark.py
```diff
@@ -0,0 +1,316 @@
...
+                     star_path if star_enabled else None, cwd=run_cwd)
+        analysis_us: list[float] = []
+        wall_ms: list[float] = []
+        star_runs: list[dict[str, float]] = []
+        for i in range(args.runs):
+            a, w, star = run_once(command, args.prepare, profile_path,
+                                  star_path if star_enabled else None, cwd=run_cwd)
+            analysis_us.append(a)
+            wall_ms.append(w)
+            if star:
```
tamird, line 340: A valid decoded profile with zero samples is {}, so this drops the entire measured run. The sparse-function zero-fill above still computes a conditional mean, understates variance, and reports the wrong run count whenever one run is fully empty. Append the empty observation whenever Starlark profiling is enabled (and distinguish an actual decode failure), with a fully empty-run regression. The nearby star_runs/run_once annotations also still say dict[str, float], but the decoder returns dict[str, tuple[float, str]].

benchmark/compare.py
```diff
@@ -0,0 +1,441 @@
...
+    pr_fns = {r["name"]: r for r in pr_sf.get("functions", [])}
+
+    main_n = max(main_total.get("runs", 1) or 1, 1)
+    pr_n = max(pr_total.get("runs", 1) or 1, 1)
+
+    main_total_ms = main_total.get("mean_ms", 0.0)
+    pr_total_ms = pr_total.get("mean_ms", 0.0)
+    delta_total = pr_total_ms - main_total_ms
+    pct_total = pct(main_total_ms, pr_total_ms) if main_total_ms else 0.0
+    total_se = _combined_se(main_total.get("stddev_ms", 0.0), main_n,
```
tamird, line 183: total_se is now calculated but never used after removing the total gate. Please delete the dead calculation; the per-function standard errors below are the values that matter.

tamird (commented):
The new revision fixes zero-filled sparse functions, per-function significance, deterministic movers, and unequal sample counts. Three attribution/aggregation gaps remain, and none of the requested statistical regressions were added. Please add focused cases for sparse and fully empty runs, cancelling speedups, zero variance, unequal counts, duplicate function names across files, and builtin caller attribution so later changes cannot silently reintroduce these diagnostics errors.

PR #1317, opened 2026-07-16
refactor(venv): rename venv.bzl to assemble_venv.bzl and split out wheel plan helpers

xangcastle:
This renames `venv.bzl` to `assemble_venv.bzl` to better reflect what the module actually contains, then refactors its internals to consume the `compute_wheel_plan` entry point so wheel lookup tables live alongside the collision resolver. The `py_venv_exec` implementation is split into smaller focused helpers, and the console-script template attribute is pointed at the correct template file.

---

### Test plan


py/private/py_venv/assemble_venv.bzl
```diff
@@ -1,13 +1,13 @@
 """Build-time assembly of a Python virtualenv via ctx.actions.symlink + write.
 
 This module is the single place in rules_py that declares the files making
-up a Python venv. Both `py_binary` / `py_test` (each with its own internal
-venv, unless `expose_venv = True` routes them to a sibling py_venv) and
-the standalone `py_venv` rule call `assemble_venv` to keep their layouts
-bit-identical.
+up a Python venv. Both ``py_binary`` / ``py_test`` (each with its own
```
jbedard, line 4: these all have double-` now?

py/private/py_venv/py_venv.bzl
```diff
@@ -129,49 +177,27 @@ def _common_providers(ctx, shared, executable = None):
...
-            ctx,
-            expand_locations(ctx, v, ctx.attr.data),
-            attribute_name = "env",
-        )
-
-    # `VIRTUAL_ENV` as the venv root's rootpath. `venv.tmpl.sh`
-    # overrides with its own absolute value when invoked directly.
-    passed_env["VIRTUAL_ENV"] = venv_root(shared.venv.bin_python)
-
+    _expand_launcher(ctx, shared)
```
jbedard: What's the reason for doing ones like this? IMO this just makes it more confusing... glancing at `_py_venv_rule_impl` now required scrolling up and down across this file while previously it all fit on one screen?

py/private/py_venv/py_venv.bzl
```diff
@@ -35,34 +30,88 @@ load(":assemble_venv.bzl", "assemble_venv")
 load(":py_venv_exec.bzl", _py_venv_exec = "py_venv_exec")
 load(":types.bzl", "VirtualenvInfo", "venv_root")
 
+_VENV_ONLY_ATTRS = [
```
jbedard: Lets at least leave basic comments for things like this

py/private/py_venv/assemble_venv.bzl
xangcastle: because looks better when //py:defs.doc_extract parse the docstring
```
`foo` → interpreted text
``foo`` → literal/code
```
file names, attrs, function names and inline code should be `literal/code`

py/private/py_venv/py_venv_exec.bzl
```diff
@@ -13,81 +13,89 @@ load("//py/private:py_semantics.bzl", _py_semantics = "semantics")
...
+    Base flags come from ``py_semantics``.  When ``isolated = False``,
+    ``-I`` is stripped from the **base** flags so PYTHONPATH is honored.
+    User ``interpreter_options`` are always appended verbatim — a user
+    explicitly passing ``-I`` survives the isolated-stripping.
+    """
+    base = list(_py_semantics.interpreter_flags)
     if not ctx.attr.isolated:
-        flags = [f for f in flags if f != "-I"]
+        base = [f for f in base if f != "-I"]
+    return base + list(ctx.attr.interpreter_options)
```
tamird, line 89: This changes `isolated = False` behavior: previously `-I` was removed from the combined default and user flags, while the refactor appends a user-supplied `-I` after filtering and silently restores isolation (ignoring `PYTHONPATH`/user site). The standalone/exposed venv helper makes the same change. Please retain combined filtering, or call out and cover the behavior change with a non-isolated/exposed test that passes `interpreter_options = ["-I"]`.

py/private/py_venv/assemble_venv.bzl
```diff
@@ -0,0 +1,360 @@
...
+      struct with ``bin_python`` (File) and ``all_files`` (list[File]).
+    """
+    tc = resolve_venv_toolchain(
+        ctx,
+        py_toolchain = py_toolchain,
+        is_windows = is_windows,
+        venv_name = venv_name,
+    )
+
+    plan = compute_wheel_plan(ctx, _py_library.make_wheels_depset(ctx).to_list())
```
tamird, line 297: `compute_wheel_plan` sorts every wheel path into `wheel_fingerprints`, but this field has no consumer in the repository. Calling it here adds an `O(wheels log wheels)` allocation/sort for every venv on an already hot analysis path. Can the unused fingerprint computation/field be dropped (or the lighter resolver/lookup path retained) until a real consumer needs it?

tamird (commented):
I found one runtime-behavior change and one avoidable hot-path cost in the venv refactor.

py/private/py_venv/py_venv_exec.bzl
tamird: The latest revision restores combined filtering for the standalone/exposed py_venv REPL, which fixes that half. py_venv_exec.bzl still filters only the base flags and then appends user interpreter_options, so a normal py_binary/py_test with isolated=False and interpreter_options=["-I"] still changes behavior. Please restore combined filtering there (and cover the binary/test path) before resolving.

tamird (commented):
The latest simplification restores combined `-I` filtering for the executable venv, but the normal binary/test launcher still appends a user `-I` after filtering and the unused per-venv wheel-fingerprint sort remains. One small rename cleanup: `py/private/py_venv/py_venv.bzl:24` still points readers at the now-deleted `venv.bzl::assemble_venv`; update it to `assemble_venv.bzl`.

tamird:
All test and benchmark steps passed on e403c87. The sole red check is the benchmark comment job: GitHub returned HTTP 503 while listing this PR’s comments. A maintainer will need to rerun the failed job; the substantive review threads are still open.

py/private/py_venv/assemble_venv.bzl
jbedard: Is that the case everywehre or only the `defs.doc_extract`? What about IDEs or BCR?

PR #1322, opened 2026-07-17
feat(py): expose Python provider interop helpers

tamird:
Expose the rules_python PyInfo identity and provider accessors from the public rules_py API so downstream rules can accept mixed Python providers without private load paths.


tamird:
@codex review

xangcastle (approved):

PR #1325, opened 2026-07-17
perf(py): batch private-venv symlinks

tamird:
Batch private-venv top-level symlinks into one action while preserving declared outputs and runfiles. The current benchmark increases analysis cost, so an end-to-end win for large dependency graphs is still required before this is ready to land.

### DIFF
--- a/py/BUILD.bazel
+++ b/py/BUILD.bazel
@@ -43,6 +43,11 @@ label_flag(
     visibility = ["//visibility:public"],
 )
 
+toolchain_type(
+    name = "venv_symlink_toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
 bzl_library(
     name = "current_py_toolchain",
     srcs = ["current_py_toolchain.bzl"],
--- a/py/private/py_venv/templates/symlinks.py
+++ b/py/private/py_venv/templates/symlinks.py
@@ -1,0 +1,18 @@
+"""Materialize declared venv site-packages symlinks from a parameter file."""
+
+import os
+import sys
+from pathlib import Path
+
+
+def main(params):
+    entries = Path(params).read_text().splitlines()
+    if len(entries) % 2:
+        raise ValueError("expected output/target pairs")
+    for output, target in zip(entries[::2], entries[1::2]):
+        Path(output).parent.mkdir(parents=True, exist_ok=True)
+        os.symlink(target, output)
+
+
+if __name__ == "__main__":
+    main(sys.argv[1])
--- a/py/private/py_venv/venv.bzl
+++ b/py/private/py_venv/venv.bzl
@@ -31,12 +31,12 @@ distutils, etc.) treat it as a real venv:
 
 The whole tree is declared at analysis time as individual
 `ctx.actions.declare_file` / `ctx.actions.declare_symlink` outputs (no
-tree-artifact + remote-exec materialisation surprises). Private non-Windows
-venvs can materialize projected site-packages symlinks in one action.
+tree-artifact + remote-exec materialisation surprises). Private venvs can
+materialize projected site-packages symlinks in one action on POSIX executors.
 """
 
 load("//py/private:py_library.bzl", _py_library = "py_library_utils")
-load("//py/private/toolchain:types.bzl", "EXEC_TOOLS_TOOLCHAIN")
+load("//py/private/toolchain:types.bzl", "EXEC_TOOLS_TOOLCHAIN", "VENV_SYMLINK_TOOLCHAIN")
 load(":toolchains_resolver.bzl", "resolve_venv_toolchain")
 load(":virtuals_resolvers.bzl", "enforce_collision_policy", "resolve_wheel_collisions")
 
@@ -57,7 +57,7 @@ def assemble_venv(
         venv_activate_tmpl,
         virtualenv_shim_py,
         site_merge_script_py,
-        venv_symlinks_script_py,
+        venv_symlink_tool,
         console_script_tmpl,
         venv_name):
     """Declare every file + symlink that makes up a venv for a target.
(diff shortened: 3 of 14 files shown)


tamird:
@codex review

py/private/py_venv/templates/symlinks.py
```diff
@@ -0,0 +1,18 @@
+"""Materialize declared venv site-packages symlinks from a parameter file."""
+
+import os
+import sys
+from pathlib import Path
+
+
+def main(params):
```
zbarsky-openai, line 8: plz rewrite to fast language. or at least make this a toolchain so we can override it on our end!

py/private/py_venv/templates/symlinks.py
```diff
@@ -0,0 +1,18 @@
...
+    entries = Path(params).read_text().splitlines()
+    if len(entries) % 2:
+        raise ValueError("expected output/target pairs")
+    for output, target in zip(entries[::2], entries[1::2]):
```
zbarsky-openai, line 12: how much did this help performance, what did you. measure

tamird: I measured the sa-server action graph: total dependency actions dropped from 25,939 to 24,153, and UnresolvedSymlink actions from 1,663 to 42. I have not measured an end-to-end wall-time win; the small upstream benchmark is effectively unchanged, as expected at that scale.

py/private/py_venv/templates/symlinks.py
tamird: Done. The helper is now selected through the public //py:venv_symlink_toolchain_type, so downstream can register a faster implementation (or an empty toolchain to retain native symlink actions). The custom-tool regression also verifies its executable and runfiles are action inputs.

py/private/py_venv/venv.bzl
```diff
@@ -143,6 +146,15 @@ def assemble_venv(
     # a cached per-wheel prefix rather than formatting every component.
     site_packages_prefix = site_packages_rel + "/"
     target_prefix_by_sp = {}
+    exec_runtime = None
+    if not is_windows and venv_symlinks_script_py != None:
```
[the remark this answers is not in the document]
tamird: Fixed. The default batching toolchains are now constrained by execution platform, with a final empty fallback for unsupported executors. A Windows executor therefore keeps Bazel's native symlink actions even for a non-Windows target; the new cross-execution analysis regression covers that case and the inverse Windows-target/POSIX-executor case.

tamird:
@codex review

py/private/py_venv/templates/symlinks.py
```diff
@@ -0,0 +1,18 @@
+"""Materialize declared venv site-packages symlinks from a parameter file."""
```
jbedard, line 1: I don't think this file is a template so it should live 1 directory up?

py/private/py_venv/templates/symlinks.py
jbedard: That sounds insignificant and not worth the code. Can you provide better benchmarks for your repo?

Implementing the toolchain with something other then python within your monorepo might improve it more to help justify the extra code within rules_py, like @zbarsky-openai was suggesting (however I wouldn't want to introduce that within rules_py at this point).

py/private/py_venv/venv.bzl
```diff
@@ -143,6 +146,11 @@ def assemble_venv(
     # a cached per-wheel prefix rather than formatting every component.
     site_packages_prefix = site_packages_rel + "/"
     target_prefix_by_sp = {}
+    symlink_outputs = []
```
jbedard, line 149: Should we disable this optimization when `len(top_level_to_site_pkgs) < X`? I would think in many scenarios launching another process (especially python) will be slower then a couple dozen `ctx.actions.symlink` calls?

py/BUILD.bazel
```diff
@@ -43,6 +43,11 @@ label_flag(
     visibility = ["//visibility:public"],
 )
 
+toolchain_type(
```
jbedard, line 46: I think I would vote to keep this private or at least put an `experimental_` prefix on it since 99% of the time we wouldn't want people to touch it or see it.

@xangcastle WDYT?

py/private/py_venv/templates/symlinks.py
xangcastle: it is a template @jbedard, is in the right place for me

jbedard: It doesn't look like a template to me...? It's directly referenced via `py_binary(main)`...

py/private/py_venv/templates/symlinks.py
```diff
@@ -0,0 +1,18 @@
...
+import sys
+from pathlib import Path
+
+
+def main(params):
+    entries = Path(params).read_text().splitlines()
+    if len(entries) % 2:
+        raise ValueError("expected output/target pairs")
+    for output, target in zip(entries[::2], entries[1::2]):
+        Path(output).parent.mkdir(parents=True, exist_ok=True)
```
jbedard, line 13: Isn't this potentially a lot of repeated `mkdir(parents=True)` calls? Doesn't that do a lot of system dir-exists checks that might be slow?

py/private/py_venv/templates/symlinks.py
tamird: I rechecked the measurement provenance. Those sa-server numbers came from an older prototype that batched both symlinks and console-script writes, not this symlink-only PR, so they do not establish a win here. The current upstream benchmark also shows about a 6% analysis regression after the overridable-toolchain plumbing (10.57s to 11.24s, with 145 additional configured targets). I am measuring the exact sa-server target before adding more machinery; if that does not show a material improvement, this approach is not worth carrying.

tamird:
I rechecked the measurement provenance and could not substantiate this optimization. The reported sa-server action reduction came from an older prototype that also batched console-script writes, not this symlink-only change. The reproducible upstream analysis benchmark regresses from 10.57s to 11.24s (about 6%) and adds 145 configured targets. I attempted to measure the exact sa-server debug venv, but this sparse checkout cannot analyze its production closure without substantial source hydration.

There is no demonstrated benefit that justifies the extra toolchain, Python action, or threshold/API complexity, so I am closing this in favor of measured runtime reductions.

PR #1327, opened 2026-07-17
feat(uv): filter installed wheel content

tamird:
Add a site-packages-relative exclude_glob override so unnecessary test,
source, and proto payload can be removed from installed wheels before
bytecode compilation. Keep wheel topology and post-install patch
validation aligned with the retained files.

Post-install patches and exclusions rebuild RECORD from the retained
install tree, so patched files cannot leave stale hashes.


tamird:
@codex review

uv/private/whl_install/rule.bzl
```diff
@@ -88,27 +89,32 @@ def _whl_install(ctx):
         exec_runtime.files,
     ]
 
-    # Patch application (happens before pyc compilation).
+    preserve_paths = {path: None for path in top_levels}
```
jbedard: This was hoisted out of the `if patch_files` but is never used outside it? Why hoist?

uv/private/whl_install/rule.bzl
```diff
@@ -219,6 +262,10 @@ lighter weight since the toolchain's files aren't inputs.
             default = 0,
             doc = "Strip count for patches (-p flag).",
         ),
+        "exclude_glob": attr.string_list(
+            default = [],
```
jbedard, line 246: Is there a common pattern that we should use as a default?

Are there any common patterns safe to exclude by default in the python ecosystem? Is it worth excluding `{README,CHANGELOG}.md` by default? etc...

jbedard: For example rules_js has a default exclude list: https://github.com/aspect-build/rules_js/blob/v3.2.3/npm/private/exclude_package_contents_presets.bzl#L5

tamird: I would keep the default empty. The downstream wins are package-specific (tests, bundled sources, and protos), and even README/CHANGELOG or other data files can be runtime resources; silently changing every wheel's installed surface and RECORD is a compatibility risk. An opt-in preset can be added later if real shared patterns emerge.

uv/private/whl_install/rule.bzl
tamird: Moved it back under the patch guard in 12ec4faa; the unconditional work is gone.

tamird:
@codex review

uv/private/whl_install/rule.bzl
```diff
@@ -120,14 +124,43 @@ def _whl_install(ctx):
...
 
+    if ctx.attr.exclude_glob:
+        filter_script = ctx.file._filter_script
+        filter_arguments = ctx.actions.args()
+        filter_arguments.add(filter_script)
+        filter_arguments.add_all([unpack_dir], expand_directories = False, before_each = "--from")
+        filter_arguments.add_all([install_dir], expand_directories = False, before_each = "--into")
+        filter_arguments.add("--python-version-major", py_toolchain.interpreter_version_info.major)
+        filter_arguments.add("--python-version-minor", py_toolchain.interpreter_version_info.minor)
+        filter_arguments.add_all(ctx.attr.exclude_glob, before_each = "--exclude-glob")
```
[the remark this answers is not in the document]
tamird: Encoded both action argument lists as --exclude-glob=<value> and added real leading-dash patch/filter coverage.

py/tools/unpack/filter.py
```diff
@@ -0,0 +1,79 @@
...
+    site_packages = args.into / site_packages_relative
+
+    for record in site_packages.glob("*.dist-info/RECORD"):
+        with record.open(newline="", encoding="utf-8") as stream:
+            rows = [
+                row
+                for row in csv.reader(stream)
+                if not row or not excluded(tuple(row[0].split("/")), args.exclude_glob)
+            ]
+        record.unlink()
```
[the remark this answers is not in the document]
tamird: Made only the copied RECORD writable before replacement; the read-only backing tree stays unchanged and is covered.

uv/private/whl_install/repository.bzl
```diff
@@ -853,6 +888,12 @@ filegroup(
             strip = post_install_patch_strip,
         )
 
+    if exclude_glob:
+        install_attrs += """
+    exclude_glob = {exclude_glob},""".format(
+            exclude_glob = indent(pprint(exclude_glob), " " * 4).lstrip(),
```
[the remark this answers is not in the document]
tamird: Modified wheels now index the actual install tree while untouched wheels retain the raw-wheel fast path; added installed-tree, top-level-removal, and .pyd coverage.

jbedard (approved):

tamird:
@codex review

tamird:
The rebased head passes the root, e2e, snapshot, and wheel-filter suites. The sole CI failure is `examples/py_venv` on Bazel 8 (`//:test_internal_venv`); Bazel 9 passes, and I reran that exact Bazel-8 target locally with `--test_output=errors` and it passes. The GitHub integration cannot retry the job (403), so could a maintainer rerun the failed job?

tamird:
Holding this as draft while I reduce the implementation. The concrete downstream exclusions remove NumPy tests, Temporal sdk-core sources/protos, and google proto data; none removes an import root. Gazelle can keep its existing index if filtering explicitly preserves top-level import roots, avoiding the mixed-wheel/sdist fallback problem and the extra sdist downloads/builds. I’ll rework the cumulative change around that narrower contract before requesting another review.

uv/private/whl_install/repository.bzl
```diff
@@ -640,10 +673,19 @@ py_library(
...
         )
 
-    if prebuilds:
+    post_install_patches = json.decode(repository_ctx.attr.post_install_patches) if repository_ctx.attr.post_install_patches else []
+    post_install_patch_strip = repository_ctx.attr.post_install_patch_strip
+    exclude_glob = repository_ctx.attr.exclude_glob
+
+    if post_install_patches or exclude_glob:
+        gazelle_index_whl = ":actual_install"
+        gazelle_index_output_group = "    output_group = \"install_dir\",\n"
```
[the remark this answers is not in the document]
tamird: Fixed in f154aaf391ae. Gazelle again indexes a locked archive independently of host wheel compatibility; the installed tree is no longer selected during manifest generation.

uv/private/gazelle_manifest/generate.py
```diff
@@ -48,37 +48,38 @@ def normalize_name(name):
...
-            parser = Parser()
-            msg = parser.parse(StringIO(metadata_content))
-
-            package_name = msg.get('Name')
-            if not package_name:
-                print(f"Warning: 'Name' field missing from METADATA in {whl_path}", file=sys.stderr)
-                return None
+        if whl_path.is_dir():
+            metadata_files = list(whl_path.glob("lib/python*/site-packages/*.dist-info/METADATA"))
+            metadata_content = metadata_files[0].read_text(encoding="utf-8") if metadata_files else None
```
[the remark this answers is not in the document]
tamird: Fixed in f154aaf391ae. Gazelle reads package metadata from the original archive, and exclusions that remove the complete dist-info directory or RECORD are rejected before installation.

uv/private/whl_install/repository.bzl
```diff
@@ -740,10 +779,15 @@ filegroup(
...
         whl_file_label = whl_file_labels[target]
-        whl_name, tls, regular, css, ns_entries, dirs_set, init_dirs, native_segments = _extract_wheel_metadata(
+        whl_name, tls, regular, css, ns_entries, dirs_set, init_dirs, native_segments, retained_paths = _extract_wheel_metadata(
             repository_ctx,
             whl_file_label,
+            exclude_glob,
         )
+        if exclude_glob:
+            for path in retained_paths:
+                if path.endswith((".py", ".so", ".dylib", ".pyd")):
```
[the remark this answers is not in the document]
tamird: Fixed in f154aaf391ae. The tuple-style Starlark endswith path is gone; filtering now happens in the single WhlInstall action.

uv/private/whl_install/repository.bzl
```diff
@@ -640,7 +675,13 @@ py_library(
 """,
         )
 
-    if prebuilds:
+    post_install_patches = json.decode(repository_ctx.attr.post_install_patches) if repository_ctx.attr.post_install_patches else []
+    post_install_patch_strip = repository_ctx.attr.post_install_patch_strip
+    exclude_glob = repository_ctx.attr.exclude_glob
+
+    if exclude_glob and prebuilds:
```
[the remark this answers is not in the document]
tamird: Addressed in f154aaf391ae by making Gazelle’s raw-wheel index explicitly an unfiltered superset. Exclusions cannot remove an original top-level import root, so source-only installs retain a resolvable package mapping; the remaining shared-namespace precision limitation is documented and covered by the source-fallback fixture.

uv/private/whl_install/repository.bzl
```diff
@@ -804,6 +852,19 @@ filegroup(
         if css:
             console_scripts_by_whl[whl_name] = sorted(css.values())
 
+    if exclude_glob:
+        gazelle_index = {"name": repository_ctx.attr.package_name}
+        if prebuilds:
+            gazelle_index["paths"] = sorted(gazelle_paths.keys())
```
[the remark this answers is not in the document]
tamird: Fixed in f154aaf391ae. Raw archive indexing covers representation differences such as backend.so versus backend.py without selecting the host install, and the restored mixed/source-fallback Gazelle fixture exercises the retained mapping. The install action validates original-root topology before filtering and compilation.

tamird:
@codex review

py/tools/unpack/unpack.py
```diff
@@ -297,15 +316,66 @@ def main():
...
+
+        # File-shaped source globs do not match their PEP 3147 or legacy
+        # caches. Remove shipped bytecode whose corresponding source was
+        # excluded before compiling the retained sources.
+        for bytecode in site_packages.rglob("*.pyc"):
+            if bytecode.parent.name == "__pycache__":
+                try:
+                    source_path = Path(importlib.util.source_from_cache(str(bytecode)))
+                except ValueError:
+                    continue
```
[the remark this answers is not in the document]
tamird: Fixed in fb65ab2c. Cache-to-source matching now parses the PEP 3147 suffix from the right, so dotted source names and their shipped bytecode are excluded consistently during unpack, RECORD reconstruction, and topology calculation. The new regression covers the dotted source/cache pair and the rebuilt RECORD.

tamird:
@codex review

py/tools/unpack/unpack.py
```diff
@@ -297,15 +315,70 @@ def main():
...
             raise SystemExit(
                 "Post-install patch changed observed package classification: {}".format(relative)
             )
-        if _native_descendants(directory) != native_descendants:
+        if _native_descendants(directory, site_packages, args.exclude_glob) != native_descendants:
             raise SystemExit(
                 "Post-install patch changed observed native files: {}".format(relative)
             )
 
+    if args.exclude_glob:
```
jbedard, line 498: Aren't we unpacking a zip/tar file? Shouldn't exclusions be done at unpack time to avoid writing the file simply to delete it later?

tamird: Yes, for unpatched wheels. Excluded site-packages payload is now skipped before read/write. The late sweep remains for the documented install → patch → exclude ordering and for metadata consumed by entry-point generation.

tamird:
@codex review

py/tools/unpack/unpack.py
```diff
@@ -297,15 +370,55 @@ def main():
...
+        for path in sorted(site_packages.rglob("*"), reverse=True):
```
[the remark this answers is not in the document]
tamird: Fixed in 5bd6537b. The existing exclusion sweep now runs again after compileall, so generated bytecode and cache paths are absent from the final tree. The regression retains `__init__.*.pyc` to prove compilation ran while excluding generated `mod.*.pyc`.

tamird:
@codex review

tamird:
I checked the current OpenAI annotations for the default-excludes question. We do not have a repo-wide preset: the existing filters are package-specific—NumPy removes its bundled tests, googleapis-common-protos removes `.proto` data, and Temporal removes the bridge SDK-core payload. I would keep this opt-in; broad test/proto/source defaults can change legitimate runtime behavior, and the useful patterns depend on the package's layout.

jbedard:
@tamird rebase and @codex review

tamird:
@codex review

jbedard:
Can we trust the RECORD data if the files are not patched?

tamird:
No. Verifying the downloaded wheel hash only proves archive identity; it does not prove that an individual RECORD row matches the archived bytes. A syntactically valid, same-size stale digest survives unchanged even without post-install patches. #1354 computes each installed-file digest from the streamed bytes and has an executed cowsay regression for the patched case, so correctness does not depend on trusting RECORD.

py/tools/unpack/exclude_glob.py
```diff
@@ -0,0 +1,68 @@
+"""Match site-packages-relative wheel exclusion globs."""
+
+import argparse
+
+
+def parse(value):
```
jbedard: **Sync hazard (main maintainability risk):** this matcher/parser is a hand-written reimplementation of the same semantics in `uv/private/whl_install/repository.bzl` (`parse_exclude_glob` / `exclude_glob_matches` / `record_path_excluded`), and the `.pyc`→source mapper here (`_path_excluded`) mirrors `record_path_excluded` there. If they ever drift, the analysis-time RECORD topology silently mismatches the on-disk tree. The shared *behavior* is tested, but the vector tables in `unpack_test.py` and `test.bzl` are copy-pasted, not shared. Suggest a cross-reference comment in each ("mirror of X — keep in sync") and ideally a single shared vector table both suites consume.

uv/private/whl_install/repository.bzl
```diff
@@ -107,6 +107,80 @@ def native_roots_for_segments(segments, collision_roots = ()):
...
+    if parts[-1]:
+        value = value[:-len(parts[-1])]
+    for part in parts[1:-1]:
+        index = value.find(part)
+        if index < 0:
+            return False
+        value = value[index + len(part):]
+    return True
+
+def exclude_glob_matches(path, pattern):
```
jbedard, line 143: Mirror of `py/tools/unpack/exclude_glob.py` — worth a cross-reference comment (see note there).

Minor asymmetry: Python's `excluded()` has a `path[0] not in ("", ".", "..")` guard that this function lacks. I traced it and it's **benign** — `..`-leading (data) RECORD paths are dropped by the pre-existing `first_segment in (".", "..")` filter before topology derivation, and on disk they live outside `site_packages` so `rglob` never reaches them. No fix needed, but a one-line note here would save a future reader the trace.

jbedard (commented):
Reviewed the `exclude_glob` change end-to-end. The hard parts are correct: the import-root safety net, the `.pyc`↔source coupling (incl. `opt-N`/dotted stems), the directory-removal ordering (trailing `**` closure + reverse sort keeps `rmdir` on empty dirs only), and the from-disk RECORD regeneration. Test coverage is excellent. A few non-blocking issues/nits inline.

Additional notes not tied to a single line:

- **Performance (opt-in only, so low concern):** when active it adds a full-tree RECORD rehash plus up to two extra `rglob` sweeps (`_remove_excluded` ×2, `_import_roots`). For the large wheels this targets (docs cite `numpy`) that's a full extra hash of the retained tree — acceptable, just flagging.

- **Style nits:** in `unpack.py` the edits leave a single blank line before `def main()` and before the `for ep_path` loop (PEP8 prefers two before a top-level def). Trivial.

py/tools/unpack/unpack.py
```diff
@@ -351,15 +457,41 @@ def main():
...
+            raise SystemExit(
+                "wheel exclusions removed top-level import roots: {}".format(
+                    ", ".join(sorted(removed_roots))
+                )
+            )
+
+    records = list(site_packages.glob("*.dist-info/RECORD"))
+    if len(records) > 1 or (args.exclude_glob and not records):
+        raise SystemExit("expected exactly one installed RECORD, found {}".format(len(records)))
+    if records and (args.patches or args.exclude_glob):
```
jbedard, line 514: **Behavior change for patches-only installs (please confirm intent):** this block fires on `args.patches or args.exclude_glob`, so a post-install-patched wheel with no exclusions now gets a from-disk RECORD rewrite. Pre-PR, patches-only never rewrote RECORD, so patched files left stale hashes. This is *more* correct (and the new `uv-sdist-fallback` RECORD assertions cover it), but it's a behavior change worth calling out in the PR description.

Separately: the guard blocks removing `RECORD` / the whole `.dist-info`, but not `METADATA`/`WHEEL`. A glob like `*.dist-info/METADATA` would pass validation and produce a superficially-valid install that breaks `importlib.metadata`. Low priority (user error, and docs say preserve roots), but a cheap guard or a doc line naming the protected metadata files would harden it.

py/tools/unpack/unpack.py
```diff
@@ -58,14 +58,99 @@ def _is_native_library(path):
...
+
+def _installer_input(path):
+    return (
+        len(path.parts) == 2
+        and path.parts[0].endswith(".dist-info")
+        and path.name in ("entry_points.txt", "RECORD")
+    )
+
+
+def _remove_excluded(site_packages, patterns):
```
jbedard: Minor: this cleans empty `__pycache__` dirs, but excluding files by pattern (e.g. `demo/file_tests/test_*.py`) can leave an empty non-`__pycache__` directory behind (`file_tests/`). Harmless — not in RECORD, not an import root — just untidy. Could generalize the empty-dir sweep, or leave as-is.

py/tools/unpack/exclude_glob.py
tamird: Done. Added reciprocal sync comments and shared valid matcher/cache-to-source vectors consumed by both test suites. Invalid patterns remain Python-only because Starlark fail is not catchable in that test harness.

uv/private/whl_install/repository.bzl
tamird: Done. The reciprocal comment now notes the intentional asymmetry: escaping RECORD paths are discarded before topology derivation, while the install sweep stays within site-packages.

py/tools/unpack/unpack.py
tamird: Confirmed and documented. Post-install patches and exclusions rebuild RECORD from the retained install tree, so patched files cannot leave stale hashes. Exclusions now require exactly one installed RECORD and its sibling METADATA, with regression coverage and documented constraints.

py/tools/unpack/unpack.py
tamird: Left unrelated empty directories intact: removing them requires another traversal and changes file-glob/namespace semantics. Empty __pycache__ directories are still removed.

tamird:
@codex review

jbedard:
@tamird rebase to resolve conflicts

tamird:
Rebased onto main and resolved the unpack conflict. Exclusions still skip payload early, sweep after patches and compilation, preserve import roots and topology checks, and rebuild the single retained RECORD with fresh hashes before compileall. The new helpers and tests now satisfy the Python 3.9 type checks.

@codex review

e2e/cases/uv-exclusions/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
+load("@aspect_rules_py//py:defs.bzl", "py_test")
+load("@aspect_rules_py//uv:defs.bzl", "gazelle_python_manifest")
+
+gazelle_python_manifest(
```
jbedard: Can we avoid gazelle in these e2e tests? Is using uv an option instead?

py/tools/unpack/BUILD.bazel
```diff
@@ -1,7 +1,12 @@
+load("@bazel_lib//:bzl_library.bzl", "bzl_library")
 load("//py:defs.bzl", "py_test")
 
 exports_files(
-    ["unpack.py"],
+    [
+        "exclude_glob.py",
+        "exclude_glob_test_vectors.bzl",
```
jbedard: Why does this test file need to be exported?

e2e/cases/uv-exclusions/BUILD.bazel
tamird: Done. Removed Gazelle and the manifest-only assertions from the exclusions e2e fixture; the test now exercises the uv install directly.

py/tools/unpack/BUILD.bazel
tamird: Done. Removed the obsolete exports_files and bzl_library declarations for the shared test vectors.

tamird:
@codex review

tamird:
@codex review

jbedard:
@tamird rebase

tamird:
@codex review

— tamirdex

jbedard:
@tamird see CI typecheck errors

tamird:
Fixed in 2da2006eed33. _import_root now narrows the non-empty parts tuple before indexing; Pyright, ty, Ruff, the focused unpack/e2e cases, and the cumulative review are clean.

@codex review

— tamirdex

py/tools/unpack/unpack.py
```diff
@@ -59,14 +59,113 @@ def _is_native_library(path: Path) -> bool:
...
+    from exclude_glob import excluded
+
+    # Keep cache-to-source matching in sync with record_path_excluded in
+    # uv/private/whl_install/repository.bzl and the shared test vectors.
+    if excluded(path.parts, patterns):
+        return True
+    if not is_file or not path.name.endswith(".pyc"):
+        return False
+    if path.parent.name == "__pycache__":
+        source, separator, tag = path.stem.rpartition(".")
```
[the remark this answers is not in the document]
tamird: Fixed at `1ed8cd32e5a2`: the installer now strips `.pyc` literally, matching RECORD-derived topology on Python 3.14. Added a namespace-wheel regression that retains another member and the malformed cache in both the installed tree and rebuilt RECORD.

— tamirdex

tamird:
@codex review

— tamirdex

py/tools/unpack/BUILD.bazel
```diff
@@ -10,7 +16,11 @@ py_test(
     size = "small",
     srcs = ["unpack_test.py"],
     args = ["$(rootpath :unpack.py)"],
-    data = [":unpack.py"],
+    data = [
+        ":exclude_glob.py",
+        ":exclude_glob_test_vectors.bzl",
```
jbedard, line 21: Why is a `.bzl` required for a `py_test`?

py/tools/unpack/unpack.py
```diff
@@ -216,7 +331,6 @@ def install_wheel(
                 if reusable_record is not None and reusable_record[1] != str(info.file_size):
                     reusable_record = None
                 installed[dest] = reusable_record
-
```
jbedard, line 219: Why remove this line?

jbedard:
@tamird rebase and review PR comments

py/tools/unpack/BUILD.bazel
tamird: The Python `unpack_test` executes the same literal exclusion and cached-source vectors as the Starlark topology tests, so the `.bzl` must be in `data` for matcher/cache-mapping parity.

— tamirdex

py/tools/unpack/unpack.py
tamird: Restored the separator in `185e04d78b03`.

— tamirdex

tamird:
@codex review

— tamirdex

PR #1329, opened 2026-07-17
chore: remove unnecessary use of rules_python

jbedard:
### Test plan


tamird (approved):
Checked the cumulative dependency removals and affected workspaces: the removed direct rules_python dependencies have no remaining consumers, and the required runtime/API dependency remains available transitively. No upgrade or interop blockers found.

xangcastle (approved):

PR #1331, opened 2026-07-17
feat(uv): apply package overrides across locks

### LINKED ISSUE #1330: Apply package overrides across multiple uv locks
When a module declares multiple uv.project locks, the same post-install patch or package exclusion currently requires a separate uv.override_package call for every lock because lock is mandatory and singular. This is easy to miss as new locks are added and repeats identical policy.

Could override_package support a selected set of locks, or an all-locks form for modification-only overrides, applying the override wherever the package/version is present? Full target replacement can remain explicitly lock-scoped. This would also make repeated exclude_glob overrides easier to keep consistent once wheel filtering lands.

tamird:
Allow a modification-only package override to omit lock and apply across every uv project declared by the same module, so shared policy cannot be missed when locks are added. Skip locks without the selected package while rejecting zero matches, ambiguous versions, duplicate overrides, and invalid declarations.

Fixes #1330.


tamird:
@codex review

uv/private/extension/defs.bzl
```diff
@@ -825,7 +845,11 @@ _annotations_tag = tag_class(
 
 _override_package_tag = tag_class(
     attrs = {
-        "lock": attr.label(mandatory = True),
+        "lock": attr.label(mandatory = False),
+        "all_locks": attr.bool(
```
jbedard: Can this not be implied simply if `lock == None`?

jbedard: @xangcastle WDYT?

jbedard (approved):

uv/private/extension/defs.bzl
tamird: Yes. I dropped all_locks; omitting lock now applies a modification across the declaring module's project locks, while target still requires an explicit lock. Updated the docs and the multi-lock, version, duplicate, and failure coverage.

tamird:
@codex review

xangcastle (approved):

PR #1333, opened 2026-07-17
feat(py): expose PyWheelsInfo

tamird:
Expose installed-wheel metadata through the public Python API so custom rules can forward it without a private load path. Preserve postorder wheel aggregation and cover forwarding with a pinned UV wheel.


tamird:
@codex review

tamird:
@jbedard this is the independent PyWheelsInfo follow-up from #1231. It keeps the provider boundary small, adds a real UV-wheel forwarding regression, and is green/Codex-clean.

jbedard (approved):

PR #313, opened 2024-03-19
fix: handle a system interpreter path on the toolchain

mattem:
Handles the path being given to the Python toolchain being a system interpreter.

Closes #302

---

### Type of change

- Bug fix (change which fixes an issue)

**For changes visible to end-users**

- Suggested release notes are provided below:
Correctly handle a system interpreter path being set on a Python toolchain.

### Test plan

- Covered by existing test cases
- Manual testing; please provide instructions so we can reproduce:

Manual testing with local interpreter.

Issue #302, opened 2024-03-08
[Bug]: "unbound variable" error in py_test, py_binary when using py_runtime with interpreter_path set


wade-arista:
### What happened?

I declared a custom toolchain that points at the system python interpreter.

```starlark
load("@rules_python//python:defs.bzl", "py_runtime", "py_runtime_pair")

py_runtime(
    name = "my_py3_runtime",
    interpreter_path = "/usr/bin/python",
    interpreter_version_info = {
        "major": "3",
        "minor": "10",
        "micro": "8",
    },
    python_version = "PY3",
)

py_runtime_pair(
    name = "my_py_runtime_pair",
    py2_runtime = None,
    py3_runtime = ":my_py3_runtime",
)

toolchain(
    name = "default",
    toolchain = ":my_py_runtime_pair",
    toolchain_type = "@bazel_tools//tools/python:toolchain_type",
)
```

I registered it in MODULE.bazel:
```starlark
register_toolchains("//toolchain:default")
```

When I tried to run a `py_test`, the sh wrapper failed:
```console
$ bazel test --action_env=RUNFILES_LIB_DEBUG=1 src_test
==================== Test output for //:src_test:
INFO[runfiles.bash]: rlocation(aspect_rules_py~override/py/tools/venv_bin/venv_bin): start
INFO[runfiles.bash]: runfiles_current_repository(2): caller's path is (/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/src_test.runfiles/_main/src_test)
INFO[runfiles.bash]: runfiles_current_repository(2): (/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/src_test.runfiles/_main/src_test) has path (_main/src_test) relative to the runfiles directory (/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/src_test.runfiles)
INFO[runfiles.bash]: runfiles_current_repository(2): (/tmp/bazel-working-directory/_main/bazel-out/k8-fastbuild/bin/src_test.runfiles/_main/src_test) corresponds to rlocation path (_main/src_test)
INFO[runfiles.bash]: runfiles_current_repository(2): (_main/src_test) lies in the main repository
[...]

PR #1334, opened 2026-07-17
refactor!: require an in-build interpreter, drop system-interpreter support

jbedard:
rules_py rules now require the resolved py3_runtime to provide an in-build `interpreter` file. System interpreters — a py_runtime declared with `interpreter_path` (an absolute host path) rather than a file — are no longer supported and now fail with a clear error at analysis time.

This removes the non-hermetic `runfiles_interpreter = False` codepath that threaded through toolchain resolution and venv assembly:

- py_semantics: fail when py3_runtime has no `interpreter` file; drop the `runfiles_interpreter` field (now always true).
- toolchains_resolver: collapse the system-interpreter branches in _pyvenv_home, _bin_python_target_path, and _uses_empty_venv_home.
- current_py_toolchain: $(PYTHON3) requires the interpreter file.

File-based runtimes registered via rules_python's py_runtime / py_runtime_pair remain usable and are still re-exported from //py:defs.bzl.

BREAKING CHANGE: py_runtime(interpreter_path = ...) toolchains are no longer accepted by rules_py rules. Register an in-build interpreter toolchain (e.g. via interpreters.toolchain()) instead.

### Test plan


xangcastle (approved):

jbedard:
@tamird is `interpreter_path` important for you?

tamird:
It isn’t used by the current OpenAI integration: we register the rules_py/PBS toolchains, and the custom FIPS runtime supplies an `interpreter` target, not `interpreter_path`.

I deliberately kept explicit `py_runtime(interpreter_path = ...)` as the escape hatch when removing `interpreters.local()`, though; container toolchains are a real downstream use case. If dropping support is intentional, please keep an expected-analysis-failure test for the new diagnostic—the PR currently removes the only system-runtime fixture.

jbedard:
> container toolchains are a real downstream use case

Can you give an example where this change will prevent that?

tamird:
The concrete rules_py case is the previously reported [Arista repro](https://github.com/wade-arista/bazel-demo/tree/23ff2cebe6058a5e99535bd3f76c72b1b2b188ce): it registers `py_runtime(interpreter_path = "/usr/bin/python", interpreter_version_info = ...)` as the standard Python toolchain and invokes rules_py's `py_test` ([#302](https://github.com/aspect-build/rules_py/issues/302), fixed by [#313](https://github.com/aspect-build/rules_py/pull/313)). This PR makes that fail analysis.

The same target-runtime pattern is what a binary relying on Python supplied by a base image needs; an exec-tools interpreter is separate and does not help. I don't have evidence of a current container consumer that would be blocked—the public `resf/distro-tools` example uses rules_python for its image binary, so it is unaffected.

If the breaking change is intentional, please retain an expected-analysis-failure case for the diagnostic; the existing interpreter-path regression was added as recently as [#1087](https://github.com/aspect-build/rules_py/pull/1087).

tamird:
Thanks, this covers `current_py_toolchain`. Could the fixture also build a `py_venv`? The new guard in `py_semantics`—used by the venv/PEX/wheel rules and replacing the deleted system-runtime venv regression—never runs here, so this test would still pass if that diagnostic regressed.

jbedard:
@tamird more tests added - anything else missing?

tamird (approved):
Nothing else missing. This now exercises both target-runtime rejection paths on Bazel 8 and 9 while keeping the exec-tools runtime separate; the file-based interop coverage is green too. Thanks.

tamird:
Nothing else missing. I verified both target-runtime rejection paths on Bazel 8 and 9, the separate exec-tools runtime, and the file-based interop coverage; approved the current head.

tamird:
No. Our registered Python toolchains provide an in-build interpreter; the custom FIPS runtime also uses `interpreter`, not `interpreter_path`.

PR #1338, opened 2026-07-18
perf(unpack): compute RECORD hashes inline instead of re-reading each file

jbedard:
### Test plan

- New test cases added


jbedard:
#1350 and #1351 did this better

PR #1344, opened 2026-07-19
fix(py): stamp py_pex_binary interpreter constraints from the binary's own interpreter

jbedard:

 No API changes. py_pex_binary no longer resolves a Python toolchain itself.
 - Interpreter constraints now track the binary's own interpreter version. The {major}/{minor}/{patch} placeholders in python_interpreter_constraints are substituted from the version of the packaged binary (its python_version), not the py_pex_binary rule's toolchain. Previously a binary pinned to e.g. 3.13 in a 3.11 workspace produced a PEX advertising 3.11 — a version it wasn't built for. No action needed; PEXes built from version-mismatched binaries now advertise the correct interpreter.
 - The brittle repo-name exclusion list (bzlmod vs. WORKSPACE, + vs. ~ separators) is gone, so interpreter/venv exclusion is robust across Bazel versions and setups.

### Test plan

- New test cases added

### DIFF
--- a/py/private/py_pex_binary.bzl
+++ b/py/private/py_pex_binary.bzl
@@ -19,9 +19,10 @@ py_pex_binary(
 """
 
 load("@bazel_lib//lib:paths.bzl", "to_rlocation_path")
+load("//py/private:providers.bzl", "PyWheelsInfo")
 load("//py/private:py_info.bzl", "PyInfo")
-load("//py/private:py_semantics.bzl", _py_semantics = "semantics")
-load("//py/private/toolchain:types.bzl", "PY_TOOLCHAIN")
+load("//py/private/py_venv:types.bzl", "PY_VENV_KINDS", "VirtualenvInfo", "venv_root")
+load("//py/private/toolchain:types.bzl", "PY_TOOLCHAIN", "interpreter_files_and_version")
 
 def _runfiles_path(file, workspace):
     if file.short_path.startswith("../"):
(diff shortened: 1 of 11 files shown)


jbedard:
@tamird (the robot) please check if there are any concerns with the proposed 2.x breaking changes here

py/private/py_pex_binary.bzl
```diff
@@ -19,113 +19,208 @@ py_pex_binary(
...
     args.add(to_rlocation_path(ctx, entrypoint), format = "--entrypoint=%s")
     args.add(ctx.attr.python_shebang, format = "--python-shebang=%s")
 
     if ctx.attr.inherit_path != "":
         args.add(ctx.attr.inherit_path, format = "--inherit-path=%s")
 
-    py_version = py_toolchain.interpreter_version_info
+    # Stamp constraints from the binary's own interpreter version, the one the
+    # PEX is built for.
+    py_version = closure.version
```
tamird, line 242: closure.version only carries major/minor/micro, so the generated constraint loses the binary's ABI mode. A free-threaded 3.13 binary can contain cp313t extensions, but CPython==3.13.* matches both GIL and free-threaded interpreters; PEX can select ordinary 3.13 and then fail importing the packaged extension. PEX added CPython+t / CPython-t in 2.78 (https://github.com/pex-tool/pex/releases/tag/v2.78.0), while rules_py still pins 2.3.1, which also predates official Python 3.13 support (https://github.com/pex-tool/pex/releases/tag/v2.5.0). Please propagate ABI mode and use an ABI-specific default with a current PEX, or explicitly reject free-threaded inputs. The regression should execute a PEX containing a cp313t dependency, not only inspect PEX-INFO.

py/private/py_pex_binary.bzl
```diff
@@ -19,113 +19,208 @@ py_pex_binary(
...
-    # metadata itself via Distribution.load() on the `--dep` directory.
-    if site_packages_i != -1:
-        if f.path.find("dist-info", site_packages_i) != -1 and f.path.count("/", site_packages_i) == 2:
-            return []
-
-        return ["--dep={}".format(f.path[:site_packages_i + len("site-packages")])]
-
-    # If the path does not have a `site-packages` in it, then put it into the standard runfiles tree.
-    return ["--source={}={}".format(f.path, dest_path)]
+def _repo_root_prefix(short_path):
```
tamird, line 33: This only records external repository roots. A supported file-based hermetic runtime whose interpreter/files live in the main repo contributes no prefix and gets copied into the PEX, despite the rule's 'does not contain an interpreter' contract. Conversely, excluding an entire external repository can silently remove unrelated application data from a custom runtime repo. Please propagate and filter the actual interpreter files, handling directory artifacts by path prefix, and cover a main-repo py_runtime(interpreter = ...).

py/private/py_pex_binary.bzl
```diff
@@ -19,113 +19,208 @@ py_pex_binary(
...
+
+    wheels = []
+    roots = []
+    interp = []
+
+    # `version` is the binary's own interpreter version: it flows only from the
+    # `venv`/`actual` hop or a venv node's toolchain, never from deps/data (which
+    # carry other binaries' interpreters).
+    version = None
+    for attr_name in ["deps", "data"]:
```
tamird, line 95: The aspect walks arbitrary targets reachable through runtime data, so attrs named deps or data are not guaranteed to be label lists; they may be scalar labels, dictionaries, or unrelated values. Iterating them can fail analysis or apply provider membership to non-Targets. Please normalize supported label-bearing values and ignore other shapes. A custom runtime-data rule with scalar data = attr.label() would cover this.

py/private/py_pex_binary.bzl
```diff
@@ -150,7 +245,13 @@ def _py_python_pex_impl(ctx):
     ]
 
 _attrs = dict({
-    "binary": attr.label(executable = True, cfg = "target", mandatory = True, doc = "A py_binary target"),
+    "binary": attr.label(
+        executable = True,
+        cfg = "target",
+        mandatory = True,
+        doc = "The py_binary (or py_venv) target to package.",
```
tamird: A direct py_venv has only its executable in DefaultInfo.files, while lines 151-160 require exactly one additional entrypoint file. It therefore deterministically fails. Please remove 'or py_venv' unless direct venv packaging gains entrypoint semantics.

tamird (changes requested):
The version transition is sound, but I found four correctness gaps around free-threaded interpreters, interpreter-file filtering, arbitrary runtime-data rules, and the documented py_venv input. OpenAI has no current py_pex_binary consumers; these affect the upstream API contract.

tamird (changes requested):
Re-reviewed because Jason requested another pass in #ext-aspect-build. The current head still has the four correctness gaps from the prior review; the provider/aspect source files are byte-identical to the previously reviewed head, and the only changed PR file is a stronger PEX-content test.

Free-threaded binaries can still advertise an ABI-agnostic constraint with a pinned PEX version that predates the required constraint syntax; filtering still drops whole external interpreter repositories while retaining a supported main-repository interpreter; scalar/non-list data attributes can still fail aspect analysis; and the documented direct py_venv input still fails the single-entrypoint invariant. The original inline anchors remain current:
https://github.com/aspect-build/rules_py/pull/1344#discussion_r3616450819
https://github.com/aspect-build/rules_py/pull/1344#discussion_r3616450823
https://github.com/aspect-build/rules_py/pull/1344#discussion_r3616450824
https://github.com/aspect-build/rules_py/pull/1344#discussion_r3616450832

The version-transition regression and current CI are otherwise sound. These API-contract issues should be resolved or the supported surface narrowed before 2.x.

— tamirdex

tamird (commented):
Two additional issues from the cumulative packaging pass:

A wheel behind the common filegroup(srcs=[...]) data wrapper is no longer recognized: the closure aspect walks only deps/data/actual/venv, so the expanded install tree is sent as loose --source content instead of a PEX distribution. The previous runfile/site-packages classifier handled this case. Please cover a wheel-under-filegroup input and either traverse label-bearing srcs/propagate the provider or retain a safe fallback (py/private/py_pex_binary.bzl:114-123,187-211).

The new source mapper also compares every expanded runfile child with every wheel-tree prefix, making packaging O(files x wheels). Large PEXes can have both many wheels and many installed files; use structural filtering or an indexed lookup before this becomes the 2.x path.

— tamirdex

xangcastle (approved):

tamird (changes requested):
Re-reviewed at Jason’s explicit request. The updated head fixes the earlier scalar-attribute and filegroup-wheel handling and correctly derives the interpreter version from the packaged binary.

One blocker remains in the existing ABI discussion: https://github.com/aspect-build/rules_py/pull/1344#discussion_r3616450819. `interpreter_files_and_version` drops the interpreter ABI, so a free-threaded 3.13 binary containing `cp313t` extensions is advertised as `CPython==3.13.*`; PEX may then select ordinary `cp313` and fail at runtime. The tool is still pinned to PEX 2.3.1, which predates Python 3.13 and free-threaded interpreter support. The added regression only inspects `PEX-INFO` and does not execute a free-threaded native-wheel PEX.

Please propagate the ABI and upgrade PEX to a version supporting Python 3.13 free-threaded constraints, or explicitly reject unsupported free-threaded inputs, and add a real `cp313t` execution regression. Exact-head CI is still pending.

— tamirdex

Issue #723, opened 2025-11-21
[Bug]: `py_test` targets cannot start with `test_` if using `pytest_main = True`


hofbi:
### What happened?

When using `pytest_main = True` on `py_test`, my test target cannot start with `test_`. If it does, I get `E   ModuleNotFoundError: No module named 'test_foo'`. See https://github.com/aspect-build/rules_py/pull/722 as a reproducer.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 8.4.0 (version used in this repo)

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: main of this repo

Language(s) and/or frameworks involved:

### How to reproduce

```shell
# Checkout https://github.com/aspect-build/rules_py/pull/722
bazel test //examples/pytest:test_foo
```

### Any other information?

Why is this important?

The simple answer is of course to just choose a different name for the targets. However, `test_` is a very common pattern in Python for naming unit test files. Now if autogenerating the targets with gazelle, we would also expect to see a target with the same name which is not possible in this case.

Issue #483, opened 2024-12-16
[Bug]: py_test venv building doesn't work with slash in the name


keith:
### What happened?

If you have a target like this:

```bzl
py_test(
    name = "pytest_test/nested",
    srcs = [
        "foo_test.py",
        ":__test__",
    ],
    env_inherit = ["FOO"],
    imports = ["../.."],
    main = ":__test__.py",
    package_collisions = "warning",
    deps = [
        ":__test__",
        "@pypi_ftfy//:pkg",
        "@pypi_neptune//:pkg",
        "@pypi_pytest//:pkg",
    ],
)
```

Running this test works with rules_python and fails with rules_py with:

```
Error:   × Unable to run command:
  ├─▶ Unable to get full source dir path
  ╰─▶ No such file or directory (os error 2)
```

### Version

Development (host) and target OS/architectures: Ubuntu 22.04

Output of `bazel --version`: 7.4.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:

### How to reproduce

Run the test modified on this branch: https://github.com/aspect-build/rules_py/pull/484

keith:
adding some logging I bet the issue is that this path join:

```
"/home/ubuntu/.cache/bazel/_bazel_ubuntu/cbd66eedddd4d083f5e4d752aed24a45/sandbox/linux-sandbox/19/execroot/aspect_rules_py/bazel-out/k8-fastbuild/bin/examples/pytest/pytest_test/nested.runfiles/.pytest_test/nested.venv/lib/python3.9/site-packages" ::: "../../../../pypi_exceptiongroup/site-packages"
```

the second path is off by 1 `../` because of the target adding another level of nesting

PR #1345, opened 2026-07-19
feat(py)!: add py_pytest_test/py_unittest_test, remove pytest_main flag

### LINKED ISSUE #640: Restructure the ruleset surface area to allow for more purpose-oriented binaries
Discussed at some length with Jimmy -- Today we have `py_pex_binary` which is good and serves a narrow purpose and the new `py_venv_binary` which also serves a clear purpose. But users are broadly under the misapprehension that what is today called `py_binary` (see #587) is generally suitable as is what `rules_python` calls `py_binary`. The existing "binaries" such as Pex and the reference runfiles binary all make specific and consequently specifically inappropriate tradeoffs. Rather than presenting one as "the" binary implementation users should be forced to make a suitable to task decision.

<details>
<summary>Thinking out loud</summary>
The reality is that Python isn't designed for "binaries", it's more consistent with the world outside of Bazel think about venvs and entrypoints. It likely makes sense to reconsider the `PyInfo` structure as consisting of three parts; files, import prefixes and a list of entrypoints (think `console_scripts`). A `py_library` happens to specify no scripts. A `py_binary` happens to specify a script. Stitching multiple binaries with internally consistent deps together into a single venv should be a clearly supportable operation. Similarly choosing an _entrypoint_ out of a venv to be hit as part of a _closed_ artifact such as a sci-e, pex or other "binary" should be a distinct operation.

The need for such a thing becomes obvious when considering optimized Docker artifacts potentially containing several "binaries" (entrypoints). It isn't obvious to users that the "correct" way to do this under Bazel should be to create a real composite rule which forces the runfiles from several subsidiary binaries to become merged into a unified container-like artifact.
</details>

### LINKED ISSUE #868: [Roadmap]: Decouple test framework drivers from py_test
Tracking issue for decoupling test framework logic from the generic `py_test` macro.

Related: #640 (restructure ruleset surface area for purpose-oriented binaries)
Related: #723, #867 (pytest naming / discovery fixes)

## Phase 1: Targeted fix (PR #867) — no API break

- [x] Dunder-wrap generated `py_pytest_main` files (`__test__<name>__.py`)
- [x] Shared pytest main for `pytest_main=True` (no per-test codegen)
- [x] `pytest_paths` rule for cross-repo test discovery
- [ ] Deprecation warning on `py_pytest_main()` without custom params

## Phase 2: Extract `py_pytest_test` — additive, no break

- [ ] New `py_pytest_test` macro (always uses pytest, no `pytest_main` flag)
- [ ] Owns `pytest_paths`, shared main injection, pytest dep validation
- [ ] Auto-detection lives here (pytest is always the driver, no ambiguity)
- [ ] `py_test(pytest_main=True)` continues working, warns to migrate

## Phase 3: `py_unittest_test` — additive, no break

- [ ] New `py_unittest_test` macro using stdlib `unittest`
- [ ] Bazel integration: coverage, sharding, JUnit XML, test filtering
- [ ] `unittest.TestLoader().discover()` for collection

## Phase 4: Clean break — deliberate API break (2.x)

- [ ] Remove `pytest_main` parameter from `py_test`
- [ ] `py_test` = generic (run a Python file as a test, nothing more)
- [ ] `py_pytest_test` = pytest driver
- [ ] `py_unittest_test` = unittest driver
- [ ] Migration guide + deprecation cycle

### API break analysis

- **Phases 1–3**: purely additive, no breaks
- **Phase 4**: breaking — `py_test(pytest_main=True)` stops working, must migrate to `py_pytest_test`

jbedard:
Introduce purpose-oriented test macros so each framework has its own entry
point instead of overloading py_test with a pytest_main flag:

- py_pytest_test: always drives pytest; owns pytest_paths discovery, shared
  main injection, and per-test codegen only for baked pytest_args/chdir.
- py_unittest_test: stdlib unittest driver that imports each declared srcs
  file directly (never discover(), so nested roots don't double-run and
  same-basename files don't collide), with a built-in JUnit XML writer (no
  third-party runner) plus coverage, sharding, and --test_filter integration.
  Runtime args are parsed (-v/-q, --failfast, --buffer, and native unittest
  -k: repeatable, ORed, fnmatch); unknown args are rejected and a filter
  matching no tests fails the run.

py_test is now strictly generic (runs a Python file as a test). All in-repo
callers migrate to py_pytest_test. py_pytest_main is kept as the low-level
escape hatch for hand-written/wrapped entrypoints (#1094); docs now steer new
suites to py_pytest_test.

Both macros reject a user-supplied main. The unittest JUnit writer records
failing subTests, and a baked chdir no longer defeats pytest discovery roots
(the .pytest_paths file is read before chdir).

Close #868
Close #640

BREAKING CHANGE: `py_test(pytest_main = True)` is removed. Use `py_pytest_test`
instead; see docs/test-drivers.md. Both `py_pytest_test` and `py_unittest_test`
treat every `srcs` file as a test module (support code belongs in `deps`,
conftest.py in `data`); select tests by name with Bazel glob() in srcs.

### Test plan

- New test cases added


jbedard:
@tamird (the robot) please check if there are any concerns with the proposed 2.x breaking changes here

py/private/unittest_main.py
```diff
@@ -0,0 +1,319 @@
...
+def _discover(start_dirs, pattern):
+    loader = unittest.TestLoader()
+    suite = unittest.TestSuite()
+    for d in start_dirs:
+        if os.path.isdir(d):
+            # top_level_dir defaults to the start dir, which puts `d` on
+            # sys.path and imports test modules by basename. This avoids
+            # requiring `d` to be an importable package relative to the
+            # workspace root (it often isn't — e.g. a hyphenated dir name).
+            suite.addTests(loader.discover(start_dir=d, pattern=pattern))
```
tamird: This recursively discovers once for every source directory, so nested roots run tests twice and sibling directories with the same filename fail collection. For example, declared `suite/outer_test.py` plus `suite/child/inner_test.py` executes the inner test twice; `a/same_test.py` plus `b/same_test.py` raises `ImportError: 'same_test' module incorrectly imported from '.../a'. Expected '.../b'` because discovery imports both by basename. The current e2e case has only one directory. Can the driver load exactly the declared files once (or otherwise collapse nested roots and preserve unique module identity), with nested/same-basename regressions?

py/private/unittest_main.py
```diff
@@ -0,0 +1,319 @@
...
+
+def _filter_suite(suite, substrs):
+    """Narrow a suite to tests whose dotted id contains every substring in
+    `substrs`. unittest has no native -k, so both Bazel's --test_filter
+    (TESTBRIDGE_TEST_ONLY) and any `-k` args are applied here as substring
+    matches (ANDed together)."""
+    matched = unittest.TestSuite()
+    for test in _iter_tests(suite):
+        tid = test.id()
+        if all(s in tid for s in substrs):
```
tamird: `unittest` already implements `-k` through `TestLoader.testNamePatterns`: repeated patterns are ORed and `*` uses fnmatch. This reimplementation ANDs repeated filters and treats `*` literally, so `-k test_pass -k test_membership` and `-k '*membership*'` select 2/1 tests with stdlib unittest but 0 here. Please preserve native `-k` semantics (and test repeated/wildcard patterns); Bazel's filter can then be combined deliberately instead of changing the framework contract.

docs/test-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+
+Include the `pytest` package (and `coverage`, for coverage) in `deps`. Baked
+pytest args or a working directory go on the `pytest_args` / `chdir`
+attributes; runtime args (forwarded to `sys.argv`) go on the standard `args`
+attribute.
+
+With Gazelle, point the `py_test` mapping at `py_pytest_test`:
+
+```
+# gazelle:map_kind py_test py_pytest_test @aspect_rules_py//py:defs.bzl
```
tamird: Mapping Gazelle directly to `py_pytest_test` is not sufficient for ordinary assert-only tests: Gazelle cannot infer `pytest` when the test never imports it, while the new macro still requires pytest in `deps`. The prior README guidance explicitly used a wrapper to inject that dependency/config, and the live Aspect starter/rules_lint examples rely on exactly that pattern. Please keep the wrapper guidance, updated to call `py_pytest_test`, instead of telling users to map directly. The README example also still pins rules_py 1.11.2, where this symbol does not exist.

tamird (changes requested):
I found two unittest execution regressions and a Gazelle migration regression. The proposed API break also has a concrete downstream cost: OpenAI currently has 84 direct `pytest_main = True` call sites across 15 BUILD files, so we will need the migration before the next snapshot bump. The inline comments include small reproductions and the missing coverage.

examples/pytest/subdir/BUILD
```diff
@@ -12,9 +12,8 @@ py_test(
     deps = [":foo"],
 )
 
-py_test(
+py_pytest_test(
     name = "test_with_pytest_main",
     srcs = ["tests/test_subdir.py"],
-    pytest_main = True,
     deps = [":foo"],
```
tamird: Can this obsolete `BUILD` be deleted? Bazel uses the sibling `BUILD.bazel`, so this duplicate is never exercised and its migrated pytest target is actually missing the pytest dependency that the live file has. Keeping two divergent examples makes the migration misleading.

tamird (commented):
One small cleanup from the downstream/migration pass.

py/private/unittest_main.py
```diff
@@ -0,0 +1,319 @@
...
+        '  <testsuite name=%s tests="%d" failures="%d" errors="%d" skipped="%d" time="%.3f">'
+        % (quoteattr(suite_name), len(records), failures, errors, skipped, total_time)
+    )
+    for r in records:
+        lines.append(
+            '    <testcase classname=%s name=%s time="%.3f">'
+            % (quoteattr(r["classname"]), quoteattr(r["name"]), r["time"])
+        )
+        if r["status"] in ("failure", "error"):
+            lines.append(
```
tamird, line 256: `quoteattr`/`escape` do not remove XML 1.0-forbidden control characters. A perfectly ordinary failure such as `self.fail("contains ESC: " + chr(27) + "[31m")` (or a message containing NUL) writes those bytes into the JUnit file and `ElementTree.fromstring` fails with `not well-formed (invalid token)`. ANSI-colored exception messages make this realistic. Please sanitize message/detail like pytest's JUnit `bin_xml_escape` and have the regression parse the emitted XML; the current shell tests only grep it.

tamird (commented):
One additional JUnit correctness issue from the adversarial pass.

py/private/unittest_main.py
```diff
@@ -0,0 +1,319 @@
...
+    suite = _shard(discovered)
+    if filters:
+        suite = _filter_suite(suite, filters)
+
+    # Record outcomes when Bazel asks for JUnit XML; the built-in writer below
+    # needs no third-party runner.
+    xml_out = os.environ.get("XML_OUTPUT_FILE")
+    runner = unittest.TextTestRunner(
+        verbosity=1 if opts.quiet else 2,
+        failfast=opts.failfast,
```
tamird: This ignores Bazel's `--test_runner_fail_fast`, which is forwarded as `TESTBRIDGE_TEST_RUNNER_FAIL_FAST=1`; only the driver-local `--failfast` changes behavior. As a result a target with two failing methods still runs both under `bazel test --test_runner_fail_fast`. Please OR the Bazel env flag into `failfast` and add a two-failure regression; the current check only passes `--failfast` to an all-passing suite.

tamird (commented):
One more Bazel integration gap from the execution pass.

py/private/py_pytest_test.bzl
```diff
@@ -0,0 +1,134 @@
...
+        py_pytest_main(
+            name = main_target,
+            args = pytest_args,
+            chdir = chdir,
+            tags = tags,
+        )
+        deps.append(":" + main_target)
+
+        # py_pytest_main renders its output to `__test__<name>__.py`.
+        main = ":__test__%s__.py" % main_target
```
tamird: This codegen path reintroduces the #723 collection bug for supported target names containing `/` (#483). With `name = "sharded/test"` and baked args/chdir, `py_pytest_main` dunder-wraps the whole path and emits `__test__sharded/test_pytest_main__.py`; the basename still starts `test_`, so pytest recurses into that directory and collects the generated main. The existing slash-name regression only exercises the shared-main path. Please dunder-wrap the final basename while preserving the directory prefix, and add a slash-name `pytest_args`/`chdir` regression.

tamird (commented):
The baked pytest-entrypoint path also regresses the existing slash-name convention.

py/private/unittest_main.py
```diff
@@ -0,0 +1,319 @@
...
+    opts = _parse_args(sys.argv[1:])
+
+    # The next two assignments are rewritten at analysis time by py_unittest_test
+    # with the target's discovery roots and collection pattern. Keep them on
+    # their own lines exactly as written — the rule keys on the bare assignment
+    # text, so editing these comments is safe but editing the code is not.
+    start_dirs = ["."]
+    pattern = "*_test.py"
+
+    discovered = _discover(start_dirs, pattern)
```
tamird: For a sharded target, the early returns below (no discovery or no filter match) happen before `_shard()` touches `TEST_SHARD_STATUS_FILE`. Bazel then replaces the useful error with `Sharding requested, but the test runner did not advertise support...`. Please advertise sharding before collection/filter validation (as the existing Bazel unittest adapters do) and cover an empty/no-match sharded run.

tamird (commented):
One sharding failure-path issue from the same execution pass.

tamird (commented):
Two related edge cases from the execution pass:

- The newly public `py_pytest_test(chdir = ...)` path exposes the existing renderer at `py/private/py_pytest_main.bzl:55`, which emits `os.chdir('{}')`. A valid runfiles directory such as `pkg/it's-data` generates an unterminated Python string (and backslashes can be reinterpreted). Render `repr(ctx.attr.chdir)` and cover a quoted path.
- The new JUnit writer declares UTF-8 but opens `XML_OUTPUT_FILE` with the platform-default encoding. On a cp1252 Windows locale, `self.fail("unicode failure: 🔥")` raises `UnicodeEncodeError` while writing the report. Open it with `encoding="utf-8"`; this can share the XML-sanitization regression.

py/private/unittest_main.py
```diff
@@ -0,0 +1,388 @@
...
+    # with the target's own-repo source files and collection pattern. Keep them
+    # on their own lines exactly as written — the rule keys on the bare
+    # assignment text, so editing these comments is safe but editing the code is
+    # not.
+    test_files: List[str] = []
+    pattern = "*_test.py"
+
+    modules = _import_test_modules(test_files, pattern)
+
+    if _suite_from(unittest.TestLoader(), modules).countTestCases() == 0:
```
tamird: This emptiness check calls loadTestsFromModule once here and then again below when selected is constructed. That invokes each module's standard load_tests(loader, tests, pattern) hook twice. A stateful/dynamic hook that returns its suite only on the first invocation produces first=1, second=0; with no filter the second empty suite reaches TextTestRunner, reports "Ran 0 tests", and exits successfully, defeating the new no-tests guard.

Please collect once and filter the resulting suite (or otherwise preserve a single load_tests invocation), with a fixture that counts calls and returns tests only on the first call.

[tamirdex]

py/private/pytest_main.py
```diff
@@ -69,6 +69,22 @@
...
+    test_dirs: List[str] = []
+    target_name = os.environ.get("BAZEL_TARGET_NAME", "")
+    target = os.environ.get("BAZEL_TARGET", "")
+    if target:
+        package = target.split(":")[0].lstrip("/")
+        paths_file = os.path.join(package, target_name + ".pytest_paths")
+        if os.path.isfile(paths_file):
+            with open(paths_file) as f:
+                for line in f:
+                    d = line.strip()
```
tamird: A workspace-root test source is encoded by pytest_paths as an empty directory entry, which this condition discards. The new public py_pytest_test then passes no positional roots and pytest recursively collects the entire runfiles tree; an unrelated failing test supplied as data can therefore fail this target, recreating the isolation problem this path is meant to solve.

Passing "." would still collect unrelated root files. Please pass the declared test files/a guaranteed private root, or reject and document workspace-root pytest targets, and add a root-package regression with an unrelated failing test in data.

[tamirdex]

py/private/unittest_main.py
```diff
@@ -0,0 +1,388 @@
...
+
+    suite = _shard(selected)
+
+    failfast = opts.failfast or os.environ.get("TESTBRIDGE_TEST_RUNNER_FAIL_FAST") == "1"
+
+    # Record outcomes when Bazel asks for JUnit XML; the built-in writer below
+    # needs no third-party runner.
+    xml_out = os.environ.get("XML_OUTPUT_FILE")
+    runner = unittest.TextTestRunner(
+        verbosity=1 if opts.quiet else 2,
```
tamird: opts.verbose is never used here: default and -v both select verbosity=2, while -q selects the normal verbosity=1. Stdlib unittest uses quiet=0, default=1, verbose=2, so the documented -v/-q interface does not behave as advertised.

Use 0 if quiet, 2 if verbose, otherwise 1, and assert the three observable output modes in the args regression.

[tamirdex]

docs/test-drivers.md
```diff
@@ -0,0 +1,88 @@
...
+```starlark
+load("@aspect_rules_py//py:defs.bzl", "py_unittest_test")
+
+py_unittest_test(
+    name = "test_foo",
+    srcs = ["foo_test.py"],
+)
+```
+
+Collects with `unittest.TestLoader().discover()` rooted at the directories
```
tamird: This and the workspace-root warning below still describe TestLoader().discover() recursing source directories, but the corrected implementation directly loads the declared files and never calls discover(). Workspace-root unittest targets consequently do not have the documented recursion hazard; that hazard now belongs to the pytest directory-discovery path instead.

Please update the collection/filter/root guidance to match the actual driver (including native -k wildcard/OR semantics).

[tamirdex]

e2e/cases/pytest-main-867/chdir_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+def test_chdir_applied() -> None:
+    # The baked chdir moved CWD into chdir_data before pytest started.
+    assert os.path.basename(os.getcwd()) == "chdir_data"
+
+
+def test_discovery_survived_chdir() -> None:
+    # Being collected at all proves discovery found this source in the package
+    # root despite the chdir. Under the bug pytest would exit 5 (no tests).
+    assert True
```
tamird: This added test is literally assert True and provides no signal beyond the adjacent test already being collected and asserting the changed working directory. Please delete it; the target's successful collection plus test_chdir_applied is the meaningful regression.

[tamirdex]

tamird (changes requested):
Re-reviewed at Jason's request after the previous feedback was addressed. The earlier discovery, filtering, JUnit, fail-fast, slash-name, chdir, and sharding findings are fixed and CI is green.

Three execution issues remain: unittest collection invokes every load_tests hook twice and can silently run zero tests, workspace-root pytest targets fall back to collecting the whole runfiles tree, and the advertised quiet/default verbosity does not match unittest. The new documentation also describes the removed discover() path, and the added assert-True test is tautological. The inline comments include small falsifying cases.

The unittest driver is now roughly 400 handwritten lines and uses Dict[str, Any] for its closed JUnit record/status model; please tighten that surface while correcting the execution path so the new type-checking gate can protect the contract.

[tamirdex]

py/private/unittest_main.py
```diff
@@ -0,0 +1,407 @@
...
+
+    lines = ['<?xml version="1.0" encoding="UTF-8"?>', "<testsuites>"]
+    lines.append(
+        '  <testsuite name=%s tests="%d" failures="%d" errors="%d" skipped="%d" time="%.3f">'
+        % (quoteattr(suite_name), len(records), failures, errors, skipped, total_time)
+    )
+    for r in records:
+        lines.append(
+            '    <testcase classname=%s name=%s time="%.3f">'
+            % (quoteattr(r.classname), quoteattr(r.name), r.time)
```
tamird: Failure/detail are sanitized, but `r.name` and `r.classname` are passed directly to `quoteattr`. A failing `self.subTest(value=AnsiRepr())` whose `__repr__` returns `"\x1b[31m"` sends that escape through `str(subtest)` into the testcase name, and `ElementTree.fromstring` raises `ParseError: not well-formed (invalid token)`. Please sanitize every interpolated XML string (including suite/class/name), and extend the control-character regression to parse a failing subtest with an ANSI parameter repr.

— tamirdex

py/private/py_pytest_main.bzl
```diff
@@ -18,41 +18,61 @@
...
+    recurse the whole runfiles tree, collecting unrelated files supplied as
+    data). Support code belongs in `deps`, and conftest.py in `data`."""
+    files = {}
     for src in ctx.files.srcs:
         p = src.short_path
 
         # Skip external-repo sources (../reponame/...) — only the test's
-        # own workspace sources should be discovery roots.
+        # own workspace sources are collected.
         if p.startswith("../"):
```
tamird: `ctx.files.srcs` may contain valid external-repository labels whose short paths begin `../`. Silently dropping them breaks both new public drivers: a pytest target containing only such a source passes no positional collection paths and recurses the entire runfiles tree, while the unittest target reports no tests found. `_unittest_main_impl` repeats the same filter. Resolve these files to their correct runfiles paths or reject unsupported external `srcs` during analysis; cover an external-repository source for both drivers.

— tamirdex

tamird (changes requested):
Re-reviewed at Jason's explicit request on current head 8cf17159764a. The previous collection, load_tests, verbosity, documentation, and tautological-test findings are fixed. Two correctness gaps remain: external-repository test sources are silently discarded by both new drivers, and the JUnit sanitizer still emits malformed XML for parameterized subtest names. The inline comments include concrete failure cases and narrow corrections.

— tamirdex

tamird (approved):
Re-reviewed at Jason's request. External-repository pytest and unittest sources now run with explicit regression coverage, and JUnit sanitizes suite, classname, test name, failure message, and detail. Collection, sharding, fail-fast, filtering, verbosity, coverage, and current-head CI are clean.

— tamirdex

xangcastle (approved):

PR #1346, opened 2026-07-19
chore: type-check Python tooling

tamird:
Annotate the maintained Python tools, tests, and examples, enforce complete function signatures, and run locked Pyright and ty checks at the supported Python floors in CI. Keep dev-only tooling out of the Bazel dependency closure by pruning unreachable SCCs, so generated targets remain valid and Gazelle does not index unused wheels.


tamird:
@codex review

tamird:
@codex review

tamird:
@codex review

tamird:
@codex review

tamird:
@jbedard ready for review. The maintained Python tools, tests, and examples now have enforced signatures plus locked Pyright/ty checks at the supported Python floors. Adding dev tooling exposed an invalid unreachable-SCC path, so the final design prunes those SCCs before repository generation; all generated targets analyze, dev wheels stay out of Gazelle, Bazel 8/9 and e2e checks are green, startup is unchanged, and fresh Codex is clean.

tamird:
@codex review

py/tools/site_merge/site_merge.py
```diff
@@ -31,18 +31,24 @@
...
 
-def _remove(path):
+def _remove(path: Path) -> None:
     """Remove an output copied from a potentially read-only input."""
 
-    def retry_readonly(function, candidate, exc_info):
+    def retry_readonly(
+        function: Callable[..., object],
+        candidate: str,
+        exc_info: Tuple[type[BaseException], BaseException, TracebackType],
```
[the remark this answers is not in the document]
tamird: Fixed with postponed annotations so site_merge remains importable on Python 3.8. I also checked the other introduced annotations for runtime-evaluated 3.9+ generics in 3.8-supported tools; none remain.

tamird:
@codex review

examples/py_venv/say.py
```diff
@@ -2,13 +2,14 @@
 
 print("---")
 import _virtualenv
+from typing import Union
```
michaelm-openai: what versions of python do these run under? Can `|` syntax be used?

tamird: This example provisions Python 3.11 in `examples/py_venv/MODULE.bazel`, so `|` is supported. Switched the annotation and dropped the `Union` import; the three `py_venv` example tests pass.

michaelm-openai: Could we make the switch elsewhere too, thx

tamird:
@codex review

examples/py_venv/say.py
tamird: i think not - only the example runs with python 3.11; impl code has to support older versions.

uv/private/pep517_whl/build_helper.py
```diff
@@ -15,11 +18,12 @@
 from os import chmod, defpath, listdir, makedirs, path, pathsep
 from subprocess import CalledProcessError, check_call, check_output, STDOUT, run
 from tempfile import TemporaryFile
+from typing import Dict, Optional
 
 try:
-    import tomllib
+    _tomllib = importlib.import_module("tomllib")
```
jbedard: Why the var rename?

py/private/py_venv/templates/_virtualenv.py
```diff
@@ -1,8 +1,23 @@
 """Patches that are applied at runtime to the virtual environment."""
 
+from __future__ import annotations
+
 import os
 import sys
 
+# Defining this locally avoids importing typing at startup.
+TYPE_CHECKING = False
+if TYPE_CHECKING:
```
jbedard, line 10: This avoids the imports at runtime while keeping them for type-checking? Is there not a more standard way of doing this? 🤔

jbedard: Normally this is imported from `typing`, but you want to completely avoid that import all together?

uv/private/pep517_whl/build_helper.py
tamird: No reason to rename it. Restored `tomllib` while keeping the dynamic import used for the Python 3.9 `tomli` fallback; the focused checkers and PEP 517 build test pass.

py/private/py_venv/templates/_virtualenv.py
tamird: Yes. This hook runs on every Python startup, so importing `TYPE_CHECKING` from `typing` would defeat the startup constraint. `TYPE_CHECKING = False` is the established pre-3.15 workaround described in [PEP 781](https://peps.python.org/pep-0781/); postponed annotations keep the names runtime-free. The pinned Python 3.9 Pyright/ty checks pass, and a `python -S` import confirms neither `typing` nor `threading` is loaded.

jbedard: So when we only support only 3.15+ this should switch to importing from `typing`? Is that worth a 1 line comment?

tamird: Added the one-line rationale and refreshed the snapshot. The proposed built-in would replace this local definition (not require importing from `typing`); PEP 781 is still Draft, so the comment intentionally does not promise a Python version. The snapshot/link tests and Python 3.9 checks pass.

tamird:
@codex review

uv/private/sdist_configure/detect_native.py
```diff
@@ -24,9 +27,10 @@
 import tempfile
 import types
 import zipfile
+from typing import Callable, Dict, Iterator, List, NoReturn, Optional, Sequence, Tuple, TypedDict
 
 try:
-    import tomllib
+    tomllib: Optional[types.ModuleType] = importlib.import_module("tomllib")
 except ModuleNotFoundError:
     tomllib = None
```
tamird, line 35: why is this None? in `uv/private/pep517_whl/build_helper.py` a different fallback is used.

tamird: They run in different environments. `detect_native.py` is the dependency-free fetch-time configure tool, so pre-3.11 has no `tomli` available and intentionally warns/skips `[build-system]` while preserving native/setup metadata detection; `test_pyproject_degrades_without_tomllib` covers that path. `build_helper.py` runs in the build environment, where the conditional `tomli` dependency is present. These should remain separate.

tamird:
@codex review

jbedard (approved):

.github/workflows/ci-workflows.yaml
```diff
@@ -33,6 +33,37 @@ jobs:
...
               run: aspect gazelle --task-key=gazelle
+    typecheck:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v6
+            - uses: aspect-build/setup-aspect@c22a8f64fb38f82f59ce809cd7eb9f8ae096da44 # v2026.23.2
+              with:
+                aspect-api-token: ${{ secrets.ASPECT_API_TOKEN }}
+            - name: Check Node runtime
+              run: node --version
```
jbedard: Why do we need node?

.github/workflows/ci-workflows.yaml
```diff
@@ -33,6 +33,37 @@ jobs:
...
+            - name: Ruff annotations
+              run: bazel run @uv -- run --locked --group dev -- ruff check .
+            - name: Typecheck maintained Python
+              run: |
+                runtime39=(
```
jbedard, line 47: So we have the 3.9+, the 3.11 (detect_native.py). Where are all the tests checked?

Can we put the detect_native.py entry in a variable like this one to better declare the reason its separate?

.github/workflows/ci-workflows.yaml
tamird: Removed the unnecessary Node probe.

.github/workflows/ci-workflows.yaml
tamird: Named the Python-3.11 set explicitly as runtime311. Ruff checks annotations across all non-excluded Python files, including tests; the existing test matrix executes them. Pyright/ty remain scoped to the maintained runtime/tool files.

tamird:
@codex review

tamird:
@codex review

uv/private/extension/defs.bzl
```diff
@@ -388,6 +388,25 @@ def _parse_projects(module_ctx, hub_specs):
                     for version, markers in versions.items():
                         marked_package_cfg_sccs.setdefault(package, {}).setdefault(cfg, {}).setdefault(package_cfg_sccs[version][cfg], {}).update(markers)
 
+            # The lock may contain inactive dev-only packages. Keep the SCCs
```
jbedard, line 391: Is this a bugfix that should include a test?

tamird: Yes. Adding the type-checking dev dependencies exposed unreachable SCCs: they were emitted into the project SCC graph and Gazelle index despite having no surface alias. `//uv/private/uv_hub:snapshots_tests` is the regression—it pins both generated project BUILD files and would restore the five dev-only indexes/SCCs if this filtering were removed. I also clarified that coverage next to the two snapshot entries; the uncached suite passes.

tamird:
@codex review

uv/private/extension/defs.bzl
jbedard: The snapshots show extra entries now present. How can we prove the bug via running code?

tamird: Added an executed regression that builds every generated project SCC. With the filter removed, analysis emits 92 SCCs and fails because `pyright__1_1_411` references the missing `//:nodeenv` alias; with the filter restored, 87 SCCs build successfully. This proves the invalid generated graph, rather than relying on snapshots.

tamird:
@codex review

uv/private/extension/defs.bzl
jbedard: Write a test from a rules_py user perspective

jbedard: Can that test be a `py_test` instead of `test.sh` or `sh_test`?

tamird: Folded the public Gazelle assertion into the existing py_test and deleted both shell tests. With the filter removed, the test fails because the generated manifest contains inactive wheel and packaging mappings; with the filter restored, it contains only cowsay and passes on Bazel 8 and 9.

tamird:
@codex review

pyproject.toml
```diff
@@ -20,3 +20,77 @@ dependencies = [
...
+   "pyright==1.1.411",
+   "ruff==0.15.22",
+   "ty==0.0.61",
+]
+
+[tool.ruff]
+target-version = "py39"
+extend-exclude = [
+   "**/third_party/**",
+   "examples/virtual_deps/cowsnake/cowsay.py",
```
jbedard: Why are so many things excluded? Is this only for old python versions? Why not exclude all `examples/` and `e2e/` on old versions?

tamird: These are synthetic/vendor fixtures that intentionally contain unannotated functions, not old-Python syntax. Ruff reports the same failures at 3.9 and 3.11, so excluding all examples/e2e would lose current coverage. I removed the redundant entries and kept only fixtures that actually violate the annotation rules; the remaining 280 files stay checked.

tamird:
@codex review

pyproject.toml
```diff
@@ -20,3 +20,56 @@ dependencies = [
...
+   "py/tests/external-deps/custom-macro",
+   "py/tests/external-deps/lib.py",
+   "py/tests/internal-deps/adder",
+   "py/tests/internal-deps/pi.py",
+   "py/tests/internal-deps/sub.py",
+   "py/tests/py-venv-standalone-interpreter/ex.py",
+   "py/tests/py_venv_conflict/lib.py",
+   "py/tests/repo_relative_imports/lib",
+   "py/tests/virtual/django/proj",
+   "py/tests/wheel-root-pth-double/a",
```
jbedard: Why must some of these remain and others were removed?

tamird: Annotated all ordinary first-party fixtures and removed their exclusions. Only vendored `third_party` and the two OCI fixture files whose exact byte sizes are asserted by four layer snapshots remain; changing those requires their canonical snapshot writers. Ruff passes at 3.9 and 3.11.

tamird:
@codex review

e2e/cases/single-project-hub/BUILD.bazel
```diff
@@ -3,14 +3,23 @@
...
+    data = [":gazelle_python_manifest"],
     dep_group = "single_project_hub",
     main = "__test__.py",
     python_version = "3.11",
     deps = [
         "@pypi_single//cowsay",
     ],
 )
+
+gazelle_python_manifest(
```
jbedard, line 21: Why add gazelle usage when adding types?
Why add gazelle usage in this test?

tamird: Adding the type-checking dev dependencies exposed the bug this covers: dev-only wheel/packaging/pyright SCCs were emitted into the project graph and Gazelle index despite having no surface alias (the full graph failed on the missing `//:nodeenv` alias). This existing single-project fixture gives an executed public regression: its lock deliberately includes an inactive dev wheel, and the generated Gazelle manifest must contain only the active cowsay mapping. Without the SCC filter the assertion fails; with it, Bazel 8 and 9 pass.

PR #1347, opened 2026-07-19
feat(uv): detect source-built console scripts

tamird:
Read console scripts from a source distribution's top-level egg-info
metadata and forward them to venv assembly, removing the need for
handwritten overrides. Explicit overrides remain authoritative and inactive
platform wheels cannot leak scripts.

### DIFF
--- a/uv/private/pep517_whl/rule.bzl
+++ b/uv/private/pep517_whl/rule.bzl
@@ -9,6 +9,7 @@ load("@bazel_lib//lib:resource_sets.bzl", "resource_set", "resource_set_attr")
 load("@rules_cc//cc:action_names.bzl", "ACTION_NAMES")
 load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
 load("//py/private/toolchain:types.bzl", "NATIVE_BUILD_TOOLCHAIN", "PY_TOOLCHAIN")
+load("//uv/private:source_built_wheel.bzl", "SourceBuiltWheelInfo")
 
 _CC_TOOLCHAIN_TYPE = Label("@bazel_tools//tools/cpp:toolchain_type")
 _TARGET_EXEC_GROUP = "target"
@@ -173,7 +174,10 @@ def _pep517_whl(ctx):
         resource_set = resource_set(ctx.attr),
     )
 
-    return [DefaultInfo(files = depset([wheel_dir]))]
+    return [
+        DefaultInfo(files = depset([wheel_dir])),
+        SourceBuiltWheelInfo(console_scripts = tuple(ctx.attr.console_scripts)),
+    ]
 
 def _pep517_native_whl(ctx):
     archive = ctx.file.src
@@ -220,7 +224,10 @@ def _pep517_native_whl(ctx):
         resource_set = resource_set(ctx.attr),
     )
 
-    return [DefaultInfo(files = depset([wheel_dir]))]
+    return [
+        DefaultInfo(files = depset([wheel_dir])),
+        SourceBuiltWheelInfo(console_scripts = tuple(ctx.attr.console_scripts)),
+    ]
 
 _PATCH_ATTRS = {
     "pre_build_patches": attr.label_list(
@@ -241,6 +248,9 @@ _pep517_whl_attrs = {
     # https://bazel.build/extending/exec-groups#defining-exec-groups
     "tool": attr.label(executable = True, cfg = config.exec(_TARGET_EXEC_GROUP)),
     "version": attr.string(),
+    "console_scripts": attr.string_list(
+        doc = "Console scripts discovered from the source distribution's entry-point metadata.",
+    ),
     "args": attr.string_list(default = ["--validate-anyarch"]),
     "monitor_memory": attr.bool(
         default = False,
--- a/uv/private/pep517_whl/rule.bzl
+++ b/uv/private/pep517_whl/rule.bzl
@@ -21,6 +21,12 @@ _INHERITED_PYTHON_ENV = (
     "PYTHONPLATLIBDIR",
 )
 
+def _wheel_providers(wheel_dir, console_scripts):
+    return [
+        DefaultInfo(files = depset([wheel_dir])),
+        SourceBuiltWheelInfo(console_scripts = tuple(console_scripts)),
+    ]
+
 def _common_env(ctx):
     # pyproject_hooks copies the build process environment and launches its
     # Python executable without -I:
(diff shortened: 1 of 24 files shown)


tamird:
@codex review

uv/private/sdist_configure/detect_native.py
```diff
@@ -418,6 +453,22 @@ def detect(archive_path, context):
             if content:
                 declared.extend(_parse_setup_cfg_build_requires(content))
 
+        # Only a top-level egg-info directory belongs to this sdist. Nested
+        # entry-point metadata may be vendored and must not create wrappers.
+        member_paths = [PurePosixPath(name) for name in members]
+        rooted = len({path.parts[0] for path in member_paths}) == 1
+        entry_points_paths = [
+            str(path) for path in member_paths
```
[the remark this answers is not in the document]
tamird: Fixed. Structural checks now use the parsed path while archive reads retain the original member name; tar and zip coverage includes `./pkg-1.0/.../entry_points.txt`.

uv/private/whl_install/rule.bzl
```diff
@@ -4,23 +4,20 @@
...
-        "console_scripts": "Complete tuple[str] encoded as name=module:object.",
-    },
-)
+load("//uv/private:source_built_wheel.bzl", "SourceBuiltWheelInfo")
 
 def _source_built_wheel_impl(ctx):
     source = ctx.attr.src[DefaultInfo]
+    console_scripts = ctx.attr.console_scripts
+    if not console_scripts and SourceBuiltWheelInfo in ctx.attr.src:
+        console_scripts = ctx.attr.src[SourceBuiltWheelInfo].console_scripts
```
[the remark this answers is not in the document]
tamird: Fixed. Override presence now remains distinct from an empty mapping, so an explicit `{}` suppresses detected scripts while unrelated package overrides still inherit them. Added source/prebuild, custom-build-file, and wheel-only coverage.

tamird:
@codex review

uv/private/sdist_configure/detect_native.py
```diff
@@ -418,6 +453,22 @@ def detect(archive_path, context):
             if content:
                 declared.extend(_parse_setup_cfg_build_requires(content))
 
+        # Only a top-level egg-info directory belongs to this sdist. Nested
+        # entry-point metadata may be vendored and must not create wrappers.
+        member_paths = [(name, PurePosixPath(name)) for name in members]
+        rooted = len({path.parts[0] for _, path in member_paths}) == 1
+        entry_points_paths = [
+            name for name, path in member_paths
+            if len(path.parts) == (3 if rooted else 2)
```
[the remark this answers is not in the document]
tamird: Fixed. The detector now matches a unique egg-info directory against the root PKG-INFO distribution name at any depth, preserves raw archive-member names, and keeps the shallow Git-archive fallback. Added src/lib/flat, prefixed tar/zip, vendored-metadata, and duplicate-own regressions.

tamird:
@codex review

uv/private/whl_install/rule.bzl
```diff
@@ -4,23 +4,20 @@
...
-    fields = {
-        "console_scripts": "Complete tuple[str] encoded as name=module:object.",
-    },
-)
+load("//uv/private:source_built_wheel.bzl", "SourceBuiltWheelInfo")
 
 def _source_built_wheel_impl(ctx):
     source = ctx.attr.src[DefaultInfo]
+    console_scripts = ctx.attr.console_scripts
+    if not ctx.attr.console_scripts_override and not console_scripts and SourceBuiltWheelInfo in ctx.attr.src:
```
jbedard: `and not console_scripts` is dead here. When `console_scripts_override` is `False`, the pipeline always emits `console_scripts = []` (`defs.bzl` sets `sbuild_console_scripts or []` and only flips the override bool when the value is non-`None`), so the conjunct is always true inside this branch. Since `console_scripts_override` is internal-only — populated solely by the extension, never user-authored — there's no direct-rule caller to guard, so I'd just drop it:

```starlark
if not ctx.attr.console_scripts_override and SourceBuiltWheelInfo in ctx.attr.src:
```

As written the conjunct reads like a third state that can't occur.

(To be clear, the `console_scripts_override` flag *itself* is necessary: it's the only thing distinguishing "unset" from an explicit empty map, which both serialize to `[]`. Only the extra conjunct is redundant.)

uv/private/sdist_configure/detect_native.py
```diff
@@ -418,6 +454,46 @@ def detect(archive_path, context):
             if content:
                 declared.extend(_parse_setup_cfg_build_requires(content))
 
+        # Match the root distribution name rather than path depth: setuptools
```
jbedard, line 498: The two `entry_points.txt` comprehensions repeat the `.egg-info` + filename predicate. Factoring the shared filter and branching only on the discriminator reads a bit cleaner and drops ~10 lines:

```python
egg_info = [
    (name, path) for name, path in member_paths
    if path.name == "entry_points.txt"
    and len(path.parts) >= 2 and path.parts[-2].endswith(".egg-info")
]
if package_name:
    entry_points_paths = [n for n, p in egg_info
        if _normalize_name(p.parts[-2][:-len(".egg-info")]) == package_name]
else:
    depth = 3 if rooted else 2
    entry_points_paths = [n for n, p in egg_info if len(p.parts) == depth]
```

Same behavior, with the two selection strategies sitting side by side.

uv/private/pep517_whl/rule.bzl
```diff
@@ -173,7 +174,10 @@ def _pep517_whl(ctx):
         resource_set = resource_set(ctx.attr),
     )
 
-    return [DefaultInfo(files = depset([wheel_dir]))]
+    return [
+        DefaultInfo(files = depset([wheel_dir])),
+        SourceBuiltWheelInfo(console_scripts = tuple(ctx.attr.console_scripts)),
```
jbedard: This and `_pep517_native_whl` (line 229) end with the identical `[DefaultInfo(files = depset([wheel_dir])), SourceBuiltWheelInfo(console_scripts = tuple(ctx.attr.console_scripts))]`. Minor, but a shared helper (e.g. `_wheel_providers(wheel_dir, ctx)`) removes the copy and keeps the two return sites from drifting.

uv/private/whl_install/rule.bzl
tamird: Done. Removed the dead console_scripts conjunct while retaining the override flag and its explicit-empty semantics.

uv/private/sdist_configure/detect_native.py
tamird: Done. The shared egg-info entry-points filter is now factored once, with distribution-name and rooted/flat fallback selection kept separate; the existing archive-layout regressions continue to pass.

uv/private/pep517_whl/rule.bzl
tamird: Done. Both pure and native paths now use the shared wheel-provider helper with the narrow wheel_dir/console_scripts interface.

tamird:
@codex review

jbedard:
@tamird rebase to resolves conflicts

tamird:
@jason-bedard rebased onto current `main` (`c752dd33`) and resolved the sdist-detector typing and console-script conflicts. Focused detector, wheel-install, PEP 517, snapshot, and sdist-fallback checks pass.

@codex review

jbedard (approved):

PR #1348, opened 2026-07-19
perf(uv): load wheel metadata lazily

zbarsky-openai:
Summary
- Create one whl_install repository per concrete fetched wheel variant.
- Keep the public install repository as a selector, so inactive platform wheels do not have their metadata extracted.

Validation
- bazel run //:buildifier -- /tmp/rules-py-lazy/uv/private/extension/defs.bzl /tmp/rules-py-lazy/uv/private/whl_install/repository.bzl
- bazel build //uv/private/extension:defs //uv/private/whl_install:repository


uv/private/extension/defs.bzl
```diff
@@ -740,15 +752,40 @@ def _uv_impl(module_ctx):
...
+                "name": variant_name,
+                "sbuild": None,
+                "sbuild_console_scripts": [],
+                "whls": json.encode(variant_whls),
+                "whl_files": [whl_target],
+            }
+            if install_cfg.post_install_patches:
+                variant_kwargs["post_install_patches"] = json.encode(install_cfg.post_install_patches)
+                variant_kwargs["post_install_patch_strip"] = install_cfg.post_install_patch_strip
+            exclude_glob = getattr(install_cfg, "exclude_glob", [])
```
tamird, line 775: This creates a selector repository even when the lock has one universal wheel. The accompanying benchmark grows from 451 to 556 loaded packages and 8,677 to 9,097 configured targets without a measured speedup. Can we keep the single-variant path direct and split only genuinely platform-varying installs? Please also port the shared-alias, both-platform, and sdist-fallback regressions from the OpenAI patch.

tamird (changes requested):
The lazy-metadata split matches the OpenAI patch, but the current CI failures are causal and the universal-wheel path now pays the extra repository/target cost without a demonstrated win. Please refresh the affected snapshots/legacy target assumptions and cover the selector behavior before landing.

uv/private/whl_install/repository.bzl
```diff
@@ -685,6 +685,95 @@ filegroup(
...
+                pyc_invalidation_mode = pyc_invalidation_mode_select,
+            )
+            if post_install_patches:
+                fallback_install_attrs += """
+    patches = {patches},
+    patch_strip = {strip},""".format(
+                    patches = repr(post_install_patches),
+                    strip = post_install_patch_strip,
+                )
+            exclude_glob = getattr(repository_ctx.attr, "exclude_glob", [])
```
jbedard, line 713: I don't think this exists yet?

jbedard:
Alternative: https://github.com/aspect-build/rules_py/pull/1356

tamird:
The Bazel 8/9 failures are causal: the extra variant repositories change the observable install roots and invalidate the venv/image snapshots and legacy wheel-chain fixture, while increasing the configured graph. #1356 keeps the stable install tree and avoids that growth, so it is the better metadata boundary. I reviewed its remaining provider/downloader gaps in https://github.com/aspect-build/rules_py/pull/1356#pullrequestreview-4733904155; once those are addressed, this approach should be retired rather than updating snapshots around the path change.

PR #1349, opened 2026-07-19
fix(py_image_layer): support multiple binaries

zbarsky-openai:
Summary
- Preserve each binary launcher path while generating mtree entries.
- Emit renamed launchers for additional binaries instead of only handling the primary binary.

Validation
- bazel run //:buildifier -- /tmp/rules-py-image/py/private/py_image_layer.bzl
- bazel build //py/private:py_image_layer

### DIFF
--- a/py/private/py_image_layer.bzl
+++ b/py/private/py_image_layer.bzl
@@ -585,9 +585,18 @@ def _apply_strip_prefix(sp, strip_prefix, root):
         return "." + root + sp[len(prefix):]
     return "./app.runfiles/_main/" + sp
 
-def _file_to_mtree_entry(f, mode = "0644", strip_prefix = "", root = "/", maybe_symlink = False):
+def _file_to_mtree_entry(
+        f,
+        mode = "0644",
+        strip_prefix = "",
+        root = "/",
+        maybe_symlink = False,
+        executable_short_path = "",
+        executable_dst = ""):
     sp = f.short_path
-    if sp == "_repo_mapping":
+    if executable_dst and sp == executable_short_path:
+        dst = executable_dst
+    elif sp == "_repo_mapping":
         # Bazel synthesizes a top-level `_repo_mapping` runfile (no `_main/`
         # prefix); replicate that placement so runfiles.bash can find it.
         dst = "./app.runfiles/_repo_mapping"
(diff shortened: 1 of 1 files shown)


py/private/py_image_layer.bzl
```diff
@@ -615,15 +624,75 @@ def _file_to_mtree_entry(f, mode = "0644", strip_prefix = "", root = "/", maybe_
...
+            effective_strip_prefix,
+            root,
+            maybe_symlink,
+            binary_short_path,
+            executable_dst,
+        )
+
+    return _source_map
+
+def _dependency_plan_signature(info):
```
tamird, line 681: This signature compares normalized package labels/layer groups and the interpreter group, not the actual wheel-install repository/version or interpreter identity. Two binaries from divergent lock universes can therefore compare equal, after which the secondary discards its own deps/interpreter and runs against the primary's runtime. Please compare the concrete runtime identities (or reject distinct install repos/interpreters) and add an e2e that executes both launchers plus a divergent-plan negative.

py/private/py_image_layer.bzl
```diff
@@ -998,6 +1120,8 @@ _py_image_layer = rule(
 def py_image_layer(
```
tamird, line 1120: `additional_binaries` and `launcher_dir` are inserted before the existing `groups` argument, so existing positional `py_image_layer(...)` calls silently rebind. Please append the new optional parameters after the established public arguments (or make the interface explicitly keyword-only without changing existing calls).

tamird (changes requested):
The multi-binary API needs a correctness guard and real execution coverage before this replaces the local patch. Different lock universes/interpreters can currently pass the plan comparison and then run a secondary launcher against the primary runtime; the public signature also breaks positional callers.

py/private/py_image_layer.bzl
```diff
@@ -964,6 +1077,15 @@ _py_image_layer = rule(
             mandatory = True,
             aspects = [_layer_aspect, _merge_aspect],
         ),
+        "additional_binaries": attr.label_list(
```
jbedard, line 1080: Why not a single `binaries` attribute and the macro can deal with `binary|binaries|additional_binaries` terminology while the real rule just gets the single list?

tamird:
Reimplemented the multi-binary path independently in #1353. It unions concrete wheel, interpreter, and first-party artifacts across binaries, preserves the existing positional API, uses one inner binaries attribute, and executes two launchers across separate lock universes/Python versions. This addresses the three open design/correctness threads here.

jbedard:
Done in #1353

PR #1350, opened 2026-07-19
perf(uv): stream wheel installs

zbarsky-openai:
Summary
- Stream non-script wheel members from the zip archive to disk with copyfileobj.
- Keep script members in memory only when shebang rewriting may be needed.

Validation
- bazel test //py/tools/unpack:unpack_test

### DIFF
--- a/py/tools/unpack/unpack.py
+++ b/py/tools/unpack/unpack.py
@@ -17,6 +17,7 @@ import configparser
 import csv
 import hashlib
 import os
+import shutil
 import subprocess
 import zipfile
 from base64 import urlsafe_b64encode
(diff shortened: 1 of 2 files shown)


py/tools/unpack/unpack.py
```diff
@@ -143,13 +144,15 @@ def install_wheel(version_major, version_minor, into, wheel_path):
...
-            dest.write_bytes(data)
+            if is_script:
+                data = zf.read(member)
+                if _has_python_shebang(data):
+                    _, _, body = data.partition(b"\n")
+                    data = _RELOCATABLE_SHEBANG.encode() + body
+                dest.write_bytes(data)
+            else:
+                with zf.open(info, "r") as source, dest.open("wb") as output:
+                    shutil.copyfileobj(source, output, length=1024 * 1024)
```
tamird, line 155: Please cover a >1 MiB member and a `.data/scripts` Python-shebang case. The former proves this path actually streams across multiple chunks; the latter protects the separate rewrite path while replacing whole-member reads.

tamird (commented):
The streaming change is straightforward and the current suite is green. Could we port the two relevant regressions from the OpenAI change before landing: a member larger than the 1 MiB copy buffer and a `.data/scripts` member with a rewritten Python shebang? Those exercise the behavior this optimization changes.

jbedard (approved):

PR #1351, opened 2026-07-19
perf(uv): reuse wheel RECORD metadata

zbarsky-openai:
Summary
- Parse reusable sha256 and size values from a well-formed wheel RECORD.
- Reuse metadata for untouched installed files and recompute it for rewritten or synthesized files.
- Fall back to hashing when RECORD data is malformed, duplicated, or non-canonical.

Validation
- bazel test //py/tools/unpack:unpack_test

### DIFF
--- a/py/tools/unpack/unpack.py
+++ b/py/tools/unpack/unpack.py
@@ -16,7 +16,9 @@ import argparse
 import configparser
 import csv
 import hashlib
+import io
 import os
+import re
 import shutil
 import subprocess
 import zipfile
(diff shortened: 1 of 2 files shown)


py/tools/unpack/unpack.py
```diff
@@ -201,10 +245,17 @@ def install_wheel(version_major, version_minor, into, wheel_path):
         requested_path = dist_info / "REQUESTED"
         requested_path.write_bytes(b"")
 
+        destination_counts = {}
+        for f, _ in installed:
+            destination_counts[f] = destination_counts.get(f, 0) + 1
+
```
tamird, line 254: A syntactically valid `sha256=...` with the correct size is reused without checking the extracted bytes. A stale or malicious digest therefore produces a false installed RECORD. The new duplicate-row test falls back, but it does not cover one canonical row with a wrong same-size digest. Please add that regression and verify (or explicitly opt into trusting) the source digest.

tamird (changes requested):
Reusing RECORD metadata is useful, but this currently emits an incorrect installed RECORD for a well-formed, same-size entry whose digest does not match the extracted bytes. Wheel installers are expected to verify RECORD hashes during extraction; please either preserve verification or explicitly gate/document the trust tradeoff and cover the canonical stale-digest case.

jbedard (approved):

tamird:
The unresolved RECORD finding is reproducible on merged main: a canonical wrong same-size digest is emitted unchanged, and post-install patches also leave changed/added/deleted files incorrectly represented. Opened #1354 to hash the actual streamed bytes and regenerate RECORD after patches, with regressions for both cases.

PR #1352, opened 2026-07-19
perf(uv): deduplicate wheel installs

zbarsky-openai:
Depends on #1348.

Summary
- Share concrete wheel-variant install repositories across lock universes when metadata, ordered wheel candidates, and exclusions match.
- Keep patched variants project-local and preserve order-sensitive wheel selection.
- Add focused coverage for ordering, exclusions, and patched variants.

Validation
- bazel run //:buildifier -- /tmp/rules-py-dedup/uv/private/extension/defs.bzl /tmp/rules-py-dedup/uv/private/extension/test_defs.bzl /tmp/rules-py-dedup/uv/private/whl_install/repository.bzl
- bazel test //uv/private/extension:shared_whl_variant_key_tests

### DIFF
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -87,6 +87,18 @@ def _deduplicate_whl_files(whls):
         whl_files.append(whl)
     return whl_files
 
+def group_whls_by_target(whls):
+    """Group every wheel filename that resolves to the same fetched label."""
+    grouped = {}
+    for whl_name, whl_target in whls.items():
+        if whl_target:
+            grouped.setdefault(whl_target, {})[whl_name] = whl_target
+    return grouped
+
+def wheel_variant_suffix(whl_target):
+    """Return the stable hash suffix from a generated wheel repository label."""
+    return whl_target.split("__")[-1].split("//")[0]
+
 def parse_declared_console_script(name, entry_point):
     """Canonicalize one override_package console-script declaration.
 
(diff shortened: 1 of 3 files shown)


uv/private/extension/defs.bzl
```diff
@@ -739,16 +762,48 @@ def _uv_impl(module_ctx):
...
+                variant_kwargs["exclude_glob"] = exclude_glob
+            whl_install(**variant_kwargs)
+            variant_installs[whl_target] = "@{}//:install".format(variant_name)
+            if shared_key:
+                shared_variant_installs[shared_key] = variant_installs[whl_target]
         install_kwargs = {
             "metadata_directory": install_cfg.metadata_directory,
             "name": install_id,
             "sbuild": install_cfg.sbuild,
+            "variant_installs": json.encode(variant_installs),
```
tamird, line 802: `exclude_glob` is forwarded to each concrete wheel variant above, but it is absent from this public-selector `install_kwargs`. If resolution falls back to an sdist, the selected install silently keeps files the caller explicitly excluded. Please forward the filter here too and exercise an actual sdist fallback; the current key tests fabricate `exclude_glob` and do not prove the selected install is filtered.

tamird (changes requested):
This contains the full lazy-metadata change and currently repeats its causal CI failures. There is also an integration hole with the approved `exclude_glob` API: wheel variants receive the filter, but an sdist fallback selected through the public repository does not. Please extract/rebase after #1348, preserve the filter for both install paths, and add a real two-lock dedupe plus sdist-filter regression.

PR #1354, opened 2026-07-20
fix(uv): correct compiled and colliding RECORDs

tamird:
Continue trusting producer RECORD hashes for unmodified wheel members so
normal installs keep the fast path. Emit one row per installed destination
when members collide, using the final member's metadata.

Refresh wheel-supplied bytecode entries after compileall can overwrite
them. Patched wheels continue to regenerate RECORD before compilation.


tamird:
@codex review

jbedard:
Provide a real e2e test case reproducing an issue with rela packages.

tamird:
Added the existing cowsay 6.0 post-install-patching case in 7fd2ff36. Its patch changes `cowsay/__init__.py`; both the normal and exposed-venv tests now verify the installed RECORD hash and size. Both fail against current main at that assertion and pass with this change.

@codex review

jbedard (commented):
Nice fix — the two invariants this restores (installed RECORD reflects disk, and RECORD survives post-install patches) match what pip / uv / the PyPA `installer` reference all do: the reference recomputes each entry from the bytes written to disk (`copyfileobj_with_hashing`) rather than trusting the wheel's own RECORD. So this drops a rules_py-specific divergence. A few non-blocking suggestions to reduce duplication, all left inline.

(For context, the one remaining intentional difference from pip/installer is that we don't *verify* wheel content against the shipped RECORD — that's fine since Bazel already pins the wheel's integrity at fetch time.)

py/tools/unpack/unpack.py
```diff
@@ -236,32 +194,34 @@ def install_wheel(version_major, version_minor, into, wheel_path):
...
-            if reusable_record is not None and destination_counts[f] == 1:
-                rows.append((rel, reusable_record[0], reusable_record[1]))
-            else:
-                rows.append((rel, _sha256(f), str(f.stat().st_size)))
-        for meta_file in (installer_path, requested_path):
-            rel = os.path.relpath(str(meta_file), str(site_packages)).replace("\\", "/")
-            rows.append((rel, _sha256(meta_file), str(meta_file.stat().st_size)))
+        for path, (digest, size) in installed.items():
+            rel = os.path.relpath(str(path), str(site_packages)).replace("\\", "/")
+            rows.append((rel, digest, size))
```
jbedard: Consider factoring the RECORD write into a helper:

```python
def _write_record(record_path, rows, site_packages):
    rel = os.path.relpath(str(record_path), str(site_packages)).replace("\\", "/")
    with record_path.open("w", newline="", encoding="utf-8") as fh:
        csv.writer(fh).writerows([*rows, (rel, "", "")])
```

The relativize + append-self-row + write logic is duplicated in the post-install regen block in `main()`. There are now two RECORD generators that must agree on the format contract; a shared writer keeps them from drifting.

py/tools/unpack/unpack.py
```diff
@@ -360,6 +320,25 @@ def main():
...
+        rows = []
+        for path in sorted(args.into.rglob("*")):
+            if not path.is_file() or path in record_paths:
+                continue
+            digest = hashlib.sha256()
+            size = 0
+            with path.open("rb") as source:
+                for chunk in iter(lambda: source.read(1024 * 1024), b""):
+                    digest.update(chunk)
+                    size += len(chunk)
```
jbedard: This read-and-hash loop is the path-based form of the streaming hash in `install_wheel`, and could be a small shared helper:

```python
def _hash_file(path):
    digest = hashlib.sha256()
    size = 0
    with path.open("rb") as f:
        for chunk in iter(lambda: f.read(1024 * 1024), b""):
            digest.update(chunk)
            size += len(chunk)
    return _sha256(digest), str(size)
```

Then this collapses to `digest, size = _hash_file(path)`.

py/tools/unpack/unpack.py
```diff
@@ -360,6 +320,25 @@ def main():
                 "Post-install patch changed observed native files: {}".format(relative)
             )
 
+    if args.patches:
```
jbedard, line 377: Two small things on this block:

1. When patches are present, `install_wheel` computes hashes and writes RECORD, then this immediately rewrites it from a full-tree re-hash — the first RECORD write is wasted. A `write_record=not args.patches` param on `install_wheel` (still writing INSTALLER/REQUESTED, just skipping the RECORD emission) would avoid the double write. Low priority since patching is rare.
2. Worth a comment that `.pyc` are intentionally absent here because `compileall` runs *after* this walk — otherwise a future reader who moves compilation earlier would silently start listing stale bytecode in RECORD.

py/tools/unpack/unpack.py
tamird: Done. Both paths now use the shared RECORD writer, including the RECORD self-row.

py/tools/unpack/unpack.py
tamird: Done. Post-patch regeneration now uses the shared path-based streaming hash helper for digest and size.

py/tools/unpack/unpack.py
tamird: Done. Patched installs now skip the superseded initial RECORD write and row construction while retaining INSTALLER/REQUESTED; the compileall-ordering rationale documents why pyc remains absent. Extraction stays single-path.

tamird:
@codex review

jbedard:
Why can't we trust the hashes from the existing RECORD if we aren't going to patch it?

jbedard:
Can't https://github.com/aspect-build/rules_py/pull/1359 fix the bug, while trusting RECORD hashes and not re-computing them?

tamird:
#1359 fixes the patch case and is cheaper for unpatched wheels, but it does not fix the other bug here: a canonical RECORD digest for different same-size bytes is copied into the installed RECORD unchanged. The lock verifies the wheel archive, not that its producer-generated RECORD describes its members; that case is reproducible and was the unresolved #1351 finding: https://github.com/aspect-build/rules_py/pull/1351#discussion_r3611402197

If trusting producer RECORDs is intentional, #1359 can replace this for the patch bug. I would still cover final RECORD contents for patch additions/deletions and mapped-path collisions. If the invariant is that installed RECORD reflects disk, we still need the streamed hashes.

jbedard:
Can we NOT compute RECORD hashes for files not modified (patched/added/deleted) by rules_py and maintain the existing RECORD values instead?

tamird:
Not reliably with the current patch interface. WhlInstall runs an external patch tool and gets no changed-path list; topology checks only preserve analysis-observed shape, and the tests deliberately use a tool that can rewrite or delete arbitrary paths. Parsing diff headers or comparing stat metadata would miss valid mutations, and deletion/rename behavior is host-dependent.

We can reuse RECORD on the unpatched fast path, but after a patch the generic way to keep hashes correct is to walk and hash the final tree, as #1359 does. Selectively preserving hashes would require restricting the patch format and tracking installed-path relocation, collisions, and final add/delete state. I would not add that complexity without a large-wheel benchmark showing the rehash matters.

jbedard:
@tamird rebase to resolve conflicts

tamird:
@codex review

py/tools/unpack/unpack.py
```diff
@@ -239,36 +230,42 @@ def install_wheel(version_major: int, version_minor: int, into: Path, wheel_path
...
-        rel_record = os.path.relpath(str(record_path), str(site_packages)).replace("\\", "/")
-        rows.append((rel_record, "", ""))
-        with record_path.open("w", newline="", encoding="utf-8") as fh:
-            csv.writer(fh).writerows(rows)
+        if write_record:
+            generated_pyc = _generated_pyc(site_packages) if compile_pyc else set()
+            rows = []
+            for path, (digest, size) in installed.items():
+                if path in generated_pyc:
+                    continue
```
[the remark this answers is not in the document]
tamird: Fixed. The final pass now snapshots bytecode that already exists after patching, runs compileall, then hashes the installed tree while excluding only newly created caches. A supplied cache is therefore retained and rehashed whether compilation overwrites it or the source fails to compile. The legacy-syntax, valid-source, orphan-cache, streaming, and final-rehash cases are covered.

tamird:
@codex review

py/tools/unpack/unpack.py
```diff
@@ -384,6 +373,19 @@ def main() -> None:
...
+    if args.patches or args.compile_pyc:
+        # Hash final installed bytes while excluding only newly generated caches.
+        record_paths = set(site_packages.glob("*.dist-info/RECORD"))
+        generated_pyc = set(site_packages.rglob("*.pyc")) - supplied_pyc
+        rows = []
+        for path in sorted(args.into.rglob("*")):
+            if not path.is_file() or path in record_paths or path in generated_pyc:
+                continue
+            relative = os.path.relpath(str(path), str(site_packages)).replace("\\", "/")
+            rows.append((relative, *_hash_file(path)))
```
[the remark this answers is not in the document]
tamird: Fixed. Unpatched compile-only installs now keep the streamed RECORD, run compileall, and refresh only the rows for bytecode that was already supplied by the wheel; non-bytecode members are not read or hashed again. Patched installs still regenerate the final tree. The focused regression includes a >1 MiB member and verifies that only the two supplied cache files are rehashed.

tamird:
@codex review

tamird:
@codex review

jbedard:
Is this change still necessary after https://github.com/aspect-build/rules_py/commit/c752dd330cb5c622bf255ba6c67b5139d79b9ffe?

tamird:
c752dd33 replaces the post-patch portion. It still trusts canonical same-size RECORD rows on unpatched installs, can emit duplicate rows when two members map to one destination, and hashes wheel-supplied bytecode before compileall can replace it. Those are the remaining regressions covered here: https://github.com/aspect-build/rules_py/pull/1351#discussion_r3611402197

The first is an explicit performance/trust tradeoff for the #1351 fast path. I would keep that fast path and narrow this to the supplied-bytecode/collision fixes; rehashing every fetched wheel merely to defend against malformed producer metadata is not worthwhile.

tamird:
@codex review

jbedard:
Can we add a real example to an e2e test?

tamird:
@jbedard added a real NumPy 1.26.4 case in b9a1d3f8. Its CPython 3.11 wheels ship `numpy/distutils/__pycache__/conv_template.cpython-311.pyc` with a RECORD hash; WhlInstall's compileall replaces that file (the macOS wheel changes from 14,266 to 14,309 bytes), so the pre-fix installer leaves a stale RECORD row. The `uv-data-purelib` e2e now checks the installed digest and size, and both Bazel 8 and 9 e2e jobs pass.

@codex review

py/tools/unpack/unpack.py
```diff
@@ -400,6 +401,19 @@ def main() -> None:
             check=True,
         )
 
+        if supplied_pyc:
+            # compileall can replace bytecode that was already listed in RECORD.
+            for record_path in site_packages.glob("*.dist-info/RECORD"):
```
jbedard, line 422: Can this loop be combined with the loop looking for files effected by patching?

py/tools/unpack/unpack.py
jbedard, line 384: Can we determine which files were patched based on patch content instead of re-hashing _everything_?

py/tools/unpack/unpack.py
```diff
@@ -206,7 +206,7 @@ def install_wheel(version_major: int, version_minor: int, into: Path, wheel_path
             if not member.endswith("/RECORD") and member not in regenerated_markers:
                 if reusable_record is not None and reusable_record[1] != str(info.file_size):
                     reusable_record = None
-                installed.append((dest, reusable_record))
+                installed[dest] = reusable_record
```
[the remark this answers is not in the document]
tamird: Fixed in 25b479758253. Repeated archive member names now invalidate the reusable RECORD row, so the final destination is hashed once. The regression writes two equal-sized fixture/member.py ZIP entries with different bytes, verifies the later bytes win, confirms that path is rehashed, and checks the installed RECORD.

tamird:
@codex review

py/tools/unpack/unpack.py
tamird: The patch input is not a reliable touched-file manifest: --patch-tool may be any executable, so stdin need not be a unified diff and may add, delete, or rewrite arbitrary paths. Parsing headers would miss non-unified formats, deletions, and same-size rewrites; preserving producer hashes for those files would leave RECORD stale. The existing final-tree rehash is the correctness boundary here.

— tamirdex

py/tools/unpack/unpack.py
tamird: Done in 8d627a6dd2f2. The destination-keyed installed map now supplies the initial cache set, and patched installs replace that set during the existing final-tree walk; the extra .pyc walk is gone. Caches compileall can overwrite are hashed once afterward, mapped .data bytecode outside site-packages remains hashed in the patch walk, and the regressions cover compile-only reuse, a newly patched cache, generated-cache exclusion, and the mapped row.

— tamirdex

jbedard (approved):

PR #1356, opened 2026-07-20
perf(uv): extract wheel metadata per-wheel to avoid fetching sibling platform wheels

jbedard:
Each public whl_install repo (@…__pkg__ver) ran _extract_wheel_metadata over
every selectable platform wheel at repo-rule time, peeking RECORD and
entry_points.txt for top-level names, namespace/native layout, and console
scripts. Reading a wheel forces fetching it, and the wheels were declared via a
whl_files label_list, so resolving one package pulled down its Linux, macOS,
Windows, and musl wheels even though only the host's is ever used.

Move the peek into a new per-wheel `whl_dist` repo that downloads one wheel and
carries its RECORD-derived layout as a PyWheelMetadataInfo provider. whl_install
reads the layout off whichever wheel the select chain resolves to, restoring
Bazel's lazy select fetching: only the selected wheel is downloaded and
inspected. Building numpy on macOS/arm64 now fetches one wheel, not all 27.

- add whl_install/metadata.bzl: RECORD/entry_points parsing, derive_layout() +
  the composed extract_install_metadata(), and the exclude_glob matchers
- add whl_install/dist_repository.bzl: the whl_dist repo rule; mirrors
  http_file's downloader contract (get_auth, URL-derived canonical_id +
  DEFAULT_CANONICAL_ID_ENV, discovered checksum for hashless wheels)
- add whl_dist build rule + PyWheelMetadataInfo; whl_install reads the provider
  and drops its per-basename string_list_dict attrs
- source_built_wheel keeps consuming SourceBuiltWheelInfo (pep517-detected
  console scripts) and now provides PyWheelMetadataInfo
- exclude_glob re-derives the selected wheel's layout at analysis time from its
  retained RECORD paths (carried on whl_dist only for excluding packages), so
  removing an initializer reclassifies regular->namespace and never leaves the
  advertised topology out of sync with the filtered install tree
- index the platform-selected `:whl` for Gazelle instead of an arbitrary wheel,
  so lazy fetching never forces an unreachable/incompatible sibling
- strip extraction, whl_files/metadata_directory, and the label-mapping helper
  from whl_install's repo rule
- e2e: uv-lazy-wheel-fetch (lazy fetch + hashless + Gazelle selection) and a
  native package in uv-exclusions; regenerate uv_hub + e2e snapshots

Close #1348

### Test plan

- New test cases added


uv/private/whl_install/rule.bzl
```diff
@@ -63,30 +134,19 @@ def _whl_install(ctx):
...
-        namespace_entries = ctx.attr.namespace_entries.get(whl_basename, [])
-        namespace_dirs = ctx.attr.namespace_dirs.get(whl_basename, [])
-        regular_roots = ctx.attr.regular_roots.get(whl_basename, [])
-        native_roots = ctx.attr.native_roots.get(whl_basename, [])
-        console_scripts = ctx.attr.console_scripts.get(whl_basename, [])
+    # The layout of whichever wheel the `select` chain resolved to for the
+    # active configuration — its own repo's RECORD-derived metadata (or empty,
+    # for a source-built wheel). A sibling platform wheel's surface can't leak
+    # in: it lives in a different repo that is never fetched or consulted here.
+    meta = ctx.attr.src[PyWheelMetadataInfo]
```
tamird, line 160: This unconditionally requires `PyWheelMetadataInfo`, but the unchanged `e2e/cases/patch-failure/BUILD.bazel:35-51` passes the raw `:fake_wheel` genrule to both installs. Both Bazel 8 and 9 jobs now stop here with:

```
<target //patch-failure:fake_wheel> (rule 'genrule') doesn't contain declared provider 'PyWheelMetadataInfo'
FAIL: expected the patch-failure diagnostic on stderr
```

So patch behavior is no longer tested. Wrap the generated wheel once in `source_built_wheel` (its layout is intentionally unknown) and point both installs at that wrapper, or retain the raw-wheel fallback if direct callers are still supported.

uv/private/whl_install/dist_repository.bzl
```diff
@@ -0,0 +1,80 @@
...
+        return ""
+    return "\n    {} = {},".format(name, indent(pprint(list(values)), " " * 4).lstrip())
+
+def _whl_dist_impl(rctx):
+    basename = rctx.attr.downloaded_file_path
+    rctx.download(
+        url = rctx.attr.url,
+        output = basename,
+        sha256 = rctx.attr.sha256,
+    )
```
tamird, line 53: Replacing `http_file` with a bare `rctx.download` drops two existing downloader guarantees. Bazel's `http_file` passes `auth = get_auth(ctx, source_urls)`, so wheel URLs requiring `$NETRC` / `~/.netrc` now fail with 401/403. It also passes the URL-derived default `canonical_id`; without that, a changed URL with a stale hash can reuse the old repository-cache entry locally and only fail on a clean machine.

Please preserve both behaviors (including the default-canonical-ID environment contract) when moving the download into `whl_dist`, and add authenticated-index coverage.

uv/private/whl_install/dist_repository.bzl
```diff
@@ -0,0 +1,80 @@
...
+        top_level_dirs = _attr("top_level_dirs", meta.top_level_dirs),
+        namespace_top_levels = _attr("namespace_top_levels", meta.namespace_top_levels),
+        namespace_entries = _attr("namespace_entries", meta.namespace_entries),
+        namespace_dirs = _attr("namespace_dirs", meta.namespace_dirs),
+        regular_roots = _attr("regular_roots", meta.regular_roots),
+        native_roots = _attr("native_roots", meta.native_roots),
+        console_scripts = _attr("console_scripts", meta.console_scripts),
+    ))
+
+    return rctx.repo_metadata(reproducible = True)
```
tamird: This is not always reproducible. Hashless wheels are explicitly supported (`uv/private/extension/lockfile.bzl:33-41` and the `uv-dep-hashes` fixture), and `defs.bzl:665` passes an empty `sha256` here. The replaced `http_file` captures `download_info.integrity` and returns `attrs_for_reproducibility` when no hash was supplied; this rule discards the download result and always claims reproducibility.

Capture the returned integrity and report it through `repo_metadata(attrs_for_reproducibility = ...)` for the hashless case (or reject unsupported unpinned input). Otherwise a mutable find-links URL can silently change or leave stale bytes/layout cached without a discovered checksum.

uv/private/whl_install/dist_repository.bzl
```diff
@@ -0,0 +1,80 @@
...
+
+def _whl_dist_impl(rctx):
+    basename = rctx.attr.downloaded_file_path
+    rctx.download(
+        url = rctx.attr.url,
+        output = basename,
+        sha256 = rctx.attr.sha256,
+    )
+
+    meta = extract_install_metadata(rctx, rctx.path(basename), _metadata_directory(basename))
```
tamird, line 55: The new repository rule and the central invariant here are not exercised by the tests. The updated analysis tests construct `whl_dist` build targets from local empty files, while the snapshot updates remove all RECORD-derived topology payload for cffi, cryptography, azure, grpcio, and numpy without replacing it with per-wheel snapshots. The performance workload uses a universal wheel or fetches everything before timing, so it also cannot show lazy sibling fetching.

Please add an e2e case with two platform wheels where the inactive URL is deliberately unavailable and the active target still builds, and retain a representative extracted-metadata/provider snapshot (including namespace/native topology). That proves both “only the selected repo is fetched” and that the moved RECORD extraction still produces the advertised layout.

tamird (changes requested):
The per-bdist provider is the cleaner lazy boundary: the existing selector can choose one wheel repo, so universal wheels avoid #1348's extra install repo/selector. The benchmark agrees (101 loaded packages / 301 configured targets unchanged, with analysis about 5.6% faster).

I found three correctness regressions and one important coverage gap:

- both e2e jobs currently fail on the new provider requirement before the patch-failure action runs;
- replacing `http_file` drops its authenticated-download and default canonical-ID behavior;
- supported hashless wheels are now reported reproducible without recording the discovered integrity;
- the tests no longer exercise the repository rule or prove that an inactive wheel stays unfetched.

This supersedes the lazy-metadata half of #1348, but it does not deduplicate `actual_install` actions across lock universes; that remains separate work. The existing downstream lazy/grouping fixtures and their `//file` labels cannot be applied unchanged on top of this design.

jbedard:
@tamird please review again, and mark/comment if your original review issues have been resolved

uv/private/whl_install/rule.bzl
tamird: Addressed at `73c63d19b155`: both patch-failure installs now consume the single `source_built_wheel` wrapper, and the Bazel 8/9 e2e jobs reach and pass the intended patch assertions.

— tamirdex

uv/private/whl_install/dist_repository.bzl
tamird: `get_auth` and the canonical-ID lookup are restored, but the downloader contract is still incomplete: this repository rule does not declare `environ = [DEFAULT_CANONICAL_ID_ENV]`. Bazel 8's `http_file` explicitly declares that environment dependency; without it, changing `BAZEL_HTTP_RULES_URLS_AS_DEFAULT_CANONICAL_ID` need not invalidate/refetch an already evaluated wheel repository, so the selected cache identity can remain stale. Please load `DEFAULT_CANONICAL_ID_ENV`, add it to the repository rule's `environ`, and cover the environment flip.

— tamirdex

uv/private/whl_install/dist_repository.bzl
tamird: This coverage gap remains at `73c63d19b155`. The Bazel 8 and 9 e2e jobs both report `//uv-dep-hashes:test_import SKIPPED`; the passing `test_lockfile_hashless` only checks that the fixture omits a hash and never installs the wheel. The updated analysis tests still construct empty local files with manually supplied metadata, and the azure/cffi snapshots now remove the RECORD-derived topology their unchanged comments claim to pin.

Please add the two-platform case with an unavailable inactive URL, assert/snapshot the selected wheel's extracted namespace/native metadata, and make the hashless import actually run. That would falsify both eager sibling fetching and a broken per-wheel extractor instead of only checking generated plumbing.

Bazel 8: https://github.com/aspect-build/rules_py/actions/runs/29855933807/job/88720125266
Bazel 9: https://github.com/aspect-build/rules_py/actions/runs/29855933807/job/88720125164

— tamirdex

tamird (changes requested):
Re-reviewed because Jason explicitly requested another pass in #ext-aspect-build and asked whether #1356 should replace #1348.

The patch-failure/provider regression is fixed, authenticated download and the canonical-ID lookup are restored, and hashless downloads now return a discovered checksum. Exact-head Bazel 8/9 and the benchmark are green; the graph stays at 101 loaded packages / 301 configured targets and the cold build drops from 11.73s to 7.30s. This is the better boundary: #1348 adds a selector/install repository even for universal wheels, grows the graph, and still has causal e2e failures.

Two issues remain before replacing it:

1. `whl_dist` calls `get_default_canonical_id` but does not declare `DEFAULT_CANONICAL_ID_ENV` in `repository_rule(environ=...)`. Bazel's replaced `http_file` explicitly tracks that environment dependency. Changing the canonical-ID policy can therefore leave an already evaluated wheel repository/cache identity stale.
2. No test falsifies lazy fetching or the moved RECORD extraction. Both e2e jobs skip the hashless import, the analysis tests manually fabricate metadata, and the azure/cffi snapshots no longer pin the topology their comments claim to cover. An active-platform wheel plus a deliberately unavailable inactive URL and a representative extracted-metadata assertion would cover the central invariant.

I marked the fixed patch-failure thread and left the two remaining threads with exact evidence. Once those are addressed, #1356 should supersede #1348 rather than carrying both approaches.

— tamirdex

tamird (changes requested):
Re-reviewed because Jason requested another pass in #ext-aspect-build. The provider regression is fixed, the authenticated/canonical-ID downloader behavior and discovered hash are restored, and the new inactive-wheel/hashless/native-layout case now runs on Bazel 8 and 9. The representative per-wheel snapshot also restores the RECORD-derived topology coverage.

Two items remain:

1. Exact-head CI is causally red: Ruff ANN201 reports `e2e/cases/uv-lazy-wheel-fetch/test_import.py:12:5` (`def main():`). Please make this `def main() -> None:`. https://github.com/aspect-build/rules_py/actions/runs/29869829963/job/88767064347

2. The downloader thread asked for a regression that actually exercises authenticated indexes and flips `BAZEL_HTTP_RULES_URLS_AS_DEFAULT_CANONICAL_ID`. The implementation now declares `DEFAULT_CANONICAL_ID_ENV`, but the new case has neither auth nor an environment flip, so removing either part of that restored contract would still pass. Please add the narrow downloader regression before resolving that thread.

The patch-failure and extracted-metadata/lazy-fetch findings are addressed; I am resolving those two.

— tamirdex

uv/private/whl_install/rule.bzl
```diff
@@ -61,30 +151,38 @@ def _whl_install(ctx):
...
+    # assembly never projects a symlink to filtered-out content. Filtering the
+    # derived entries (rather than re-deriving from RECORD, which now happens
+    # exclude-agnostically per wheel) matches for the load-bearing cases — a
+    # fully-excluded top-level/entry matches its own glob — and is only
+    # conservative for native_roots: a root whose sole native file was excluded
+    # stays listed, so venv assembly declines a merge it could have made (safe).
+    # console_scripts live under bin/, not site-packages, so are untouched.
+    if ctx.attr.exclude_glob:
+        patterns = [parse_exclude_glob(pattern) for pattern in ctx.attr.exclude_glob]
+        top_levels = [e for e in top_levels if not _entry_excluded(e, patterns)]
```
tamird: This filters already-derived roots, not the RECORD entries used to derive them, so exclusions can change package topology without updating the advertised layout. For example, wheel A containing `pkg/__init__.py` and `pkg/a.py`, plus wheel B containing `pkg/b.py`, with `exclude_glob = ["pkg/__init__.py"]`: installation removes A's initializer, but `namespace_top_levels` still omits `pkg`, so venv assembly treats A as a regular package and cannot merge B. Similarly, excluding the only remaining `pkg/only.py` leaves a dangling `pkg` claim, and excluding a source leaves its source-associated `.pyc` metadata advertised.

The previous implementation filtered each RECORD path with `record_path_excluded` before deriving namespace roots. Preserve the retained RECORD paths in the selected-wheel provider and derive topology after exclusions, then cover an excluded initializer/source `.pyc` with two overlapping wheels.

— tamirdex

e2e/cases/uv-lazy-wheel-fetch/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+# discovered-checksum path.
+load("@aspect_rules_py//py:defs.bzl", "py_test")
+
+py_test(
+    name = "test_import",
+    srcs = ["test_import.py"],
+    dep_group = "uv-lazy-wheel-fetch",
+    main = "test_import.py",
+    python_version = "3.11",
+    deps = ["@pypi-lazy-wheel-fetch//charset_normalizer"],
```
tamird, line 30: This only builds the import target, but Gazelle uses a different path that still bypasses platform selection: `uv/private/whl_install/repository.bzl:244-268` sets `gazelle_index_whl = prebuilds.values()[0]`, and `uv_project`/`gazelle_python_manifest` aggregate and build that arbitrary wheel. Thus Gazelle on macOS fetches the first Linux wheel; if an unreachable/incompatible wheel is first, Gazelle fails despite the selected host wheel being valid.

Use the platform-selected `:whl` target for `gazelle_index_whl`, put the unreachable inactive wheel first in this fixture, and build the Gazelle manifest as well as the import target.

— tamirdex

tamird (changes requested):
Re-reviewed because Jason explicitly re-requested review.

The patch-failure provider regression, canonical-ID environment declaration, hashless download, inactive-platform/native import regression, and prior Ruff failure are fixed; exact-head CI is green. The patch-failure and prior lazy-wheel/extracted-metadata threads are addressed, although repository permissions prevent me from resolving them directly.

Two correctness regressions remain: exclusions are applied after wheel topology has already been derived, and Gazelle still indexes an arbitrary platform wheel instead of the selected wheel. The existing downloader thread also still lacks the requested authenticated-index/canonical-policy-flip regression.

— tamirdex

xangcastle (approved):

tamird (approved):
Reviewed at Jason’s explicit request. The exact head fixes platform-selected Gazelle indexing and re-derives excluded-wheel topology from retained RECORD entries. The unreachable inactive-wheel regression now exercises the selected Gazelle wheel and native import; authenticated downloads, canonical-ID invalidation, and hashless integrity are preserved. Exact-head CI is green.

— tamirdex

PR #1357, opened 2026-07-20
refactor!: remove unstable_configure_command, remove unstable_ prefix from annotate_packages

jbedard:

BREAKING: remove `unstable_configure_command`, remove `unstable_` prefix from `annotate_packages`

### Test plan


jbedard:
@tamird can confirm if you are using these?

tamird:
OpenAI does not use `unstable_configure_command`. It does use `uv.unstable_annotate_packages` eight times in `MODULE.bazel` (the main lock and seven additional universes), so the rename will require a mechanical update in the downstream bump. The new `uv.annotate_packages` API covers those calls.

e2e/cases/uv-invalid-build-overrides/custom-complete/MODULE.bazel
```diff
@@ -1,33 +0,0 @@
...
-)
-
-uv = use_extension("@aspect_rules_py//uv:extensions.bzl", "uv")
-uv.declare_hub(hub_name = "custom_complete_invalid_overrides")
-uv.project(
-    default_build_dependencies = [],
-    hub_name = "custom_complete_invalid_overrides",
-    lock = "//:uv.lock",
-    pyproject = "//:pyproject.toml",
-    # sdist_build reparses command strings from its generated repository.
```
tamird, line 16: This deletes the only complete-custom-BUILD fixture that proves pre-build patch context reaches `configure_command` and that `resource_set`, `env`, `monitor_memory`, and `toolchains` are rejected when `build_file_content` bypasses the generated wheel rule. The `sdist_build` branch and the documented constraints remain live, and #1347 extends this exact fixture/contract for explicit-empty console scripts. Could we retain or move a focused repository-rule regression, or remove the now-unreachable custom-content machinery/docs consistently?

tamird (changes requested):
The API cleanup is fine for OpenAI, but this removes the only coverage for a still-live and documented custom-BUILD/pre-build-patch path. Please either retain a focused repository-rule regression or remove that unreachable machinery and its documented contract together.

tamird:
I traced reachability further: after removing `uv.project(...unstable_configure_command)`, the UV extension can only pass the bundled detector, which never emits `build_file_content`, and I found no direct repository-rule consumers. The smaller option is to delete the now-dead `build_file_content`/configure-context branch and its documentation/validation, then drop #1347's corresponding custom-complete changes. Retaining a regression only makes sense if the lower-level hook is intended to remain supported.

e2e/cases/uv-invalid-build-overrides/custom-complete/MODULE.bazel
tamird: The code deletion now looks consistent. Two obsolete `build_file_content` passages remain in `docs/uv-patching.md`: the monitoring paragraph around lines 183-186 and the constraints bullet around lines 220-224 that still promises `pre_build_patches` in the configure context. Please delete both.

tamird (changes requested):
Re-reviewed at Jason's request after the custom-BUILD path was removed. The unstable hooks, obsolete fixtures, documentation passages, and all annotate_packages call sites are otherwise coherent, and CI is green.

One piece of the deleted interface remains. Pre-existing in uv/private/sdist_build/repository.bzl:24-25, _write_context_file still serializes pre_build_patches and pre_build_patch_strip for the configure tool; uv/private/sdist_configure/defs.bzl:27-39 still documents that context as input for generating build-file content. The bundled detector only declares those fields and never reads them, and this PR removes the only custom/complete-BUILD consumer. Patches are applied by the generated wheel rule, so this is now dead contract/plumbing. Please delete both fields and update the context description together.

This should land before #1347 is rebased: that PR still modifies the now-deleted custom-complete fixture, which should be dropped rather than restored.

[tamirdex]

tamird (approved):
Re-reviewed because Jason requested another pass in #ext-aspect-build. The cumulative change now removes the unreachable custom-BUILD path, obsolete configure-context fields, fixtures, and documentation together; every `annotate_packages` call site and error is renamed consistently. Exact-head CI is fully green (86 checks). LGTM.

#1347 will need to rebase after this lands and drop its changes to the deleted custom-complete fixture.

— tamirdex

PR #1359, opened 2026-07-20
fix: update and correct RECORD hashes for files changed via patches

jbedard:
### Test plan

- New test cases added

### DIFF
--- a/e2e/cases/uv-patching-829/__test__.py
+++ b/e2e/cases/uv-patching-829/__test__.py
@@ -1,17 +1,53 @@
 #!/usr/bin/env python3
 
-"""Test that post-install patches are applied to cowsay."""
+"""Test that post-install patches are applied to cowsay and tracked in RECORD."""
+
+import csv
+import hashlib
+from base64 import urlsafe_b64encode
+from pathlib import Path
 
 import cowsay
 
-# Verify the post-install patch added PATCHED_POST_INSTALL to __init__.py
+# Verify the post-install patch modified __init__.py.
 assert hasattr(cowsay, "PATCHED_POST_INSTALL"), (
     "Post-install patch was not applied: cowsay.PATCHED_POST_INSTALL is missing"
 )
 assert cowsay.PATCHED_POST_INSTALL is True
 print("post_install patch: OK")
 
-# Verify cowsay still works after patching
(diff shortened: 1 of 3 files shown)


jbedard:
@tamird can this replace #1354? Is this change more efficient?

e2e/cases/uv-patching-829/__test__.py
```diff
@@ -11,6 +16,15 @@
 assert cowsay.PATCHED_POST_INSTALL is True
 print("post_install patch: OK")
 
+cowsay_init = Path(cowsay.__file__)
+record_path = next(cowsay_init.parent.parent.glob("cowsay-*.dist-info/RECORD"))
+with record_path.open(newline="", encoding="utf-8") as record:
+    rows = {path: (digest, size) for path, digest, size in csv.reader(record)}
+content = cowsay_init.read_bytes()
+digest = urlsafe_b64encode(hashlib.sha256(content).digest()).decode().rstrip("=")
+assert rows["cowsay/__init__.py"] == (f"sha256={digest}", str(len(content)))
```
tamird: This proves the rewritten row is updated. Can we also exercise a safe patch that adds and deletes files and assert that the added row exists and the deleted row is absent? Those are the other two cases the regeneration branch explicitly claims to handle.

tamird (commented):
The patch rewrite path is correct, but the regression only covers modifying an existing file. Please also cover final RECORD contents when a safe patch adds or deletes a file. Distinct wheel paths can also map to one installed destination; that collision still produces duplicate RECORD rows in the reuse path and is worth a regression if producer RECORD reuse remains intentional.

zbarsky-openai (approved):

tamird (approved):
The updated regression now covers rewritten, added, and deleted files and correctly handles GNU/BSD patch deletion behavior. The final RECORD regeneration matches the installed tree.

e2e/cases/uv-patching-829/__test__.py
tamird: Confirmed at 5274b1f: the updated patch now adds cowsay/patched_added.py and deletes cowsay/tests/test_api.py, and the test verifies the added digest plus the GNU/BSD deletion outcomes. This addresses the requested coverage.

PR #1363, opened 2026-07-23
feat: Add cross-compilation support for sdist packages with native extensions

xangcastle:
Enable pep517_native_whl to cross-compile Python sdists containing C or C++ extensions when the target platform differs from the exec host. 
The native_build_toolchain sentinel is now optional, so a missing resolution signals cross-compilation mode and falls through to a user-registered cross CC toolchain (such as `toolchains_llvm`) instead of hard-failing. 
A new `cc_layer.bzl` module extracts compiler paths, `CFLAGS`, and `LDFLAGS` from the resolved cross CC toolchain at analysis time, and build_helper.py generates compiler wrappers that re-inject -target/--sysroot identity flags, filter incompatible linker flags from the exec host, and override LDSHARED to use the cross CC toolchain's link flags. The target interpreter's `_sysconfigdata` is loaded via `_PYTHON_SYSCONFIGDATA_NAME` so setuptools produces correct `EXT_SUFFIX` and `SOABI` tags, and a post-build check validates the wheel's platform tag against the target before accepting the output.

---

### Test plan

- New test cases added


uv/private/pep517_whl/rule.bzl
```diff
@@ -190,6 +222,32 @@ def _pep517_native_whl(ctx):
     wheel_dir = ctx.actions.declare_directory("whl")
     patch_args, patch_inputs = _patch_args_and_inputs(ctx)
 
+    eg_toolchains = ctx.exec_groups[_TARGET_EXEC_GROUP].toolchains
+    cross = eg_toolchains[NATIVE_BUILD_TOOLCHAIN] == None
```
tamird: Absence of an optional native-build toolchain does not establish that the execution and target platforms differ. A native build with that registration missing now enters cross mode, changing existing native build behavior and wheel semantics. Derive this decision from the actual execution and target Python/platform identities. rules_pycross already models those identities explicitly. [0]

[0] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/toolchain.bzl#L33-L55

— tamirdex

uv/private/pep517_whl/cc_layer.bzl
```diff
@@ -0,0 +1,193 @@
...
+        flags[key] = " ".join([_absolutize_flag(f) for f in raw])
+
+    needs_pic = cc_common.action_is_enabled(
+        feature_configuration = feature_configuration,
+        action_name = ACTION_NAMES.cpp_link_dynamic_library,
+    )
+
+    target_os, target_cpu = get_target_platform(ctx)
+
+    return struct(
```
tamird, line 182: This returns compiler tools and flags, but not the CcInfo headers, dependent static/shared libraries, or target C++ runtime that a real source-built extension requires. The wheel action cannot materialize that dependency closure; inspecting the toy geohash ELF cannot establish that the resulting extension links or loads. The existing pycross layer explicitly gathers all three. Please compose with that layer or explain and test the real alternative. [0]

[0] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/private/build/actions/cc_layer.bzl#L126-L182

— tamirdex

uv/private/pep517_whl/rule.bzl
```diff
@@ -213,12 +271,56 @@ def _pep517_native_whl(ctx):
...
+        ]
+
+        py_toolchain = ctx.toolchains[PY_TOOLCHAIN]
+        if py_toolchain != None:
+            runtime = getattr(py_toolchain, "py3_runtime", None)
+            if runtime:
+                sysconfig_file = _find_sysconfigdata(runtime)
+                if sysconfig_file:
+                    extra_inputs.append(depset([sysconfig_file]))
+                    env["RULES_PY_TARGET_SYSCONFIGDATA"] = sysconfig_file.path
```
tamird, line 352: Copying one target _sysconfigdata file into a build still executed by the host interpreter does not create a target Python environment. Build backends continue to observe host interpreter identity, packaging tags, and platform behavior; the later validation checks only the final platform segment, so a host cp313 ABI can pass for a cp312 target with the same OS and CPU. pycross carries both interpreters and builds the appropriate cross environment instead. [0]

[0] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/private/build/actions/pep517_action.bzl#L81-L103

— tamirdex

uv/private/pep517_whl/rule.bzl
```diff
@@ -156,6 +157,37 @@ def _cc_toolchain_inputs_and_tools(ctx):
...
+    """Derive _PYTHON_HOST_PLATFORM from target platform constraints.
+
+    Linux: libc does not affect the platform string — always linux-{cpu}.
+    macOS: uses arm64 (not aarch64) and requires a version component.
+    """
+    if target_os == "linux":
+        return "linux-" + _PYTHON_CPU_MAP.get(target_cpu, target_cpu)
+    if target_os == "darwin":
+        cpu = "arm64" if target_cpu == "aarch64" else target_cpu
+        return "macosx-11.0-" + cpu
```
tamird, line 196: The macOS deployment version is a property of the target interpreter/SDK, not always 11.0. Hard-coding it produces an incorrect compatibility tag for targets requiring another deployment version. Derive the platform and deployment target from target sysconfig, as pycross already does, and add an actual Darwin cross-build regression. [0]

[0] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/private/build/tools/utils/sysconfig_utils.py#L97-L169

— tamirdex

uv/private/pep517_whl/build_helper.py
```diff
@@ -270,6 +418,58 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
...
+    if expected_os not in platform_tag:
+        print(
+            "Error: wheel platform tag '{}' does not contain target OS '{}'.".format(
+                platform_tag, expected_os,
+            ),
+            file=sys.stderr,
+        )
+        exit(1)
+
+    if expected_cpu not in platform_tag:
```
tamird, line 902: A valid Windows x86-64 wheel has platform tag win_amd64, but _expected_cpu_in_tag returns x86_64 here. Consequently every such wheel fails this validation despite being correct; Windows x86 has the analogous win32/i686 mismatch. Conversely unrecognized target OS/CPU values silently skip validation above. Derive supported wheel tags from actual target metadata and fail closed for unsupported targets.

— tamirdex

e2e/cases/uv-deps-650/crossbuild/BUILD.bazel
```diff
@@ -158,5 +160,119 @@ build_test(
...
+    outs = ["_geohash_arm64_so_arch.ok"],
+    cmd = """
+tar_file=$$(echo $(SRCS) | tr ' ' '\\n' | grep 'geohash_layers.*\\.tar' | head -1)
+$(location check_so_arch.sh) $$tar_file b700
+touch $@
+""",
+    tools = [":check_so_arch.sh"],
+)
+
+build_test(
```
tamird: These new checks prove the packaged Linux ELF architecture and filename, but never import or execute the cross-built extension. They therefore pass if the Python ABI is wrong or a target C/C++ dependency is absent at runtime. Add a target-side load/runtime regression with a real native dependency, and cover native-toolchain absence and a non-Linux target, before treating these checks as proof of general cross-compilation.

— tamirdex

tamird (changes requested):
Reviewing this draft at the author's explicit request.

rules_pycross already provides the native CC layer, separate execution and target Python interpreters, PEP 517 cross-build environment, and target-aware sysconfig that this change reimplements. Its v2 alpha also exposes those primitives through a public backend integration API. [0] [1] [2] [3]

The proposed implementation instead infers cross-compilation from an absent optional native toolchain, omits target native dependency and runtime closures, hard-codes macOS deployment tags, and rejects valid Windows wheels. The Linux-only fixture checks the ELF architecture and filename without proving that a cross-built native extension actually loads.

Please first evaluate integration with rules_pycross, including the v2 alpha's transitive-dependency compatibility. If that integration is not viable, document the actual blocker and propose the smallest reusable upstream interface before introducing a parallel cross-compilation implementation. [4]

[0] https://github.com/jvolkman/rules_pycross/blob/v2.0.0-alpha.2/pycross/backend.bzl#L63-L83
[1] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/toolchain.bzl#L33-L55
[2] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/private/build/actions/cc_layer.bzl#L126-L182
[3] https://github.com/jvolkman/rules_pycross/blob/74ee87c7d8eab76a673c07df3ec9e55d1e629e49/pycross/private/build/tools/utils/sysconfig_utils.py#L97-L169
[4] https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/rules_pycross/2.0.0-alpha.2/MODULE.bazel

— tamirdex

tamird (changes requested):
Re-reviewing because Jason explicitly requested review of 9f3b372131b6.

The new execution transition still selects host_platform and repository-time host libc, rather than the wheel action’s actual execution platform. The associated frontend test removes its execution-platform assertion. This cannot build correctly under heterogeneous remote execution; the new documentation itself confirms that the design assumes exec == host.

The five existing review threads remain current and unanswered. The replacement CC layer does not stage transitive CcInfo headers, native libraries or runtime dependencies; the PEP 517 frontend still runs the host interpreter with only a target sysconfig file; macOS wheel tags still hard-code deployment target 11.0; and wheel validation still rejects valid win_amd64 tags. The Linux fixture checks one patched extension’s ELF and filename but never executes the cross-built artifact or verifies a native dependency, target Python ABI, Darwin or remote worker.

rules_pycross already publicly exports target-aware CC extraction, PEP 517 actions, wheel repair and execution-platform transitions. Please evaluate composing those existing primitives before maintaining a second incomplete cross builder. If a dependency or compatibility issue prevents reuse, identify it and add real target-runtime and remote-execution coverage before calling this general cross-compilation.

https://github.com/jvolkman/rules_pycross/blob/v2.0.0-alpha.2/pycross/backend.bzl#L28-L39
https://github.com/jvolkman/rules_pycross/blob/v2.0.0-alpha.2/pycross/backend.bzl#L63-L83

— tamirdex

PR #1370, opened 2026-07-24
perf(uv): deduplicate wheel installs

jbedard:
Close #1352

Performance: share wheel installs across lockfiles when equivelent

### Test plan

- New test cases added


jbedard:
@tamird (the robot) looking-for-review

tamird (approved):
Reviewing at Jason’s explicit request.

The wheel-install identity preserves ordered content-addressed wheels and exclusion policies while refusing to share source builds, patches, extra dependencies, or extra data. Both project SCC references and source-build dependencies are rewritten before duplicate repositories are removed, so platform and runfiles resolution continue to target the canonical install.

The three-universe end-to-end fixture measures the actual reduction from 26 reachable install repositories to 12, verifies both matching and different exclusion policies, and runs imports from each project. All 86 exact-head checks pass.

— tamirdex

xangcastle (approved):

PR #1371, opened 2026-07-25
feat: support PEP 427 data

### LINKED ISSUE #1366: [Bug]: v2 uv venv omits wheel data_files (share/…), breaking sys.prefix/share data discovery (e.g. nbconvert templates)
### What happened?

Under `aspect_rules_py` `2.0.0-alpha.5`, the `py_venv` generated for a target symlinks only the wheel install's `lib/.../site-packages` into the venv — **not** the install's `share/` tree. So a wheel's **data files** (the RECORD `*.data/data/` category, which uv installs to `<wheel-install-root>/share/…`, a sibling of `lib/`) are present in the `whl_install` output but unreachable on the venv's `sys.prefix`/`share` path. Tools that resolve data relative to `sys.prefix/share` therefore can't find it.

**Concrete repro — nbconvert.** nbconvert ships its Jinja templates as data files under `share/jupyter/nbconvert/templates/`. Under v2 they land at:

```
…/aspect_rules_py++uv+whl_install__…__nbconvert__7_17_1/actual_install.install/
├── bin/
├── lib/python3.13/site-packages/nbconvert/      # symlinked into the venv
└── share/jupyter/nbconvert/templates/python/…   # NOT symlinked into the venv
```

The generated venv has a `lib/.../site-packages` but no `share/`, so `nbconvert.exporters.PythonExporter()` fails:

```
ValueError: No template sub-directory with name 'python' found in the following paths:
	<venv>/share/jupyter
	~/.local/share/jupyter
	/usr/local/share/jupyter
	/usr/share/jupyter
KeyError: 'template_paths'
```

Any `jupyter_core`-based tool that reads `<sys.prefix>/share/jupyter` is affected (nbconvert, jupyter, etc.). This works under `rules_python`'s `pip.parse` (and pip/conda), which lay the data out where `sys.prefix/share` discovery finds it.

### Version

`2.0.0-alpha.5`

### How to reproduce

1. `uv.project` hub with `nbconvert` (7.x) in the lock.
2. A `py_test`/`py_binary` that does `from nbconvert.exporters import PythonExporter; PythonExporter().from_notebook_node(...)`.
3. Fails with `KeyError: 'template_paths'` — the `share/jupyter` templates are absent from the venv (they ARE present in the `whl_install`'s `actual_install.install/share/`).

### Anything else?

The data is installed — it's just not surfaced on the venv path, and there's no venv-level knob (that I can find) to include a wheel's `share/` data. Our workaround is to resolve the venv symlink back to the wheel-install root and point the tool at `<install-root>/share/jupyter`, but that's tool-specific and relies on internal layout. It'd be great for the venv to expose wheel `data_files` (at least `share/`) on `sys.prefix` the way pip/rules_python do.

jbedard:
Fix #1366

Data-file paths are extracted from each wheel's RECORD and venv assembly now projects each wheel's PEP 427 .data/data/ tree (e.g. share/jupyter/…, etc/…) into the venv.

### Test plan

- New test cases added


tamird:
Reviewing this for the rules_py migration: at 1fdd483b0, the removed-file case is covered, but `py/tools/unpack/unpack_test.py` explicitly accepts a patch adding `share/added.bin`. That test checks only the unpacked tree. `uv/private/whl_install/rule.bzl` still forwards the pre-patch `meta.data_files`, so venv assembly never projects the added file under `sys.prefix`; the patch succeeds but its data is missing from the actual venv.

Could we derive and forward the post-patch data-file list and add an end-to-end regression that reads the added file through `sys.prefix`? If added data files are intentionally unsupported, the patch should fail instead of silently omitting them.

[ tamirdex ]

xangcastle (approved):

jbedard:
@tamird will you have any issues with this at openai? Does this feature need an opt-out or better exclusion API?

py/private/py_venv/venv.bzl
```diff
@@ -155,6 +155,31 @@ def assemble_venv(
...
+    # into the prefix (e.g. `share/jupyter/...`, `etc/...`) is symlinked at
+    # `<venv>/<path>` back into the owning wheel's install tree, so tools that
+    # discover resources via `sys.prefix/share` find them. Per-file (not
+    # per-dir) so wheels contributing to a shared prefix directory union
+    # instead of one shadowing the other. The install-tree root is the wheel's
+    # site-packages rfpath minus its fixed `lib/<pyver>/site-packages` suffix.
+    sp_suffix = "/lib/{}/site-packages".format(wheel_py_ver)
+    install_root_by_sp = {}
+    for datapath, wheel_site_pkgs in data_file_to_site_pkgs.items():
+        out = ctx.actions.declare_symlink(venv_name + "/" + datapath)
```
tamird, line 172: A PEP 427 data path is allowed anywhere beneath the prefix, so data_scheme_segments accepts bin/python, bin/activate, pyvenv.cfg, and lib/pythonX.Y/site-packages/... . This unconditionally declares each as a new output, but the same function later declares the interpreter, activation script, configuration, console scripts, and projected package outputs. Bazel reports conflicting generating actions before package_collisions can resolve anything. Please reserve virtual-environment-owned paths and resolve data-versus-console-script and data-versus-package claims in the existing collision planner; add a wheel fixture that actually assembles a venv with one conflicting prefix path.

[tamirdex]

tamird (changes requested):
Reviewed at explicit request against the complete current change at 4e0c81d7179de70c34911800dec2374b6a71fe09. The previously reported post-install data-file inconsistency is now properly rejected and covered. One blocking issue remains: prefix data is projected without checking paths already owned by the assembled virtual environment. A valid data entry such as bin/python, bin/activate, pyvenv.cfg, an existing console script, or a projected site-packages path makes Bazel declare the same output twice. This bypasses package_collisions and turns an ordinary wheel installation into an analysis failure. Please resolve reserved and cross-kind prefix collisions centrally and add a real venv regression; restricting this feature to one project or adding an opt-out would not correct the ownership boundary.

[tamirdex]

uv/private/whl_install/metadata.bzl
```diff
@@ -77,6 +77,40 @@ def site_packages_segments(path, data_directory):
...
+_VENV_OWNED_ROOTS = ["pyvenv.cfg", "bin", "lib"]
+
+def data_segments_contained(segments):
+    """True when the prefix-relative segments are safe to project.
+
+    Data paths become `ctx.actions.declare_symlink` outputs under the prefix, so
+    an empty, `.`, `..`, or absolute component must not be projected, nor may
+    the path fall in a venv-owned root.
+    """
+    if segments[0] in _VENV_OWNED_ROOTS:
```
tamird: This duplicates the reserved-root policy in virtuals_resolvers.bzl, but here it silently removes valid RECORD entries before resolve_wheel_collisions can observe them. A normal prebuilt wheel containing .data/data/bin/python or .data/data/pyvenv.cfg therefore produces no data-file collision, even with package_collisions = error; the new reserved_paths_test passes only because py_unpacked_wheel injects data_files directly. It also makes post-install verification compare filtered expected metadata against the actual installed prefix. Please let data_segments_contained validate containment only, forward valid reserved paths, and let _resolve_data_files apply the single documented collision policy. Add a RECORD-backed whl_install regression in addition to the direct-provider fixture.

[tamirdex]

tamird (changes requested):
Re-reviewed at the author’s explicit GitHub review request against c51c580d1dfe42699470f9c71a58a5172de9d156. The new centralized collision planner, reserved-path handling, nested-file detection, and assembled-venv regression address the original conflicting-output finding. One remaining issue prevents approval: prebuilt-wheel metadata independently discards the same reserved data paths before they reach that planner. Consequently the normal uv installation path silently bypasses package_collisions, while the regression exercises only py_unpacked_wheel, which supplies the data list directly. Keep path-containment validation in metadata extraction, but pass valid prefix paths to the single collision owner and extend the regression to an actual RECORD-backed prebuilt wheel.

[tamirdex]

jbedard:
@codex

jbedard:
@codex

jbedard:
@codex

tamird (approved):
Re-reviewed at the author’s native request against c2ae1d613c4a94f2019f4c85caa910f973fb5311. The RECORD-backed prebuilt-wheel path now forwards all contained prefix entries to the single collision planner, including virtual-environment-owned paths; analysis tests verify the real selected-wheel provider and collision policy. Runtime regressions cover shared Jupyter data, reserved interpreter and configuration paths, nested collisions, and PEX parity. The post-install manifest prevents patch drift without overflowing argument limits. My previous findings are addressed; no project-specific opt-out is required. The explicitly documented source-built-wheel limitation remains separate from this change.

[tamirdex]

PR #1379, opened 2026-07-27
feat(uv): first-class CcInfo dependencies for native sdist builds

sallustfire:
`uv.override_package(env = ..., toolchains = ...)` lets a native sdist build reach a C/C++ dependency, but only through the raw make-variable escape hatch: the user hand-writes `CPPFLAGS`/`LDFLAGS`, anchors every include and archive path with `$(EXECROOT)`, and re-derives the dependency's transitive closure and link order by hand, then keeps all of it current as the dep graph evolves. That is `cargo_build_script`'s make-var bus without the other half; the equivalent of putting a `cc_library` in a `rust_binary`'s `deps` is still missing.

This PR adds that other half, as discussed in #1233. `cc_deps` takes targets that provide `CcInfo` and flattens them for the PEP 517 build:

```starlark
uv.override_package(
    name = "native-package",
    cc_deps = ["//third_party/mylib"],  # a cc_library / cc_import
)
```

At analysis time the rule merges the deps' `CcInfo`, derives compile flags (includes, quote/system includes, defines, Apple framework search paths) and a link plan (each linker input's PIC-preferred static archive as a full path, in topological order), anchors every path with the existing execroot marker, and writes one structured params file passed to the build helper. At build time, before the backend changes into the unpacked source tree, the helper resolves the marker against the real execroot and routes the inputs: compile flags append to `CPPFLAGS`, archives become `[build_ext] link_objects` and `-l` names become `[build_ext] libraries` in a config file injected through setuptools' `DIST_EXTRA_CONFIG` (both post-object slots, which is what single-pass linkers need), and the small set of accepted residual link flags appends to `LDFLAGS`. A package's own `setup.cfg [build_ext]` values are folded back in front of ours because `DIST_EXTRA_CONFIG` replaces rather than merges same-named options.

The three conditions from the #1233 thread are addressed directly:

- **Setuptools-specific behavior stays scoped to that backend.** Everything setuptools about this feature lives behind the detected-backend gate in the build helper, reusing the existing backend set (including the legacy `setup.py` path). `cc_deps` on any other backend is a hard error, never a silent no-op. The analysis side is backend-neutral: it only emits slots, so a future meson-python or scikit-build adapter is a runtime-only change.
- **The configured C++ driver and EXECROOT handling are preserved.** No `cc_deps` code path sets `CC`/`CXX`/`AR`/`LD`/`STRIP` or touches the tool selection, companion inference, or wrapper generation; injected flags ride through the wrappers, and paths use the same marker mechanism as `$(EXECROOT)`. With `cc_deps` empty the build action is byte-identical to before (verified by ActionKey).
- **An executable regression proves both paths survive the backend chdir** (details below).

## Link flags: a deliberate fail-closed contract

A dependency's `linkopts` cannot be forwarded blindly: the two reachable setuptools slots split them into pre-object `LDFLAGS` and the post-object libraries list, so relative order between `-l` entries and other flags cannot be preserved. Rather than silently reordering (or maintaining a denylist of known-dangerous flags, which fails open for anything unenumerated), `cc_deps` accepts a small allowlist of position-insensitive shapes: `-pthread`, glued `-L<dir>`, `-Wl,-rpath`/`-rpath-link`/`--version-script` (comma or `=` forms), `-Wl,-z` with a reviewed keyword set (`relro`, `now`, `noexecstack`, `origin`), and `-Wl,--enable-new-dtags`. `-Wl,` tokens are validated per directive, so `-Wl,-z,relro,-z,now` works while `-Wl,-rpath,/x,--as-needed` is rejected naming the smuggled directive. Everything else fails at analysis with three workarounds in the message: repeat the library name for archive cycles (`-la -lb -la`; `-l` order is preserved within the libraries slot), set global toggles such as `--as-needed` in the override's `env` `LDFLAGS`, and use `pre_build_patches` for anything beyond that. `-framework` is deliberately absent: ld64 resolves frameworks as library inputs in command-line order, which the slot split cannot preserve. Starting fail-closed means later changes are relaxations, never breakage.

Other limitations, each enforced with an actionable error: setuptools `>= 65.4.0` in the build environment (the release that added `DIST_EXTRA_CONFIG`); static or PIC-static archives only (dynamic-only and `alwayslink` deps fail at analysis); archives must contain PIC objects since they are folded into a shared object; and each `cc_deps` label must be visible to the generated build repository.

## Regression tests

The executable regression at `uv/private/pep517_whl/tests/cc_deps/executable` builds a real `setuptools.build_meta` sdist whose extension spans a two-level `cc_library` chain, with a header that exists only in the `cc_deps` include path and a transitive symbol that resolves only through the post-object archives. A `py_test` extracts the built wheel, imports the module, and calls through the chain. Both halves were red first: without the include path the compile fails with `'dep.h' file not found`, and with the transitive archive withheld the import fails with `undefined symbol: dep2_value`, the transitive leaf, proving closure and post-object ordering together. The fixture also proves the `-l` slot end to end: the sdist's own `build_clib` compiles a static library hermetically in-build and the dep's `-lextra` linkopt is the only thing naming it on the link (setuptools' automatic clib linking is suppressed in the fixture precisely because it otherwise masks the slot; that vacuity was caught by demonstrating red before wiring green), plus a genuine archive cycle resolved through the documented repeat-the-library workaround. Defines propagate transitively and are consumed by an `#error` guard in the backend-compiled source.

Analysis tests pin the params-file contract (marker anchoring, topological archive order, the `-l` partition, per-directive allowlist validation with a golden copy so the accepted set cannot drift from its tests) and the fail-clear paths (dynamic-only, alwayslink, rejected flags naming the exact directive). A subprocess harness drives the build helper through every runtime gate: the non-setuptools backend error, the setuptools floor via a fake dist-info, pre-existing `DIST_EXTRA_CONFIG`, `SETUPTOOLS_USE_DISTUTILS`, whitespace guards, marker survival, and the `setup.cfg [build_ext]` fold-back, asserted against the actual generated config file. A new `sdist_build` snapshot case pins the `uv.override_package(cc_deps = ...)` label threading onto the generated `pep517_native_whl(...)` call; the three existing sdist snapshots are byte-identical.

## Open design questions

Raising these here rather than deciding them unilaterally:

1. **Attribute surface and naming.** This PR uses a dedicated cc_deps = attr.label_list(providers = [[CcInfo]]) on the tag class and rule. Two alternatives were seriously considered. A generic deps with provider introspection (the rules_rust shape, where a cc_library goes straight into deps) reads more naturally but loses Bazel's analysis-time missing-provider errors, and deps would be a false cognate here. The rule's primary dependency axis is the lockfile, and the tag already has extra_deps meaning runtime Python deps, so an unqualified deps next to it invites the wrong guess. Harvesting CcInfo through the existing toolchains bus adds zero API surface but conflates tool selection with link inputs, cannot carry a provider constraint (it is Bazel's built-in attribute), and silently changes behavior for any existing user whose toolchain target happens to carry CcInfo. The dedicated attribute is opt-in by construction, self-describing next to extra_deps, and can be widened later without breaking anyone; the name follows rules_go's cdeps precedent for exactly this pattern. Naming is cheap to change now and breaking to change later, so if you prefer a different name or shape, let's decide it now.
2. **Allowlist growth.** Is extend-on-request through issues the right process for the link-flag allowlist? `-Wl,--exclude-libs` is the entry I would expect to be requested first.
3. **Backend generality.** The params file is deliberately mechanism-agnostic so a meson-python or scikit-build adapter can consume the same slots without analysis changes. Should the slot schema be treated as the stable seam for it?
4. **Whole-archive and shared deps.** `alwayslink` and dynamic-only deps are rejected. Whole-archive needs a channel that can express the wrapping dialect; shared deps additionally need rpath and wheel-staging decisions.
5. **macOS coverage.** Framework search paths (`-F`) are emitted and analysis-tested, but behavioral coverage needs a macOS lane; with #1372 enabling macOS runners, a follow-up behavioral case may now be feasible. Relatedly, is the `-framework` rejection acceptable?
6. **Pre-existing: the e2e sdist runtime tests never run.** While validating the snapshot case I found the generated hub alias is `target_compatible_with` incompatible in the default config (nothing activates the dep_group setting), so `//uv-sdist-native-build:test` is silently skipped everywhere including CI as far as I can tell, and forcing the group fails on missing `:build`/`:setuptools` aliases (the lock's build deps are orphan entries outside `dep_to_scc`). Inherited from main, zero delta from this PR; it bounds this PR's snapshot proof to BUILD-text emission (the executable proof is in-repo).
7. **Snapshot-case lock.** The new e2e case duplicates python-geohash's `uv.lock` under a renamed project to avoid new network deps; if you would rather share or generate these locks, happy to adjust.

- Searched for relevant documentation and updated as needed: yes (`docs/uv-patching.md` gains a "Linking native C/C++ dependencies" section and the constraints list; attr docs on the tag, repo rule, and build rule)

> `uv.override_package(cc_deps = [...])` wires Bazel `cc_library` (or any `CcInfo`) targets directly into a native sdist build: the dependency's transitive headers, include paths, and defines are added to the compile and its static archives are linked into the extension, with closure and link order taken from `CcInfo`. This is the declarative alternative to hand-writing `CPPFLAGS`/`LDFLAGS` through `env`. Setuptools-backed sdists and static archives only; see "Linking native C/C++ dependencies" in `docs/uv-patching.md`.

## Testing

- `bazel test //uv/private/pep517_whl/...` (23 tests: analysis contract, allowlist accept/reject, runtime gate matrix, and the executable build-import-call regression)
- `bazel test //:snapshots_15_test //:snapshots_16_test //:snapshots_17_test //:snapshots_18_test` from `e2e/cases` (the new cc_deps snapshot plus byte-stability of the three existing sdist_build snapshots)


xangcastle:
hi @sallustfire thanks for this PR, and using the main HEAD, we are working in the cross builds with differents backend in a single structured solution, please wait a bit on our implementation

PR #1389, opened 2026-07-29
fix: set tempdir vars to a shorter symlink to TEST_TMPDIR to prevent long path issues

### LINKED ISSUE #1387: [Bug]: pytest set tmp dir to TEST_TMPDIR unfriendly to multiprocess SyncManager
### What happened?

commit https://github.com/aspect-build/rules_py/commit/0c83a3e12bcbe62103fd52a0cdae62954617d6b5 set all temp dir to TEST_TMPDIR,  usually a very long path, and python 

```python
multiprocessing.get_context("spawn").Manager()
```

will create AF_UNIX socket under temp dir, which has max 108 length limit https://github.com/python/cpython/issues/93852, this leads error

```
----------------------------- Captured stderr call -----------------------------
Process SyncManager-1:
  File "/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/5ebdfe55fbe7b973dbe986e774f3671e/external
/aspect_rules_py++python_interpreters+python_3_12_x86_64_unknown_linux_gnu/lib/python3.12/multi
processing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/5ebdfe55fbe7b973dbe986e774f3671e/external
/aspect_rules_py++python_interpreters+python_3_12_x86_64_unknown_linux_gnu/lib/python3.12/multi
processing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/5ebdfe55fbe7b973dbe986e774f3671e/external
/aspect_rules_py++python_interpreters+python_3_12_x86_64_unknown_linux_gnu/lib/python3.12/multi
processing/managers.py", line 591, in _run_server
    server = cls._Server(registry, address, authkey, serializer)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/5ebdfe55fbe7b973dbe986e774f3671e/external
/aspect_rules_py++python_interpreters+python_3_12_x86_64_unknown_linux_gnu/lib/python3.12/multi
processing/managers.py", line 156, in __init__
    self.listener = Listener(address=address, backlog=128)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/5ebdfe55fbe7b973dbe986e774f3671e/external
/aspect_rules_py++python_interpreters+python_3_12_x86_64_unknown_linux_gnu/lib/python3.12/multi
processing/connection.py", line 464, in __init__
    self._listener = SocketListener(address, family, backlog)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xinyuwang/.cache/bazel/_bazel_xinyuwang/5ebdfe55fbe7b973dbe986e774f3671e/external
/aspect_rules_py++python_interpreters+python_3_12_x86_64_unknown_linux_gnu/lib/python3.12/multi
processing/connection.py", line 608, in __init__
    self._socket.bind(address)
OSError: AF_UNIX path too long
```

### Version

Output of `bazel --version`: 8.5.1

3ccde61ffbcb61f57f85484f91785007cf64775b

Language(s) and/or frameworks involved: python

### How to reproduce

minimal pytest 

```python
py_pytest_test(
    name = "a",
    srcs = [
        "a.py",
    ],
    deps = [
    "@pip//pytest"],
)
```

a.py

```python
import multiprocessing


def test_a():
    mgr = multiprocessing.get_context("spawn").Manager()
```

### Any other information?

_No response_

jbedard:
Fix #1387 

### Test plan

- New test cases added

### DIFF
--- a/py/private/launcher_env/aspect_rules_py_launcher_env.py
+++ b/py/private/launcher_env/aspect_rules_py_launcher_env.py
@@ -5,7 +5,10 @@ Both pytest_main.py and unittest_main.py import this before anything else
 resolves the environment they set up, so the two drivers cannot drift apart.
 """
 
+import atexit
+import hashlib
 import os
+import stat
 from pathlib import Path
 from typing import TYPE_CHECKING, Dict, Optional, Tuple
 
@@ -15,6 +18,70 @@ if TYPE_CHECKING:
 # coveragepy resolves manifest entries through symlinks; Bazel wants the
 # original spelling back in the LCOV (coveragepy#963).
 _absfile_mapping: Dict[str, str] = {}
+
+
+def _alias_dir() -> Optional[str]:
+    """A short directory to hold temp dir aliases, or None if this host won't
+    give us one only we can write to.
+
+    Traversable by others (0711) so a test that drops privileges can still
+    reach its own TMPDIR; planting an alias needs write, which they lack.
+    """
+    path = "/tmp/rpy-%d" % os.getuid()
+    try:
+        os.mkdir(path, 0o711)
+        return path
+    except FileExistsError:
+        pass
+    except OSError:
+        return None
+    # Only reuse an existing entry nobody else can write to; otherwise the
+    # aliases inside it are attacker-controlled.
+    try:
+        st = os.lstat(path)
+    except OSError:
+        return None
+    if not stat.S_ISDIR(st.st_mode) or st.st_uid != os.getuid():
+        return None
+    return path if not stat.S_IMODE(st.st_mode) & 0o022 else None
+
+
+def _unlink_alias(path: str, owner_pid: int) -> None:
+    # A fork inherits atexit callbacks, so a child exiting normally would take
+    # away the TMPDIR its parent is still using.
+    if os.getpid() != owner_pid:
+        return
+    try:
+        os.unlink(path)
+    except OSError:
+        pass
+
+
+def _short_tmpdir(real: str) -> str:
+    """A short alias for `real`, or `real` itself if none can be made."""
+    if os.name != "posix":
+        return real
+    directory = _alias_dir()
+    if directory is None:
+        return real
+    # Named for the target, not the invocation: a run killed before its exit
+    # hook leaves one entry, which its next run reuses instead of adding another.
+    # fsencode, not encode: a path holding non-UTF-8 bytes reaches us as
+    # surrogate escapes, which UTF-8 refuses to encode.
+    link = os.path.join(directory, hashlib.sha256(os.fsencode(real)).hexdigest()[:16])
+    try:
+        os.symlink(real, link)
+    except FileExistsError:
+        # Left by another run of this target; reuse without taking ownership.
+        # A still-live creator exiting first takes the alias with it.
+        try:
+            return link if os.readlink(link) == real else real
+        except OSError:
+            return real
+    except OSError:
+        return real
+    atexit.register(_unlink_alias, link, os.getpid())
+    return link
 
 
 def set_test_tmpdir() -> None:
@@ -29,11 +96,19 @@ def set_test_tmpdir() -> None:
     `noexec` — a test that writes an executable helper into pytest's `tmp_path`
     and runs it gets EACCES. TMP/TEMP are set too so `tempfile` resolves
     consistently on Windows.
+
+    TEST_TMPDIR is itself too deep to hold an AF_UNIX socket, whose address is
+    capped at 104 bytes of sun_path (issues/1387) — `multiprocessing`'s spawn
+    Manager binds one and dies with "AF_UNIX path too long". bind() measures the
+    string it is handed, not the resolved path, so TMPDIR gets a short symlink
+    and the files still land in TEST_TMPDIR.
     """
     if "TEST_TMPDIR" not in os.environ:
         return
+
+    tmpdir = _short_tmpdir(os.path.abspath(os.environ["TEST_TMPDIR"]))
     for var in ("TMPDIR", "TMP", "TEMP"):
-        os.environ[var] = os.environ["TEST_TMPDIR"]
+        os.environ[var] = tmpdir
 
 
 def shard_info() -> Optional[Tuple[int, int]]:
--- a/py/tests/tmpdir-af-unix/af_unix_probe.py
+++ b/py/tests/tmpdir-af-unix/af_unix_probe.py
@@ -1,0 +1,184 @@
+"""Shared assertions for https://github.com/aspect-build/rules_py/issues/1387.
+
+The generated pytest/unittest launchers point TMPDIR at TEST_TMPDIR, a deeply
+nested path under the bazel output base. An AF_UNIX socket created there
+overflows sun_path (108 bytes on Linux, 104 on macOS), so
+`multiprocessing.get_context("spawn").Manager()` -- which binds one via
+`connection.arbitrary_address()` -- dies with "OSError: AF_UNIX path too long".
+"""
+
+import atexit
+import hashlib
+import multiprocessing
+import os
+import socket
+import stat
+import tempfile
+
+import aspect_rules_py_launcher_env as launcher_env
+
+# multiprocessing.connection.arbitrary_address() builds
+# <tmp>/pymp-XXXXXXXX/listener-XXXXXXXX
+SOCKET_OVERHEAD = len("/pymp-abcd1234/listener-abcd1234")
+
+# The smaller of the two sun_path limits, so the assertion means the same thing
+# on every platform.
+SUN_PATH_MAX = 104
+
+# Mirrors the launcher's own guard: it only shortens TMPDIR on POSIX, so the
+# sun_path budget and the alias scheme are only asserted there.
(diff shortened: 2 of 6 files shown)


jbedard:
@tamird are you using `py_pytest_test` or `py_unittest_test`? Any comments/concerns with TMP vars being a synlink?

py/tests/tmpdir-af-unix/af_unix_probe.py
```diff
@@ -0,0 +1,45 @@
...
+SOCKET_OVERHEAD = len("/pymp-abcd1234/listener-abcd1234")
+
+# The smaller of the two sun_path limits, so the assertion means the same thing
+# on every platform.
+SUN_PATH_MAX = 104
+
+
+def check_tmpdir_leaves_room_for_a_socket() -> None:
+    tmp = tempfile.gettempdir()
+    assert len(tmp) + SOCKET_OVERHEAD < SUN_PATH_MAX, f"temp dir too deep for AF_UNIX: {tmp}"
```
tamird, line 36: This probe runs unconditionally in the pytest, unittest, and raw targets, but both launchers deliberately keep the original path on non-POSIX systems. A normal deep Windows test directory will therefore fail a Unix-socket length assertion even though that limit does not apply. Skip this assertion on non-POSIX platforms alongside the socket cases, or constrain these targets accordingly; keep the cross-platform containment check.

[tamirdex]

py/private/pytest_main.py
```diff
@@ -31,9 +32,39 @@
...
+def _short_tmpdir(real: str) -> str:
+    """A short alias for `real`, or `real` itself if none can be made."""
+    if os.name != "posix":
+        return real
+    link = "/tmp/rpy-" + os.urandom(8).hex()
+    try:
+        os.symlink(real, link)
+    except OSError:
+        return real
+    atexit.register(_unlink_quietly, link)
```
tamird: `atexit` never runs on test timeouts, `SIGKILL`, `os._exit()`, or native crashes. Because every invocation adds a fresh randomized entry to host-global `/tmp`, affected workers permanently accumulate dangling `rpy-*` aliases after Bazel cleans `TEST_TMPDIR`. Give aliases a deterministic, securely reclaimable or lifecycle-managed owner and exercise abnormal termination. Apply the same correction to `unittest_main.py` rather than leaving the two launchers with divergent cleanup.

[tamirdex]

tamird (changes requested):
Reviewing at the maintainer’s explicit request against cdbe6185dfd19844f898549c9aad9f6483c09be3. We do use `py_pytest_test`, so this changes real downstream test launchers. A short POSIX symlink is a sound way to preserve `TEST_TMPDIR` containment while fitting `AF_UNIX`, but the unconditional Windows probe and unreclaimable host-global aliases are blocking. The same lifetime fix must apply to both launchers.

[tamirdex]

tamird:
@jbedard Yes: we use `py_pytest_test`; I found no `py_unittest_test` consumers. A short POSIX symlink is reasonable: `AF_UNIX` checks the socket’s lexical path while the files still resolve into `TEST_TMPDIR`. The remaining concerns are the unconditional Windows assertion and aliases surviving abnormal exits; I’ve covered both [in the review](https://github.com/aspect-build/rules_py/pull/1389#pullrequestreview-4812819907).

[tamirdex]

jbedard:
@codex

py/private/launcher_env/aspect_rules_py_launcher_env.py
```diff
@@ -17,6 +20,60 @@
...
+def _short_tmpdir(real: str) -> str:
+    """A short alias for `real`, or `real` itself if none can be made."""
+    if os.name != "posix":
+        return real
+    directory = _alias_dir()
+    if directory is None:
+        return real
+    # Named for the target, not the invocation: a run killed before its exit
+    # hook leaves one entry, which its next run reuses instead of adding another.
+    link = os.path.join(directory, hashlib.sha256(real.encode()).hexdigest()[:16])
```
tamird: A POSIX path may contain arbitrary non-NUL bytes; Python represents invalid UTF-8 bytes in `TEST_TMPDIR` with surrogate escapes. `real.encode()` therefore raises `UnicodeEncodeError` and aborts both launchers before test startup. Hash `os.fsencode(real)` instead, update `check_tmpdir_alias_is_stable` to use the same encoding, and exercise a surrogate-escaped path. For example, `os.fsdecode(b"/tmp/rpy-\xff").encode()` fails, while `os.fsencode(os.fsdecode(b"/tmp/rpy-\xff"))` correctly returns the original bytes.

[tamirdex]

tamird (changes requested):
Re-reviewed f8c956a1621751b19bce5016b2069c3c84c69e82 at the explicit downstream-review request. The POSIX-only test guards, shared launcher helper, and per-user deterministic alias address both previous findings. One blocking correctness issue remains: `TEST_TMPDIR` may contain valid non-UTF-8 POSIX filename bytes, and hashing `real.encode()` raises `UnicodeEncodeError` before either launcher can run. Use `os.fsencode(real)` and cover a surrogate-escaped path in the regression; update the alias probe to derive its expected digest the same way.

[tamirdex]

xangcastle (approved):

jbedard:
@codex

py/private/launcher_env/aspect_rules_py_launcher_env.py
```diff
@@ -17,6 +20,62 @@
...
+    except FileExistsError:
+        # Left by an earlier run of this target; reuse without taking ownership,
+        # so a concurrent user of the same alias keeps it.
+        try:
+            return link if os.readlink(link) == real else real
+        except OSError:
+            return real
+    except OSError:
+        return real
+    atexit.register(_unlink_quietly, link)
```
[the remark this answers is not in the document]
tamird: Confirmed against eb330e7b23829fb52f03a9b0656d2b47486708b2 with a real fork: after the child exits normally, alias_exists_after_child_exit=False and parent_tmpdir_available=False. The parent therefore retains a dead TMPDIR. Capture the creating PID in the cleanup callback and cover fork/normal child termination.

[tamirdex]

tamird (changes requested):
Re-reviewed eb330e7b23829fb52f03a9b0656d2b47486708b2 at the maintainer’s request. The platform guards and filesystem-byte hashing fix the prior findings. One blocking lifecycle defect remains: a normally exiting forked child inherits the creator’s atexit callback and deletes its parent’s live TMPDIR alias. I reproduced this directly: after the child exits, both alias_exists_after_child_exit and parent_tmpdir_available are false. Restrict cleanup to the creating PID and add a fork/normal-exit regression; account for concurrent processes reusing the same alias as well.

[tamirdex]

py/tests/tmpdir-af-unix/af_unix_probe.py
```diff
@@ -0,0 +1,112 @@
...
+def check_reused_alias_is_left_for_its_owner() -> None:
+    """Only the creator registers cleanup, so a second process sharing the
+    alias never unlinks it out from under the first."""
+    real = os.path.join(os.environ["TEST_TMPDIR"], "reuse-1387")
+    os.makedirs(real, exist_ok=True)
+    alias = launcher_env._short_tmpdir(real)
+
+    before = atexit._ncallbacks()
+    assert launcher_env._short_tmpdir(real) == alias
+    assert atexit._ncallbacks() == before, "reuse registered a second cleanup"
```
tamird: This test says it covers a second process, but both `_short_tmpdir` calls execute in the same process and only assert the callback count. The untested reverse lifetime still fails: after a real second process reuses the alias, the creator’s normal cleanup removes it while that process is alive (`reuser_alias_survives_owner_cleanup=False` in a two-process reproduction). Either exercise both processes and define the intended shared-alias lifetime, or narrow the test/docstring to the same-process callback behavior actually covered.

[tamirdex]

tamird (commented):
Re-reviewed 936e39607882bb7e83b7292e13aa5401c5322468 at the maintainer’s request. The creator-PID guard fixes the forked-child defect: I repeated the real fork reproduction and the parent’s alias remains available. POSIX guards and filesystem-byte hashing also address the earlier findings. One ordinary correctness/test gap remains: the purported second-process reuse test never starts a second process, and an independently verified live reuser still loses its alias when the creator exits. This is non-blocking feedback rather than a request for changes.

[tamirdex]

tamird (commented):
The POSIX guards, filesystem-byte hashing, deterministic aliases, and creator-PID cleanup address the previous blockers. One nonblocking lifecycle gap remains: if independent processes reuse an alias and its creator exits first, cleanup removes the live reuser's TMPDIR; the current fork regression covers only the opposite exit order. The existing inline already documents this case.

[tamirdex]

PR #1390, opened 2026-07-29
refactor: extract shared test launcher env setup

jbedard:
pytest_main.py and unittest_main.py each carried their own copy of the bazel setup for tempdir, coverage etc, so a change to one silently diverged from the other. Move it to `//py/private/launcher_env`, imported by both.

No functional change.

### Test plan


jbedard:
@codex

xangcastle (approved):

PR #1392, opened 2026-07-30
feat: add flag to emit legacy rules_python providers

### LINKED ISSUE #1391: Public API for @rules_python PyInfo interop (reverse direction: native py_* consuming rules_py targets)
### Problem

`py/private/py_info_interop.bzl` documents that rules_py never emits `@rules_python`'s `PyInfo`:

> rules_py never *emits* `@rules_python`'s provider — the reverse direction (a `@rules_python` rule consuming a rules_py target) is not supported.

So a native `@rules_python` `py_binary`/`py_library`/`py_test` cannot depend on a rules_py target. This blocks incremental adoption in repos with a **mixed** provider tree — e.g. a shared library built with rules_py that is consumed by a downstream project still on `@rules_python`.

Our concrete case: we publish a bzlmod module built with rules_py that is consumed by a separate repo whose Python rules are `@rules_python`- and conda-based. To make its native `py_*` targets able to depend on our rules_py libraries, we currently work around the gap by:

1. reaching into private `py_library_utils` (`//py/private:py_library.bzl`) to build a custom `py_library` that emits **both** rules_py's and `@rules_python`'s `PyInfo`; and
2. reaching into private `get_py_info` / `has_py_info` (`//py/private:py_info_interop.bzl`) so provider-forwarding wrapper rules forward whichever `PyInfo` flavor a wrapped target carries.

Both loads sit behind `# buildifier: disable=bzl-visibility`, and private API has no stability contract, so an internals refactor can break us silently.

### Requests

1. **A supported way for rules_py library targets to also provide `@rules_python`'s `PyInfo`** (e.g. an opt-in attribute, or a documented wrapper), so `@rules_python` consumers work unchanged — covering the reverse direction the interop doc currently calls unsupported.
2. **Alongside/failing that, promote the read-side accessors** `has_py_info`, `get_py_info`, and the `RulesPythonPyInfo` re-export from `//py/private:py_info_interop.bzl` to a public load path (e.g. `//py:defs.bzl`, which already publicly re-exports rules_py's `PyInfo`), so provider-forwarding rules can read either flavor without a private load.

Either removes a class of `bzl-visibility` workarounds for mixed-tree consumers. Happy to contribute the PR if you're open to a direction.

jbedard:
Close #1391

### Test plan

- New test cases added

### DIFF
--- a/.github/workflows/ci-workflows.yaml
+++ b/.github/workflows/ci-workflows.yaml
@@ -81,6 +81,7 @@ jobs:
                 - { path: "e2e/interpreter-input-validation", slug: "e2e-interpreter-input-validation", runner: "ubuntu-latest" }
                 - { path: "e2e/interpreter-build-config", slug: "e2e-interpreter-build-config", runner: "ubuntu-latest" }
                 - { path: "e2e/rules-python-interop", slug: "e2e-rules-python-interop", runner: "ubuntu-latest" }
+                - { path: "e2e/rules-python-provider-compat", slug: "e2e-rules-python-provider-compat", runner: "ubuntu-latest" }
                 - { path: "e2e/rules-proto-grpc-python", slug: "e2e-rules-proto-grpc-python", runner: "ubuntu-22.04-32core" }
                 - { path: "examples/debugger", slug: "examples-debugger", runner: "ubuntu-latest" }
                 - { path: "examples/dev_deps", slug: "examples-dev_deps", runner: "ubuntu-latest" }
--- a/docs/migrating.md
+++ b/docs/migrating.md
@@ -29,6 +29,22 @@ bazel run //path/to/package:target.venv_link
 See [IDE Integration](/README.md#ide-integration) for target declarations and
 editor configuration.
 
+## rules_python provider compatibility layer
+
+Mid-migration, a `@rules_python` target depending on an already-converted
+rules_py library fails analysis with `does not have mandatory providers:
+'PyInfo'`. To keep it building:
+
+```
+# .bazelrc
+common --@aspect_rules_py//py:emit_rules_python_providers
+```
+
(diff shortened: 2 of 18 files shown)


xangcastle (approved):

ctcjab:
_(Claude here, commenting on behalf of @ctcjab.)_

Thanks for turning this around so fast — the flag is exactly what we need, and the `interop-rules-python-consumes-rules-py` e2e covers our case precisely: a downstream repo mid-migration consuming our rules_py `py_library` targets through the published bzlmod module, direct and transitive. We'll adopt `--@aspect_rules_py//py:emit_rules_python_providers` and delete the custom `py_library` wrapper + private `py_library_utils` reach-in we'd been carrying.

On **"would you only need this for `py_library`?"** — for us, yes:

- **`py_library`** — yes; that's the whole reverse edge (you `deps` on libraries).
- **`py_test`** — no; tests are leaves, nothing `deps` on them.
- **`py_binary`** — we don't need it. It's the one arguable gap for a *general* compat layer: a `py_binary` does provide `PyInfo` and Bazel permits it in `deps`, so a half-migrated tree *could* hit it — but it's uncommon and usually an anti-pattern (binaries are `data` deps, not `deps`). Same three-line shim if you want the layer complete/consistent; equally fine to keep the surface minimal. Your call.

One orthogonal note (the other half of #1391, not this PR's job): the **read** side. We forward whichever `PyInfo` flavor a *wrapped* target carries via `has_py_info`/`get_py_info` from `//py/private:py_info_interop.bzl`. We can reimplement that inline against the public `PyInfo` symbols, so it's not a blocker — but since you're already in that file, promoting those two accessors to a public load path would be a nice complement for provider-forwarding rules.

Thanks again!

PR #1393, opened 2026-07-31
test: remove rules_python from main e2e test cases and isolate to e2e/rules-python-interop

jbedard:
### Test plan

- New test cases added


e2e/cases/cross-repo-610/case.MODULE.bazel
```diff
@@ -1,30 +0,0 @@
-"Bazel dependencies"
```
jbedard, line 1: Nothing imported this, it has a sibling `setup.MODULE.bazel` file that is imported 🤷

e2e/cases/MODULE.bazel
```diff
@@ -6,7 +6,6 @@ bazel_dep(name = "bazel_skylib", version = "1.4.2")
 bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "rules_cc", version = "0.2.16")
 bazel_dep(name = "tar.bzl", version = "0.10.1")
-bazel_dep(name = "rules_python", version = "1.9.0")
```
jbedard, line 9: This is the main goal... the primary e2e test cases should be pure rules_py tests and not (directly) involve rules_python

e2e/rules-python-interop/MODULE.bazel
```diff
@@ -1,36 +1,77 @@
-"""Interop: rules_py rules share the interpreter provisioned by rules_python.
+"""Interop: rules_py and rules_python rules share one interpreter, either way round.
 
-This module deliberately provisions its only interpreter through rules_python's
-`python.toolchain()` — no `python_interpreters` — so rules_py and rules_python
-targets must both resolve, and run on, that single runtime.
+Both rulesets provision interpreters, split by version so neither shadows the
```
acozzette, line 3: I'm a little bit confused by this because both rules_py and rules_python appear to provision 3.12. How does that work?

acozzette:
**`e2e/rules-python-interop/test.sh:52`** — The version=3.12 sweep of the rules_python-flag-only check can pass even if rules_py's flag-fallback matching is completely broken, because rules_python's own default 3.12 toolchain independently satisfies resolution and reports the same version.

Verified experimentally: building with `--@aspect_rules_py//py:python_version=3.11 --@rules_python//python/config_settings:python_version=3.12` (a config where rules_py's hub has no toolchain matching either branch for 3.12) still resolves to rules_python's own 3.12 toolchain, and `report_version.sh` still prints `3.12`. A future regression that breaks the `_312_rpy_fallback` config_setting in `py/private/interpreter/repository.bzl` would go undetected at 3.12 (masked) while still failing at 3.9/3.10/3.13 — a coverage gap introduced by moving this check into a module where rules_python's default version coincides with one of the swept versions.

acozzette:
**`e2e/rules-python-interop/BUILD.bazel:216`** — `py_console_script_binary` `pycowsay` now sources from a two-version, flag-selected `@pip` hub instead of the old deterministic single-version `@rpy_pip` hub, but no test asserts which hub version actually backs the generated binary.

Unlike `pip_deps_test.py`, which explicitly asserts `six.__file__` contains `pip_311_six`, `console_script_test.sh` only checks stdout content (`moo-from-rules-py`), so a regression that silently resolves `pycowsay` to the wrong pip-hub version or Python ABI would go undetected.

acozzette:
**`e2e/rules-python-interop/report_version.bzl`** — This file is a near-verbatim duplicate of `report_exec_version.bzl` in the same directory, differing only in the toolchain-type label and the provider field read.

Both were added/touched in this same commit, so any future fix to the shared launcher-writing logic (error diagnostics, launcher script format, version-string formatting) must be hand-applied to both files. A small rule factory parameterized by toolchain type/provider field would avoid the duplication.

acozzette:
I posted a few comments from Claude that sound like things that could potentially be worth fixing. I have not dug into them carefully so it's entirely possible that Claude is wrong or that they're not worth the effort to address.

jbedard:
@codex

acozzette (approved):

PR #1397, opened 2026-08-05
fix(uv): pass the full compiler path as argv[0] in pep517 wrappers

sallustfire:
The pep517 compiler wrappers exec the resolved compiler with a bare basename as `argv[0]` (`build_helper.py`, wrapper template). Clang resolves its installation directory from `argv[0]` when `-no-canonical-prefixes` is in effect, which most hermetic toolchains set for reproducibility, and the pep517 build env pins `PATH`, so the basename resolves to nothing: the driver loses its `InstalledDir` and every driver-relative lookup breaks. The visible failure with a hermetic LLVM toolchain is `-fuse-ld=lld` in `LDFLAGS` rejected with `clang: error: invalid linker name in argument '-fuse-ld=lld'`, because `ld.lld` is searched next to the driver; resource dir inference and `-B` defaults break the same way.

Fix: exec with the full compiler path as `argv[0]`. Mode selection is unaffected since clang inspects the file name portion, which the full path preserves.

## Related hazard: legacy `tool_paths` shim directories (documented, not fixed)

This surfaced a second, adjacent hazard worth recording. For toolchains configured via legacy `tool_paths`, `cc_toolchain.compiler_executable` can resolve into a generated shim directory that contains none of the driver's sibling tools (no `ld.lld`), so driver-relative lookups fail there even with correct `argv[0]`. The mainline path has been immune since #1328 (`cc_common.get_tool_for_action` returns the true driver); `compiler_executable` survives only in the fallback branches for toolchains that expose no action configs, where it is the only signal available. This PR adds a comment at that fallback stating the hazard and pointing at the action-config path as the reliable source.

- Searched for relevant documentation and updated as needed: yes (comments at the wrapper template and the legacy fallback)

> Native sdist compiler wrappers now exec the compiler with its full path as `argv[0]`, so driver-relative lookups (`-fuse-ld=lld`, resource dirs, `-B` defaults) work with hermetic toolchains that set `-no-canonical-prefixes`.

### Test Plan

`wrapper_argv0_test` drives `build_helper` with a synthetic sdist and a fake driver that echoes its `argv[0]`. The driver is a `cc_binary` built by the configured toolchain and supplied through the test's `data` (no host compiler assumptions); it must be a real executable rather than a shebang script, because the kernel rewrites `argv[0]` to the script path for interpreter execution, which would make the test pass vacuously. The test copies the driver out of runfiles into a directory deliberately absent from the pinned `PATH`, so a basename `argv[0]` cannot be recovered by `PATH` lookup. Before the fix the driver observes the bare basename (`fakecc`); after, the full path.

- `bazel test //uv/private/pep517_whl/...` (11 tests, including the new `wrapper_argv0_test`; red before the fix with driver saw `argv[0]` `fakecc`, green after)


jbedard:
@codex

jbedard (approved):

PR #1402, opened 2026-08-06
fix: preserve files under a nested external/ directory in py_image_layer

### LINKED ISSUE #1398: [Bug]: py_image_layer corrupts regular files from rules_python pip hubs into malformed relative symlinks
### What happened?

Standalone-ish repro: https://github.com/jasonprado/rules-py-sympy-repro

The package SymPy has symlinks in its wheel:

```
  sympy/external/__init__.py -> external/__init__.py
  sympy/external/importtools.py -> external/importtools.py
```

In rules_py v2 these symlinks are corrupted. Forgive my LLM copy-paste but this is what seems to be happening:

> The image packer replaces real SymPy source files with relative symlinks such as sympy/external/__init__.py -> external/__init__.py. Because symlink targets are resolved relative to the link’s directory, that points to the nonexistent nested path sympy/external/external/__init__.py, so Python cannot read the package initializer. Python then treats sympy.external as an empty namespace package, causing imports like import_module to fail.

### Version

```
$ cat .bazelrc
common --@aspect_rules_py//py:python_version=3.12
common --@rules_python//python/config_settings:python_version=3.12
$ cat .bazelversion
8.7.0
```

It reproduces with both
- aspect_rules_py `v2.0.0-alpha.5`
- current upstream main at `6c45397935c13759e95a4ede933188469e444402`

### How to reproduce

```shell
Standalone-ish repro: https://github.com/jasonprado/rules-py-sympy-repro

Check it out and `./repro.sh`

It's ~150 LOC so you should be able to read it or create your own version.
```

### Any other information?

_No response_

jbedard:
Fix #1398

### Test plan

- New test cases added

### DIFF
--- a/e2e/cases/MODULE.bazel.lock
+++ b/e2e/cases/MODULE.bazel.lock
@@ -217,6 +217,22 @@
   },
   "selectedYankedVersions": {},
   "moduleExtensions": {
+    "//oci/py_image_layer:nested_external_repo.bzl%nested_external": {
+      "general": {
+        "bzlTransitiveDigest": "Pk1bq58WjcrJrtVupQC4OQBJgyIHh2kc0HxgVPvbyX0=",
+        "usagesDigest": "r1kk/efoyd5zfQf39m7JfHvbiH+60s2LJLfA6WW16pk=",
+        "recordedFileInputs": {},
+        "recordedDirentsInputs": {},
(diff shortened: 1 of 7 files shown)


thesayyn (approved):

jbedard:
@codex

PR #1405, opened 2026-08-08
perf(py_venv): collapse namespace entries to single-owner directories

jbedard:
A PEP 420 namespace entry descends to the first directory carrying an __init__.py, so a wheel shipping an __init__.py-free tree resolves every installed file to its own entry. Projecting those per file costs a declared output and a symlink action per file in every consuming venv. For example nvidia-cuda-cccl and nvidia-cuda-runtime together put 2108 symlinks and 156 directories into each venv that depends on them.

Collapse the resolved projection to its coarsest cover before it reaches venv assembly, mirroring _collapse_data_projection: bind the whole directory where one wheel ships everything below it, and descend only where wheels genuinely share one. The same two wheels now project 82 symlinks and 2 directories.

Ownership is counted over every claimant's raw ns_dirs — the wheel's implicit-namespace skeleton, precomputed per top-level claim by make_wheel_record — not over the resolved winners: a claimant routed to .pth still ships files under the directory, so it must block collapse above them or the PEP 420 __path__ union loses its contribution.

Candidacy is the inverse restriction: a directory binds only when it is a strict ancestor of an entry that survived resolution, and never when it contains a native projection or merge root, since those declare their own outputs there. A single-owner directory whose entries were all excluded or lost projects nothing under the per-entry set, so binding it would declare an output the resolution decided against — and below a merge root, one that nests inside the merge's own output, failing analysis. That is the azure-namespace-sibling shape: azure-core's regular azure/core becomes a physical merge, azure-identity's surviving sibling entry keeps collapse running, and the tracing wheel's single-owner azure/core/tracing must not bind inside the merge.

Purely a projection change: every skip and package_collisions report is still decided against the per-entry set.

Changes are visible to end-users: no

Test plan

- New unit cases in virtuals_resolvers_test.bzl: the nvidia shape collapses to four directory projections; a .pth-routed loser blocks collapse of the directory it shares; the azure-namespace-sibling shape (merge root plus a surviving sibling entry) binds nothing at or below the merge root.

- New e2e case venv-namespace-collapse with both real nvidia wheels. The test runs inside the venv and asserts the projection shape — shared directories stay real, single-owner subtrees bind as one symlink each, headers still resolve through the bound directories — and caps the symlink count. Reverting the collapse takes the projection from 82 symlinks back to 2108.
- The existing azure-namespace-sibling e2e case covers the analysis-time output conflict the candidacy restriction prevents.


jbedard:
@codex

py/private/py_venv/virtuals_resolvers.bzl
```diff
@@ -34,6 +34,20 @@ def _shallowest(roots):
             out.append(root)
     return out
 
+def _coarsest_cover(kept, candidate_dirs, owners):
+    """Bind each minimal-cover root of *candidate_dirs* in place of the paths below it."""
```
acozzette: I think a more detailed docstring here would be good, because it's not immediately obvious to me what each parameter represents.

acozzette (approved):

py/private/py_venv/virtuals_resolvers.bzl
xangcastle: agree

xangcastle (approved):

Issue #479, opened 2024-12-12
[Bug]: The tests parent directory (`"."`) is added to the PYTHONPATH when using `py_pytest_main`


keunhong:
### What happened?

When using `py_pytest_main` the target's parent directory is added to the PYTHONPATH causing the same issues as in https://github.com/aspect-build/rules_py/issues/368.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 8.0.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
```
bazel_dep(name = "rules_python", version = "1.0.0")
bazel_dep(name = "aspect_rules_py", version = "1.0.0", dev_dependency = True)

python_version = "3.10"
python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(
    configure_coverage_tool = True,
    is_default = True,
    python_version = python_version,
)
use_repo(python, "python_3_10", "python_versions", "pythons_hub")
```

Language(s) and/or frameworks involved: python

### How to reproduce

```shell
py_pytest_main(
    name = "__test__",
    deps = ["@pip//pytest"],
)

py_test(
    name = "my_test",
    srcs = ["my_test.py", "//:__test__"],
    main = "__test__.py",
    deps = [
        "//:__test__",
        "@pip//pytest",
    ],
)
```
If `my_test.py` has an import for e.g. `concurrent.futures` and there is a `concurrent.py` in the same directory it will cause a clash.
```


### Any other information?

_No response_

keunhong:
This seems to be because pytest adds the test directory to `sys.path` through the mechanism described here: https://docs.pytest.org/en/latest/explanation/pythonpath.html#pythonpath

Setting `--import-mode=append` fixes this.

alexeagle:
Cool, thanks for tracking down the culprit and a workaround. Do you think rules_py should include that flag for all users of py_pytest_main? Document its existence?

keunhong:
> Cool, thanks for tracking down the culprit and a workaround. Do you think rules_py should include that flag for all users of py_pytest_main? Document its existence?

I think `append` is a less surprising default, but I haven't considered all of the use cases. I think at least documenting it would be nice so no one has to go on a wild goose chase like I had to.

arrdem:
I suspect this has nothing to do with Pytest and everything to do with our implementation of `py_pytest_main`.

[...]

PR #1408, opened 2026-08-10
test(pytest): demonstrate consider_namespace_packages for Bazel packages

ctcjab:
A regression test and worked example for #479 and the `py_pytest_main` half of
#368. No rule changes — it uses `py_pytest_test`'s existing `env`.

## The bug

pytest names a test module by walking up from the file while each directory has
an `__init__.py`. Bazel packages routinely have none, so the walk stops early and
the module is imported — and registered in `sys.modules` — under a truncated name.

For `//mypkg/azure:file_transfer_test`, where `mypkg/` has no `__init__.py`:

```
module __name__ == "azure.file_transfer_test"      # not mypkg.azure....
sys.modules["azure"] is <the first-party mypkg/azure package>
```

The first-party package now shadows the installed `azure` distribution **for the
code under test**, so a product module doing `from azure.storage.blob import
BlobServiceClient` raises `ModuleNotFoundError: No module named 'azure.storage'`
— under Bazel only. The same code is correct under a pip or conda install. #368
predicted exactly this: *"if instead of `adder.py` your users create `numpy.py`,
`import numpy` will import this file instead of the `numpy` library."*

It hides well. The shadowing only breaks product code that reaches for the
shadowed distribution, which the affected tests tend to stub out. In our repo the
Azure client could not work in any Bazel-run process and the suite was green
throughout.

## Why the workarounds in #479 don't fix it

`--import-mode` changes where pytest *inserts* paths, not how it *names* modules.
Measured on rules_py 2.0.0-alpha.5 with pytest 9.0.3:

| pytest args | `sys.path[0]` | module `__name__` | SDK import |
|---|---|---|---|
| (default, prepend) | `<runfiles>/_main/mypkg` | `azure.file_transfer_test` | FAILS |
| `--import-mode=append` | stdlib zip | `azure.file_transfer_test` | FAILS |
| `--import-mode=importlib` | stdlib zip | `azure.file_transfer_test` | FAILS |
| `-o consider_namespace_packages=true` | (unchanged) | `mypkg.azure.file_transfer_test` | **OK** |

So `--import-mode=append` — the workaround this thread converged on — removes the
`sys.path` entry without fixing the shadowing, because the binding comes from the
module's *name*, not the path. Also ruled out: adding `__init__.py` to the parent
(it is not in the test target's runfiles, so pytest never sees it), and
`PYTHONSAFEPATH=1`.

## Why this is opt-in rather than the default

I tried it as a default first, and it fails on this repo's own e2e suite. The
generated name is derived from the directory path, and **a Bazel package name may
contain characters a Python identifier may not**:

- `//pytest-mock-530:test_mock_py_test` →
  `ValueError: invalid format: 'pytest-mock-530.test_mock'`. Collection tolerates
  the hyphenated name, but `pkgutil.resolve_name` rejects it, so `pytest-mock`'s
  `mocker.patch` fails. `--import-mode=importlib` does not help.
- `//unconstrained-dependencies:get_python_package_version_{main,humble}_test` →
  `ModuleNotFoundError`, because they import a sibling module by bare name, which
  only resolved while the package directory was an import root.

Hyphens are legal in Bazel package names and common in practice, so a repo-wide
default would break exactly that shape. Per-target opt-in lets a package whose
directories are valid Python identifiers get correct naming without imposing it
on packages that can't have it.

## What this does

`py_pytest_test` already exposes everything needed, so this adds no API:

```python
py_pytest_test(
    name = "test_namespace_package_module_naming",
    srcs = [...],
    env = {"PYTEST_ADDOPTS": "-o consider_namespace_packages=true"},
)
```

`pytest_args` works too, at the cost of a private per-test entrypoint via
`py_pytest_main`. The ini option arrived in **pytest 8.1**; on an older pytest,
pytest itself reports it — `PytestConfigWarning: Unknown config option`, and a
failed run under `--strict-config`.

An earlier revision of this PR added a `consider_namespace_packages` attribute
and then a version check behind it. Both are gone: @jbedard rightly pointed out
that the attribute wrapped a one-line `env` setting, and the version check only
reworded a message pytest already emits while suppressing the `--strict-config`
failure.

## Validation

- `//py/...` — **128/128 pass**.
- `pyright`, `ty`, and this repo's ruff config all clean on the changed files.
- The regression case is load-bearing: removing `consider_namespace_packages`
  from the target makes it fail.
- `e2e/cases //...` — 10 failures, and the **same 10 fail on an unmodified tree**
  in this environment (`oci/*`, `interpreter-features-*`). No regressions.
- New regression case in `e2e/cases/pytest-subdir-imports`, reusing that case's
  pytest hub. Verified load-bearing: removing the attribute from the target makes
  it fail.
- Downstream, in a 633-target Bazel repo: enabling this repo-wide left 4 failures,
  all of which were tests importing a sibling by bare name — i.e. relying on the
  bug. Correcting those to the full dotted path took the repo to **634/634**.

## One incidental change

`py/private/py_pytest_test.bzl` had no module docstring, so `buildifier-lint`
(which only lints changed files) blocked the commit. Added one line after the
existing comment block rather than restructuring it.

### DIFF
--- a/e2e/cases/pytest-subdir-imports/BUILD.bazel
+++ b/e2e/cases/pytest-subdir-imports/BUILD.bazel
@@ -42,3 +42,20 @@ py_pytest_test(
     env = {"PYTEST_ADDOPTS": "-o consider_namespace_packages=true"},
     deps = ["@pypi_pytest_subdir_imports//pytest"],
 )
+
+# Same case through `pytest_args` instead of `env`, proving both routes reach
+# pytest. This one bakes a private per-test entrypoint via py_pytest_main, where
+# the `env` target above stays on the shared prebuilt main.
+py_pytest_test(
+    name = "test_namespace_package_module_naming_via_args",
+    srcs = [
+        "namespace_pkg/sub/__init__.py",
+        "namespace_pkg/sub/module_name_test.py",
+    ],
+    dep_group = "pytest-subdir-imports",
+    pytest_args = [
+        "-o",
+        "consider_namespace_packages=true",
+    ],
+    deps = ["@pypi_pytest_subdir_imports//pytest"],
+)
--- a/e2e/cases/pytest-subdir-imports/BUILD.bazel
+++ b/e2e/cases/pytest-subdir-imports/BUILD.bazel
@@ -16,17 +16,29 @@ py_pytest_test(
     ],
 )
 
-# Regression for #479 / #368: `namespace_pkg/` has no `__init__.py`, so without
-# consider_namespace_packages pytest names this module `sub.module_name_test`
-# and binds `sub` as a top-level module -- shadowing any installed distribution
-# of that name for the code under test.
+# Regression and worked example for #479 / #368.
+#
+# `namespace_pkg/` has no `__init__.py`, which is the shape of an ordinary Bazel
+# package. pytest names a test module by walking up only through directories that
+# have one, so it stops there and imports this file as `sub.module_name_test` --
+# binding `sub` as a top-level module. A package whose directory name matches an
+# installed distribution then shadows that distribution for the code under test,
+# and `from <dist> import ...` fails under Bazel while the same code is correct
+# under pip.
+#
+# No new rule attribute is needed: `consider_namespace_packages` (pytest 8.1+)
+# makes pytest name the module by its full path, and `env` reaches it through
+# PYTEST_ADDOPTS. `pytest_args` would work too, at the cost of a private per-test
+# entrypoint. Directory names must be valid Python identifiers for this -- a test
+# under `my-pkg/` becomes `my-pkg.foo_test`, which collection tolerates but
+# `pkgutil.resolve_name` (and so pytest-mock's `mocker.patch`) rejects.
 py_pytest_test(
     name = "test_namespace_package_module_naming",
     srcs = [
         "namespace_pkg/sub/__init__.py",
         "namespace_pkg/sub/module_name_test.py",
     ],
-    consider_namespace_packages = True,
     dep_group = "pytest-subdir-imports",
+    env = {"PYTEST_ADDOPTS": "-o consider_namespace_packages=true"},
     deps = ["@pypi_pytest_subdir_imports//pytest"],
 )
--- a/e2e/cases/pytest-subdir-imports/BUILD.bazel
+++ b/e2e/cases/pytest-subdir-imports/BUILD.bazel
@@ -15,3 +15,18 @@ py_pytest_test(
         "@pypi_pytest_subdir_imports//pytest",
     ],
 )
+
+# Regression for #479 / #368: `namespace_pkg/` has no `__init__.py`, so without
+# consider_namespace_packages pytest names this module `sub.module_name_test`
+# and binds `sub` as a top-level module -- shadowing any installed distribution
+# of that name for the code under test.
+py_pytest_test(
+    name = "test_namespace_package_module_naming",
+    srcs = [
+        "namespace_pkg/sub/__init__.py",
+        "namespace_pkg/sub/module_name_test.py",
+    ],
+    consider_namespace_packages = True,
+    dep_group = "pytest-subdir-imports",
+    deps = ["@pypi_pytest_subdir_imports//pytest"],
+)
--- a/py/private/py_pytest_test.bzl
+++ b/py/private/py_pytest_test.bzl
@@ -122,7 +122,9 @@ def py_pytest_test(
             `my-pkg/` becomes `my-pkg.foo_test`, which collection tolerates but
             `pkgutil.resolve_name` (and so `pytest-mock`'s `mocker.patch`)
             rejects. Enable it for packages whose directories are valid Python
-            identifiers.
+            identifiers. Requires pytest 8.1, which added the ini option; on an
+            older pytest the driver leaves module naming alone and warns rather
+            than passing an unknown `-o` key.
         **kwargs: forwarded to the underlying test rule and sibling py_venv.
     """
     if "main" in kwargs:
@@ -131,12 +133,12 @@ def py_pytest_test(
     kwargs["testonly"] = True
 
     if consider_namespace_packages:
-        # Through PYTEST_ADDOPTS rather than pytest_args: the latter renders a
-        # private per-test entrypoint, and this needs no codegen. Any
-        # PYTEST_ADDOPTS the caller set is preserved.
+        # An env var the shared driver reads, rather than a baked pytest arg: only
+        # the driver can gate it on the pytest this target resolved (see
+        # pytest_main.py), and `pytest_args` would render a private per-test
+        # entrypoint where none is needed.
         env = dict(kwargs.pop("env", {}))
-        pytest_opt = "-o consider_namespace_packages=true"
-        env["PYTEST_ADDOPTS"] = (env["PYTEST_ADDOPTS"] + " " + pytest_opt) if "PYTEST_ADDOPTS" in env else pytest_opt
+        env["RULES_PY_CONSIDER_NAMESPACE_PACKAGES"] = "1"
         kwargs["env"] = env
 
     if resolutions:
--- a/py/private/py_pytest_test.bzl
+++ b/py/private/py_pytest_test.bzl
@@ -9,6 +9,39 @@
 #     `pytest_args`/`chdir` are requested,
 #   * pytest-dependency validation.
 
+"""Pytest-driving test macro and the driver wiring it shares with py_test.
+
+## Test module naming under Bazel packages
+
+pytest names a test module by walking up from the file while each directory has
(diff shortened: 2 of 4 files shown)


xangcastle (approved):

jbedard:
@codex

py/private/py_pytest_test.bzl
```diff
@@ -107,13 +110,35 @@ def py_pytest_test(
...
         fail("py_pytest_test provides its own entrypoint; `main` is not supported. Use py_pytest_main + py_test for a custom main.")
 
     kwargs["testonly"] = True
 
+    if consider_namespace_packages:
+        # Through PYTEST_ADDOPTS rather than pytest_args: the latter renders a
+        # private per-test entrypoint, and this needs no codegen. Any
+        # PYTEST_ADDOPTS the caller set is preserved.
+        env = dict(kwargs.pop("env", {}))
+        pytest_opt = "-o consider_namespace_packages=true"
```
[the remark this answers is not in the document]
jbedard: I guess that's another reason to put it behind a flag in the rule? Or is pytest <8.1 old enough we shouldn't even care?

ctcjab:
(Posted by Claude on my behalf)

Good catch, and it's a real gap — the guard existed in the first revision of this
PR and I lost it when I moved the flag out of the driver to make it opt-in. A
macro can't see the pytest version at analysis time, so the check had nowhere to
live. Fixed in 21bf45b: the attribute now sets an env var and the driver decides,
since only the driver sees the pytest a target actually resolved.

On "is pytest <8.1 old enough we shouldn't even care?" — I don't think so. 8.1 is
2024-03-03, so the 8.0.x window is only about five weeks, but the exposure isn't
8.0.x; it's the whole 7.x line, which is still widely pinned. And `py_pytest_test`
deliberately accepts whatever pytest a caller puts in `deps`, so rules_py doesn't
get to assume. The failure mode is the bad kind either way: a warning plus exactly
the truncated naming the caller opted out of, or a hard failure under
`--strict-config`.

Behavior now, verified both ways by forcing the threshold above the installed
pytest:

- pytest >= 8.1 — `-o consider_namespace_packages=true` is passed; the regression
  case passes.
- pytest < 8.1 — the option is *not* passed, stderr gets
  `WARNING: consider_namespace_packages requires pytest >= 8.1; this target
  resolved pytest X. Test modules keep their truncated names.`, and the
  regression case fails, which is the honest outcome: the naming really is
  truncated.

The 8.1 minimum is now on the attribute docstring too, which is where I should
have put it to begin with rather than only in the PR body.

`//py/...` is 128/128; the `e2e/cases` matrix is unchanged from an untouched tree
in my environment (the same 10 `oci/*` and `interpreter-features-*` failures with
or without this branch).

jbedard:
@ctcjab is that a gap? See my other comment - I don't know if it's the version sniffing logic just to get a different error message like this... 🤷

ctcjab:
(Posted by Claude on my behalf)

You're right, and checking what pytest actually does made that clear. On an
unknown ini key, pytest 9.0.3 gives:

```
PytestConfigWarning: Unknown config option: consider_namespace_packages
1 passed, 1 warning
```

and under `--strict-config` the run fails outright, no tests executed.

So it was never silent — Codex's "silently receive none of the advertised
protection" is what I took at face value, and I over-corrected on it. Worse, the
version check made things worse in the case that matters: by withholding the
option it also suppressed the `--strict-config` failure, which is exactly the
escalation someone turns `--strict-config` on to get. All it really bought was a
different wording of a message pytest already emits, at the cost of version logic
in the driver.

Dropped in 5718a1ec. `pytest_main.py` is back to untouched, so the diff is the
macro plus the e2e case again. The 8.1 minimum stays on the attribute docstring,
now describing what an older pytest does rather than trying to paper over it.

py/private/py_pytest_test.bzl
```diff
@@ -107,13 +110,37 @@ def py_pytest_test(
...
+        env["PYTEST_ADDOPTS"] = (env["PYTEST_ADDOPTS"] + " " + pytest_opt) if "PYTEST_ADDOPTS" in env else pytest_opt
```
jbedard: It seems like we're just modifying the `env`, or could potentially modify `args` instead... why not just do that yourself without adding a new public API?

I think this PR could change to just a test and demo of how to use `py_pytest_test(env|args)` in a useful way?

xangcastle: let me change it to and e2e demoing how to use the current API in for this cases

ctcjab:
(Posted by Claude on my behalf)

Agreed, and done in 2821b463 — the attribute is gone and the e2e case uses the
existing API:

```python
env = {"PYTEST_ADDOPTS": "-o consider_namespace_packages=true"},
```

`py/private/` is now byte-identical to main; the diff is three files, all under
`e2e/cases/pytest-subdir-imports`. Verified the case still fails without that
`env` line, so it is a real regression test and not just a demo. PR title and
description updated to match.

## One thing I would push back on, gently

Dropping the attribute is right — it wrapped a one-line `env` setting and that is
not worth permanent API. But the attribute was also carrying the *discoverability*,
and I do not think an `e2e/cases/` entry replaces that.

This failure is invisible from the symptom. What a user sees is:

```
ModuleNotFoundError: No module named 'azure.storage'
```

from code that works fine under pip. Nothing in that points at test module
naming. Nobody debugging it thinks "my test module's name was truncated at the
first directory without an `__init__.py`, so my own package is now bound as
top-level `azure` and is shadowing the SDK." I only found it because I was
auditing lazy imports for an unrelated lint rule, and the deferral that was
hiding it had been there long enough that everyone assumed it was load-bearing.

Anyone with a Bazel package named after a distribution they depend on is exposed,
and `azure`, `google`, `data`, `common` and `tools` are not unusual package
names. They will hit it as an inexplicable import error, not as something they
know to search for.

So: no API, but could this land somewhere a user would actually meet it — a note
in the pytest docs page, or in the `py_pytest_test` docstring, along the lines of
"Bazel packages have no `__init__.py`, so pytest may name test modules by a
truncated path; set `consider_namespace_packages` if a first-party package shares
a name with a distribution"? Happy to write that in this PR if you tell me where
you would want it, or to leave it and open it separately.

e2e/cases/pytest-subdir-imports/BUILD.bazel
```diff
@@ -15,3 +15,30 @@ py_pytest_test(
...
+# under `my-pkg/` becomes `my-pkg.foo_test`, which collection tolerates but
+# `pkgutil.resolve_name` (and so pytest-mock's `mocker.patch`) rejects.
+py_pytest_test(
+    name = "test_namespace_package_module_naming",
+    srcs = [
+        "namespace_pkg/sub/__init__.py",
+        "namespace_pkg/sub/module_name_test.py",
+    ],
+    dep_group = "pytest-subdir-imports",
+    env = {"PYTEST_ADDOPTS": "-o consider_namespace_packages=true"},
```
jbedard, line 42: Can we have another `py_pytest_test` that proves it works with `args` instead of `env`?

jbedard:
This should be pointed out in the main docs for the `py_pytest_test.bzl` file and `py_pytest_test` macro so they will ideally show in IDEs but also whatever docs website renders these

e2e/cases/pytest-subdir-imports/BUILD.bazel
ctcjab: (Posted by Claude on my behalf)

Added in 32325566 as `test_namespace_package_module_naming_via_args`, driving the
same case through `pytest_args = ["-o", "consider_namespace_packages=true"]`.
Both targets pass, and I checked the new one is load-bearing rather than
decorative — drop the `pytest_args` and it fails.

The BUILD comment notes the difference between the two routes: the `args` variant
bakes a private per-test entrypoint via `py_pytest_main`, where the `env` one
stays on the shared prebuilt main.

ctcjab:
(Posted by Claude on my behalf)

Done in 32325566, in both places you named.

`py_pytest_test.bzl` now carries a module docstring with a "Test module naming
under Bazel packages" section — the mechanism, a copy-pasteable snippet, and the
identifier constraint that makes it opt-in per target rather than a default. The
`py_pytest_test` docstring gets a short pointer to it, so it shows on hover where
a caller is actually writing the target.

Both lead with the symptom rather than the cause:

```
ModuleNotFoundError: No module named 'azure.storage'
```

from code that is correct under pip. That is the string someone will search for;
"module naming" is not, because they have no reason yet to suspect it.

You were right that this belongs in docs rather than in an attribute — stardoc
and IDE hover reach people who do not yet know the problem exists, which is the
population that needs it. An attribute only helps someone already looking for it.

Two side notes: the `args` variant you asked for is in the same commit (replied in
that thread), and the module docstring also clears buildifier's
`module-docstring` warning on this file, which it did not previously carry.

jbedard (approved):

xangcastle (approved):

PR #1410, opened 2026-08-10
fix: drop bytecode orphaned by a site-packages merge

jbedard:
A wheel can overlay another wheel's `mod.py` without supplying its `__pycache__/mod.*.pyc`, leaving the earlier wheel's bytecode beside the winning source. Bytecode now follows the same overlay order as every other merged file: a cache is dropped only when its wheel precedes the one that supplied the source.

### Test plan

- New test cases added


jbedard:
@codex

acozzette (approved):

PR #1413, opened 2026-08-11
test(e2e): port rules_pycross sdist-build cases

xangcastle:
Ports four cases from rules_pycross's e2e suite, exercising the **existing** sdist→wheel pipeline (host builds only — no cross-compilation involved). Split out of #1363, where these cases don't depend on the cross feature; every test passes against current `main` unchanged.

## Cases

- **pycross-setuptools** (`build_setuptools`): three setuptools C-extension sdists — PyYAML (Cython-generated extension, with an observable `PYYAML_FORCE_LIBYAML=0` env override), setproctitle (plain C, `pre_build_patches`), zstandard (vendored libzstd, `resource_set`).
- **pycross-patches** (`patches_and_hooks`): pre-build and post-install patch phases stacked on setproctitle — the post-install hunk carries the pre-build patch's output as context, so an ordering regression fails the build itself.
- **pycross-pure-python** (`build_pure_python`): hatchling and flit-core backends, runtime imports, site-packages placement hygiene, plus a `collect_wheels` matrix asserting anyarch wheels stay `-none-any` under non-host platform transitions (exec-platform resolution of the build tooling, no native toolchain needed).
- **pycross-distutils-probe**: build-action env hygiene — a fresh child interpreter spawned by the backend must resolve `distutils` on Python 3.12+, guarding `rule.bzl`'s `_INHERITED_PYTHON_ENV` filter. Fails only in the child, so it's invisible to a plain "does it build" check.

Shared `tools/`: `collect_wheels` macro (adds the wheel-tag assertion the rules_pycross original lacks) and `check_wheel_tags.py`.

## Not included

`pycross-setuptools`' cross matrix (`native_wheels*`, `check_wheel_native.py`): building a C extension for a non-host platform needs `native_build_toolchain_type` resolution for that platform, which lands with #1363. The second commit documents that scope cut.

---

### Test plan

- New test cases added


e2e/cases/tools/collect_wheels.bzl
```diff
@@ -0,0 +1,71 @@
...
+gathers the results into one directory. rules_pycross stops at "it builds";
+this version adds a runtime assertion over the collected wheel filenames so a
+wheel tagged for the wrong platform fails the test instead of passing silently.
+"""
+
+load("@aspect_rules_py//py:defs.bzl", "py_test")
+load("@bazel_lib//lib:run_binary.bzl", "run_binary")
+load("@bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
+
+def collect_wheels(name, wheels, platforms, expected_tags = None, **kwargs):
```
jbedard: I don't see this used anywhere?

jbedard (approved):

PR #1414, opened 2026-08-11
test(e2e): remove uv-deps-650's crossbuild sub-case

xangcastle:
Removes the `crossbuild` sub-case from `e2e/cases/uv-deps-650` (amd64/arm64 image-layer snapshots, pyc and exec-toolchain checks, ~4300 lines). It is superseded by the dedicated `e2e/crossbuild` workspace arriving with #1363, which covers the same ground per PEP 517 backend with real QEMU execution and per-arch structural checks.

Landing the removal ahead of #1363:
- keeps duplicate coverage out of the tree once that PR merges
- shrinks #1363's `e2e/cases` footprint to only feature-coupled edits
- trims the `pypi_uv_deps_650` hub back to three projects (the `uv-gazelle-778` manifest regenerates accordingly)

Pure deletion plus the two mechanical follow-ups — nothing on `main` references the removed targets (verified: `//:snapshots_*_test` points at `e2e/cases` root snapshots, and the only other "crossbuild" match is the unrelated `windows-crossbuild-837`). Full `e2e/cases` suite green locally under `--lockfile_mode=error`.


jbedard:
Can this not move to `e2e/crossbuild` now? I'd rather move now and modify later instead of just deleting it...

PR #1432, opened 2026-08-14
refactor: move unpack.py into a configurable toolchain

jbedard:
This way the toolchain can be customized in scenarios such as large repos where the default unpack.py is not performant enough.

### Test plan

- New test cases added


jbedard:
@codex

jbedard:
@tamird would this enable your custom wheel unpacker without the patch you have today? Review please

PR #1439, opened 2026-08-14
perf: avoid unnecessary unpack.py imports, lazy load

jbedard:
Python interpreter start is a fixed 12.2ms floor, so the script-attributable overhead went 39.5 → 24.6ms — 38% cut. Remaining 24.6ms is mostly zipfile (plus its bz2/lzma/zstd/shutil pulls), _hashlib, re, csv — not removable without vendoring an unzip.

- Saving is per `PyUnpackedWheel`/`WhlInstall` action, fixed regardless of wheel size. Small wheels (majority of a lockfile) it's ~30% of the action; for `numpy`-sized wheels extraction dominates and this change is only noise.
- top-500-wheel graph: ~7.5s CPU shaved; wall-clock gain smaller since actions run parallel, but these fire in bulk on cold builds / lockfile bumps.
- Numbers are host-python 3.14; hermetic exec toolchain will differ slightly in absolute ms (argparse's _colorize→inspect chain is 3.14-specific, older versions save a bit less there) — same shape though.

### Test plan

### DIFF
--- a/py/private/py_unpacked_wheel.bzl
+++ b/py/private/py_unpacked_wheel.bzl
@@ -19,8 +19,10 @@ def _py_unpacked_wheel_impl(ctx):
     args.add(unpack_script)
     args.add_all([unpack_directory], expand_directories = False, before_each = "--into")
     args.add("--wheel", ctx.file.src)
-    args.add("--python-version-major", py_toolchain.interpreter_version_info.major)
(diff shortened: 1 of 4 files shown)


xangcastle (approved):

xangcastle (approved):

jbedard:
@codex

PR #1441, opened 2026-08-15
refactor: output pep517 wheels as files, always pass a .whl file to unpack

jbedard:
This makes some of the rule inputs+outputs strictly `.whl` files instead of sometimes being directories. The tests need to adapt a bit, but this narrows the API requirements of `unpack.py` and allows removing one `.glob()` call.

Two contract clauses drop out for anyone writing a custom tool:

1. `--wheel` is one shape, not two. Before, the contract was "either a .whl file or a directory containing exactly one .whl" — every implementation had to stat the path, glob a directory, enforce the exactly-one rule, and pick error behaviour for zero/many. Now it's "open this zip." The C example on the toolchains-unpack branch silently didn't implement the directory case, meaning it was broken for every source-built wheel; that whole failure class is gone.
2. Filename parsing left the contract. The .data/ prefix fallback used to be derived from the basename — split on -, take two fields, plus percent-decoding (unquote) for escaped stems. A conforming implementation needed URL-decoding and PEP 427 filename-format knowledge. Now the prefix comes only from inside the archive (RECORD's dist-info dir, else the first top-level *.dist-info/ member) — data the implementation is already iterating over to extract. The basename is contractually meaningless, which is also what made the fixed whl.whl output name possible at all.

### Test plan


jbedard:
@codex

PR #1445, opened 2026-08-17
refactor!: make resolutions a plain string-to-label dict

jbedard:
Bazel 8's `attr.string_keyed_label_dict` lets the resolutions attribute take the natural `{"name": label}` mapping directly, so the label-keyed inversion and the macro-level helper struct are no longer needed.

The legacy `resolutions` attribute is now a plain `{"name": ":label"}` paramater.

Replace `resolutions.from_requirements(all_whl_requirements_by_package, requirement)` with `{pkg: requirement(pkg) for pkg in all_whl_requirements_by_package.keys()}` and `.override({...})` with dict union (`base | {...}`).

### Test plan


xangcastle (approved):

jbedard:
@codex

PR #1446, opened 2026-08-17
refactor!: require a PyInfo provider on resolution targets

jbedard:

All `requirements` attributes must now provide a valid python target and no longer accept generic `DefaultInfo` such as `filegroup`. Normally basic `DefaultInfo` outputting targets can be replaced or wrapped with `py_library`.

### Test plan


jbedard:
@codex

PR #1450, opened 2026-08-17
feat(uv): cross-compile native sdist builds

xangcastle:
Second slice of #1434 (after #1447): the cross-compilation core, with no e2e fixtures or CI changes — those follow in separate PRs.

When the exec and target platforms differ, `pep517_native_whl` now enters **cross mode** instead of failing toolchain resolution:

- **`exec_transition.bzl`** — pins the PEP 517 frontend to the host platform and resets the `platform_libc`/`platform_version` Starlark flags, so the frontend's Python build deps resolve for the exec interpreter instead of leaking the target configuration. `tool` handling is 1:2-safe via `tool_files_to_run`.
- **`cc_layer.bzl`** — extracts the target-configured C++ toolchain (compiler, binutils, sysroot, builtin include dirs) and forwards it to the build action.
- **`build_helper.py`** — derives the cross build environment from that layer: compiler wrappers that stay valid across the backend's chdir, generated meson cross files, CMake toolchain files, cargo target env for maturin/setuptools-rust, an ar-style argv translator over `llvm-libtool-darwin` for sandboxes that don't mount `llvm-ar`, sysconfig overrides for the target interpreter, and post-build validation that the produced wheel's platform tag names the target (not the exec host).
- **`build_helper_cross_test`** — exercises the helper's cross env derivation under the real pypa/build frontend (linux-only).

Native (exec == target) builds keep their previous code path and remain covered by the existing `pep517_whl` tests.

Cross builds of native sdists now work when a cross-capable C++ toolchain (e.g. `toolchains_llvm`) is registered; previously they failed toolchain resolution by design. Pure-Python sdists are unaffected. `docs/uv.md` gains a cross-compilation section.

### Test plan

- New `build_helper_cross_test` (linux-only) covers the cross env derivation
- `bazel test //uv/...`: 163 passed, 3 skipped (linux-only on a darwin host)
- End-to-end coverage (bcrypt, numpy, contourpy, pydantic_core, jpype1, … built for both linux archs and validated by ELF `e_machine` inspection) lands with the e2e/crossbuild suite PRs stacked on this one

### DIFF
--- a/uv/private/pep517_whl/common.bzl
+++ b/uv/private/pep517_whl/common.bzl
@@ -2,6 +2,7 @@
 
 load("@bazel_lib//lib:resource_sets.bzl", "resource_set_attr")
 load("//uv/private:source_built_wheel.bzl", "SourceBuiltWheelInfo")
+load(":exec_transition.bzl", "exec_transition")
 
 TARGET_EXEC_GROUP = "target"
 
@@ -16,6 +17,12 @@ def wheel_providers(wheel_file, console_scripts):
         DefaultInfo(files = depset([wheel_file])),
         SourceBuiltWheelInfo(console_scripts = tuple(console_scripts)),
     ]
+
+def tool_files_to_run(ctx):
+    tool = ctx.attr.tool
+    if type(tool) == "list":
+        tool = tool[0]
+    return tool[DefaultInfo].files_to_run
 
 def common_env(ctx):
     # pyproject_hooks copies the build process environment and launches its
@@ -65,10 +72,16 @@ _PATCH_ATTRS = {
 
 PEP517_WHL_ATTRS = {
     "src": attr.label(allow_single_file = True),
-    # The wheel action uses the named group below, so its frontend must use the
-    # same execution platform:
-    # https://bazel.build/extending/exec-groups#defining-exec-groups
-    "tool": attr.label(executable = True, cfg = config.exec(TARGET_EXEC_GROUP)),
+    # The wheel action runs in the "target" exec group; its frontend must run
+    # on the exec (host) platform but resolve its Python build dependencies
+    # for the *target* Python version. exec_transition pins --platforms to the
+    # host and resets the platform_libc/platform_version flags to host values
+    # (Starlark flags otherwise leak from the target configuration and break
+    # wheel selection for build deps). See exec_transition.bzl.
+    "tool": attr.label(executable = True, cfg = exec_transition),
+    "_allowlist_function_transition": attr.label(
+        default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+    ),
     "version": attr.string(),
     "console_scripts": attr.string_list(
         doc = "Console scripts discovered from the source distribution's entry-point metadata.",
--- a/uv/private/pep517_whl/pep517_native_whl.bzl
+++ b/uv/private/pep517_whl/pep517_native_whl.bzl
@@ -137,21 +137,24 @@ def _find_sysconfigdata(runtime):
             return f
     return None
 
+# Deployment-target floor for the macOS platform string when the real value
+# is unknown: 11.0 is the first macOS release with arm64 support, so it is
+# the most conservative version every hermetic arm64 interpreter satisfies.
+_MACOS_DEPLOYMENT_FALLBACK = "11.0"
+
 def _derive_python_host_platform(target_os, target_cpu):
     """Derive _PYTHON_HOST_PLATFORM from target platform constraints.
 
     Linux: libc does not affect the platform string — always linux-{cpu}.
-    macOS: uses arm64 (not aarch64) and requires a version component. The
-    version here is only an analysis-time fallback: the deployment target is
-    a property of the target interpreter, which analysis can't read, so
-    build_helper re-derives it at build time from the target sysconfigdata's
-    MACOSX_DEPLOYMENT_TARGET whenever that file is available.
+    macOS: uses arm64 (not aarch64) and requires a version component,
+    defaulting to _MACOS_DEPLOYMENT_FALLBACK when the interpreter's real
+    deployment target is not known.
     """
     if target_os == "linux":
         return "linux-" + _PYTHON_CPU_MAP.get(target_cpu, target_cpu)
     if target_os == "darwin":
         cpu = "arm64" if target_cpu == "aarch64" else target_cpu
-        return "macosx-11.0-" + cpu
+        return "macosx-{}-{}".format(_MACOS_DEPLOYMENT_FALLBACK, cpu)
     return None
 
 def _interpreter_platform_triple(runtime):
--- a/uv/private/pep517_whl/pep517_native_whl.bzl
+++ b/uv/private/pep517_whl/pep517_native_whl.bzl
@@ -258,14 +258,21 @@ def _pep517_native_whl(ctx):
             extra_inputs.append(cc_layer.static_runtime_files)
         if cc_layer.static_runtime_paths:
             env["RULES_PY_CXX_STATIC_RUNTIME"] = ":".join(cc_layer.static_runtime_paths)
-        if cc_layer.cflags:
-            env["CFLAGS"] = cc_layer.cflags
-        if cc_layer.cxxflags:
-            env["CXXFLAGS"] = cc_layer.cxxflags
-        if cc_layer.ldflags:
-            env["LDFLAGS"] = cc_layer.ldflags
-        if cc_layer.ldshared_flags:
-            env["LDSHAREDFLAGS"] = cc_layer.ldshared_flags
+        # Toolchain flags first, then any flags the package set via `env`
+        # (uv.override_package): the package's -D/-std/feature-baseline
+        # additions must survive, and trailing position lets them override.
+        # Values inherited from the ambient shell env stay excluded — only
+        # an explicit `env` entry merges.
+        for key, toolchain_flags in (
+            ("CFLAGS", cc_layer.cflags),
+            ("CXXFLAGS", cc_layer.cxxflags),
+            ("LDFLAGS", cc_layer.ldflags),
+            ("LDSHAREDFLAGS", cc_layer.ldshared_flags),
+        ):
+            if not toolchain_flags:
+                continue
+            package_flags = env.get(key) if key in ctx.attr.env else None
+            env[key] = toolchain_flags + " " + package_flags if package_flags else toolchain_flags
         if cc_layer.ccshared:
             env["CFLAGS"] = (env.get("CFLAGS", "") + " " + cc_layer.ccshared).strip()
             env["CXXFLAGS"] = (env.get("CXXFLAGS", "") + " " + cc_layer.ccshared).strip()
--- a/uv/private/pep517_whl/pep517_native_whl.bzl
+++ b/uv/private/pep517_whl/pep517_native_whl.bzl
@@ -1,13 +1,11 @@
-"""PEP 517 sdist to platform-specific whl build rule.
-
-Uses `python -m build` (the pypa/build frontend) which delegates to whatever
-build backend the sdist declares in its `[build-system]` table.
-"""
+"""PEP 517 sdist to platform-specific whl build rule."""
 
 load("@bazel_lib//lib:resource_sets.bzl", "resource_set")
 load("@rules_cc//cc:action_names.bzl", "ACTION_NAMES")
 load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
-load("//py/private/toolchain:types.bzl", "NATIVE_BUILD_TOOLCHAIN", "PY_TOOLCHAIN")
+load("//py/private/interpreter:versions.bzl", "PLATFORMS")
+load("//py/private/toolchain:types.bzl", "EXEC_TOOLS_TOOLCHAIN", "NATIVE_BUILD_TOOLCHAIN", "PY_TOOLCHAIN")
+load(":cc_layer.bzl", "CC_LAYER_ATTRS", "extract_cc_layer")
 load(
     ":common.bzl",
     "PEP517_WHL_ATTRS",
@@ -15,8 +13,10 @@ load(
     "common_env",
     "memory_args",
     "patch_args_and_inputs",
+    "tool_files_to_run",
     "wheel_providers",
 )
+load(":exec_transition.bzl", "exec_transition")
 
 _CC_TOOLCHAIN_TYPE = Label("@bazel_tools//tools/cpp:toolchain_type")
 _EXECROOT_MARKER = "__ASPECT_RULES_PY_EXECROOT__"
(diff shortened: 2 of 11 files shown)


xangcastle:
CI caught a missed adaptation: `e2e/cases/pep517-frontend-exec-group` asserted the frontend's own configuration matched the target arch. Under `exec_transition` the frontend is deliberately host-configured now, so that assertion is removed — the case keeps verifying that the **target** exec group's CC toolchain reaches the build (compiler/tool checks are untouched), with a NOTE documenting the semantics. Amended into the single commit.

jbedard:
@codex

uv/private/pep517_whl/pep517_whl.bzl
```diff
@@ -29,17 +30,18 @@ def _pep517_whl(ctx):
     # the action sandbox, which means the venv shim can find the interpreter
     # via the standard runfiles mechanism regardless of whether the interpreter
     # comes from an external repo or the main workspace.
+    tool = tool_files_to_run(ctx)
     ctx.actions.run(
         mnemonic = "PySdistBuild",
         progress_message = "Source compiling {} to a whl".format(archive.basename),
-        executable = ctx.executable.tool,
+        executable = tool,
```
jbedard, line 37: This changed from `ctx.executable.tool` to `ctx.attr.tool[DefaultInfo].files_to_run` - what does that do?

uv/private/pep517_whl/pep517_native_whl.bzl
```diff
@@ -116,11 +116,113 @@ def _cc_toolchain_inputs_and_tools(ctx):
...
+    version here is only an analysis-time fallback: the deployment target is
+    a property of the target interpreter, which analysis can't read, so
+    build_helper re-derives it at build time from the target sysconfigdata's
+    MACOSX_DEPLOYMENT_TARGET whenever that file is available.
+    """
+    if target_os == "linux":
+        return "linux-" + _PYTHON_CPU_MAP.get(target_cpu, target_cpu)
+    if target_os == "darwin":
+        cpu = "arm64" if target_cpu == "aarch64" else target_cpu
+        return "macosx-11.0-" + cpu
```
jbedard: There's an 11 hardcoded in there? 🤔

uv/private/pep517_whl/pep517_native_whl.bzl
```diff
@@ -116,11 +116,113 @@ def _cc_toolchain_inputs_and_tools(ctx):
...
+        if f.basename.startswith("_sysconfigdata") and f.basename.endswith(".py") and prefix in f.path:
+            return f
+    return None
+
+def _derive_python_host_platform(target_os, target_cpu):
+    """Derive _PYTHON_HOST_PLATFORM from target platform constraints.
+
+    Linux: libc does not affect the platform string — always linux-{cpu}.
+    macOS: uses arm64 (not aarch64) and requires a version component. The
+    version here is only an analysis-time fallback: the deployment target is
```
jbedard: Only an "analysis-time fallback"? How do you know? That implies you know everywhere this util is used and should just inline it instead. I think a util method shouldn't have such details about its use cases in the doc string?

uv/private/pep517_whl/pep517_native_whl.bzl
```diff
@@ -116,11 +116,113 @@ def _cc_toolchain_inputs_and_tools(ctx):
...
+    "x86": "i686",
+    "arm": "arm",
+}
+
+def _find_sysconfigdata(runtime):
+    """Locate _sysconfigdata*.py inside the target interpreter's file tree."""
+    if not runtime or not hasattr(runtime, "interpreter_version_info"):
+        return None
+    info = runtime.interpreter_version_info
+    prefix = "lib/python{}.{}".format(info.major, info.minor)
```
jbedard, line 134: What about `pythonM.mt`?

uv/private/pep517_whl/pep517_native_whl.bzl
```diff
@@ -116,11 +116,113 @@ def _cc_toolchain_inputs_and_tools(ctx):
...
+    for f in runtime.files.to_list():
```
jbedard, line 135: `.to_list()`? :(

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
+# import (borrowing symbols another loaded .so pulled in) yet fail at
+# runtime. GNU-ld-only syntax: skipped on Darwin, where clang++ links libc++
+# implicitly.
+_STATIC_LIBSTDCXX_FLAGS = ("-Wl,-Bstatic", "-lstdc++", "-Wl,-Bdynamic")
+
 _COMPILER_WRAPPER = """#!/usr/bin/env python3
 import os
 import sys
 
 filtered_args = [arg for arg in sys.argv[1:] if arg != "{debug_flag}"]
```
jbedard, line 70: We want to continue the python-within-python instead of creating separate files that tag args like `--cxx`? IDK what is best, just curious...

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
+is_cxx = {is_cxx!r}
+is_darwin = {is_darwin!r}
+# Same non-link detection as the cross wrapper: compiles, preprocessing,
```
jbedard, line 73: Whats "the cross wrapper"?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
+    if _arg in ("-c", "-E", "-S", "-fsyntax-only", "--version", "-dumpmachine", "-dumpversion", "-###") or _arg.startswith("-print-"):
+        is_link = False
+        break
+if is_cxx and is_link and not is_darwin:
+    filtered_args = filtered_args + {static_libstdcxx_flags!r}
 sysroot = {sysroot!r}
 if sysroot and "-isysroot" not in filtered_args:
     filtered_args = ["-isysroot", sysroot] + filtered_args
+
+# argv[0] must be the compiler's full path, not just its basename: GCC's
```
jbedard: Is this comment required? I see no `basename` logic here, so why are we talking about it?

I do know we had a bug recently fixed where we were using `basename`, did some AI-slop slip through (there and here?)?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
 import sys
 
 filtered_args = [arg for arg in sys.argv[1:] if arg != "{debug_flag}"]
+is_cxx = {is_cxx!r}
+is_darwin = {is_darwin!r}
+# Same non-link detection as the cross wrapper: compiles, preprocessing,
+# and compiler-introspection probes must not receive link-only flags.
+is_link = True
+for _arg in filtered_args:
+    if _arg in ("-c", "-E", "-S", "-fsyntax-only", "--version", "-dumpmachine", "-dumpversion", "-###") or _arg.startswith("-print-"):
```
jbedard, line 77: What is `-###`?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
+
+# Not a link if compiling/preprocessing (-c/-E/-S/-fsyntax-only) or if this
+# is a compiler-introspection probe (meson runs "-E -v -", "-print-*",
+# "--version" through us): appending link inputs there is at best noise and
+# at worst catastrophic — "-E" *preprocesses* positional inputs, so an
+# appended static archive becomes megabytes of binary on stdout that
+# meson then strictly utf-8 decodes.
+is_link = True
+for a in args:
+    if a in ("-c", "-E", "-S", "-fsyntax-only", "--version", "-dumpmachine", "-dumpversion", "-###") or a.startswith("-print-"):
```
jbedard, line 149: I've seen this before, but previously it didn't have a 5 line comment above it. This makes me wonder about changing these constants to real `.py` or `.py.tmpl` files that can be shared?

uv/private/pep517_whl/common.bzl
```diff
@@ -65,10 +72,16 @@ _PATCH_ATTRS = {
...
-    # same execution platform:
-    # https://bazel.build/extending/exec-groups#defining-exec-groups
-    "tool": attr.label(executable = True, cfg = config.exec(TARGET_EXEC_GROUP)),
+    # The wheel action runs in the "target" exec group; its frontend must run
+    # on the exec (host) platform but resolve its Python build dependencies
+    # for the *target* Python version. exec_transition pins --platforms to the
+    # host and resets the platform_libc/platform_version flags to host values
+    # (Starlark flags otherwise leak from the target configuration and break
+    # wheel selection for build deps). See exec_transition.bzl.
+    "tool": attr.label(executable = True, cfg = exec_transition),
```
jbedard, line 81: Claude, so feel free to ignore:
>  RBE regression — uv/private/pep517_whl/common.bzl:81
Tool attr switched from cfg = config.exec(TARGET_EXEC_GROUP) to target-config transition pinned to --host_platform. Action still runs on exec platform, but frontend py_binary + interpreter + build-dep wheels now built for host. macOS host + linux remote workers: worker gets Mach-O binary, every sdist build dies with "exec format error" — pure-Python sdists included. Docs call RBE "untested", but this breaks a previously-working path.

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
+is_cxx = {is_cxx!r}
+is_darwin = {is_darwin!r}
+# Same non-link detection as the cross wrapper: compiles, preprocessing,
+# and compiler-introspection probes must not receive link-only flags.
+is_link = True
+for _arg in filtered_args:
+    if _arg in ("-c", "-E", "-S", "-fsyntax-only", "--version", "-dumpmachine", "-dumpversion", "-###") or _arg.startswith("-print-"):
+        is_link = False
+        break
+if is_cxx and is_link and not is_darwin:
```
jbedard, line 80: Claude:
> Native builds not untouched — build_helper.py:72
NATIVE-mode _COMPILER_WRAPPER now appends -Wl,-Bstatic -lstdc++ -Wl,-Bdynamic to every non-Darwin C++ link. Hosts without libstdc++.a (Fedora/RHEL minus libstdc++-static, libc++ setups): previously-fine links fail with "cannot find -lstdc++". Where archive exists, wheels silently switch to static libstdc++. Flag not gated to gcc_toolchain despite comment rationale.

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -190,19 +578,115 @@ def _compiler_env(tmpdir: str, execroot_marker: str | None = None) -> dict[str,
...
+            # get_platform() also honors $MACOSX_DEPLOYMENT_TARGET directly.
+            if target_os == "darwin":
+                deployment_target = _macosx_deployment_target(target_sysconfig)
+                if deployment_target:
+                    env.setdefault("MACOSX_DEPLOYMENT_TARGET", deployment_target)
+                    cpu = "arm64" if target_cpu == "aarch64" else target_cpu
+                    env["_PYTHON_HOST_PLATFORM"] = "macosx-{}-{}".format(deployment_target, cpu)
+
+        if target_os and target_cpu:
+            site_dir = _generate_cross_site(tmpdir, target_os, target_cpu, deployment_target)
```
jbedard, line 681: Claude:
> Wrong machine string for darwin cross — build_helper.py:667
sitecustomize fakes platform.machine() as raw target_cpu ("aarch64") but real macOS arm64 reports "arm64" — same file renames aarch64 to arm64 for wheel tag two lines earlier. setup.py branching on platform.machine() == 'arm64' takes wrong path; wheel tag correct so validation passes, corruption silent until runtime.

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -31,22 +37,181 @@
...
+# Standalone (non-"-shared") executables built here are never the wheel
+# deliverable — only throwaway feature probes (meson sanity checks,
+# cc.run()) executed under QEMU, where dynamically-linked target-arch PIE
+# binaries hit a real qemu-user/glibc startup bug ("Inconsistency detected
+# by ld.so: ... DT_VERSYM"), reproducible with a single-toolchain
+# hello-world. "-static" sidesteps ld.so entirely; the .so deliverables
+# always pass "-shared" and load via dlopen's unrelated path. Mutually
+# exclusive with the static-libstdc++ trick: its trailing "-Wl,-Bdynamic"
+# would undo a bare "-static" for driver-appended libs (libc included).
+is_shared = "-shared" in filtered
```
jbedard: Claude:
> Shared-link detection too narrow — build_helper.py:167
Cross wrapper detects shared link only via literal -shared; darwin -bundle/-dynamiclib misclassified, -Wl,-undefined,dynamic_lookup never added, ld64 fails with unresolved _PyObject_*. Also exe_link_flags branch shadows darwin branch even for genuine -shared.

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -269,6 +1101,42 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
         )
     )
 
+_WHEEL_OS_MAP = {"linux": "linux", "darwin": "macosx", "windows": "win"}
+
+
+def _expected_cpu_in_tag(target_os: str, target_cpu: str) -> str:
+    if target_os == "darwin" and target_cpu == "aarch64":
+        return "arm64"
+    return {"x86": "i686", "arm": "armv7l"}.get(target_cpu, target_cpu)
```
jbedard, line 1146: Claude:
> arm CPU map self-contradicts — build_helper.py:1110
_PYTHON_CPU_MAP maps arm→"arm" (wheel tag "linux_arm") but _expected_cpu_in_tag demands "armv7l". Every @platforms//cpu:arm cross build fails own validation on correct wheel. Real armv7 CPython reports "linux-armv7l" — map side is the wrong half.

uv/private/pep517_whl/pep517_native_whl.bzl
```diff
@@ -144,12 +246,60 @@ def _pep517_native_whl(ctx):
...
+        env["RULES_PY_TARGET_OS"] = cc_layer.target_os or ""
+        env["RULES_PY_TARGET_CPU"] = cc_layer.target_cpu or ""
+        if cc_layer.static_runtime_files:
+            extra_inputs.append(cc_layer.static_runtime_files)
+        if cc_layer.static_runtime_paths:
+            env["RULES_PY_CXX_STATIC_RUNTIME"] = ":".join(cc_layer.static_runtime_paths)
+        if cc_layer.cflags:
+            env["CFLAGS"] = cc_layer.cflags
+        if cc_layer.cxxflags:
+            env["CXXFLAGS"] = cc_layer.cxxflags
```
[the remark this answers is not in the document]
jbedard: Claude agrees

uv/private/pep517_whl/pep517_native_whl.bzl
```diff
@@ -116,11 +116,113 @@ def _cc_toolchain_inputs_and_tools(ctx):
...
+    The exec interpreter (EXEC_TOOLS_TOOLCHAIN, selected by exec platform)
+    and the target interpreter (PY_TOOLCHAIN, selected by target platform)
+    carry their origin repo's platform triple. Identical triples prove
+    exec == target, so a native build stays native even when the optional
+    NATIVE_BUILD_TOOLCHAIN sentinel is not registered for the platform.
+
+    When either identity is unrecognizable (custom interpreters), fall back
+    to the sentinel's absence as the cross signal — the previous behavior.
+    """
+    exec_tc = eg_toolchains[EXEC_TOOLS_TOOLCHAIN]
```
jbedard, line 190: Claude:
> Cross detection vs actual runner disagree — pep517_native_whl.bzl:184
_cross_compile decides from exec-group toolchain (exec platform), but frontend that runs is host-transitioned. When exec != host: can pick native mode, host interpreter builds wrong-platform wheel with zero validation.

uv/private/pep517_whl/tools/build_helper.py
jbedard, line 80: Codex:
> Native C++ builds now unconditionally link static libstdc++ (https://github.com/aspect-build/rules_py/blob/87f613b6df95de119fb76630e364cd2652656c71/uv/private/pep517_whl/tools/build_helper.py#L72). The native wrapper appends -Wl,-Bstatic -lstdc++
    -Wl,-Bdynamic for every non-Darwin C++ link. Hosts without libstdc++.a fail, while other hosts silently change wheel linkage behavior. This needs to be limited to the toolchains that require it

uv/private/pep517_whl/common.bzl
jbedard, line 81: Codex:
> The frontend is transitioned to the host platform while the action still executes on the exec platform (https://github.com/aspect-build/rules_py/blob/87f613b6df95de119fb76630e364cd2652656c71/uv/private/pep517_whl/common.bzl#L81). Under heterogeneous
    remote execution, this can place host binaries/build dependencies into an action running on a different worker OS/CPU, causing exec format error even for pure-Python sdists

uv/private/pep517_whl/tools/build_helper.py
jbedard: Codex
> Darwin shared-extension links are misclassified (https://github.com/aspect-build/rules_py/blob/87f613b6df95de119fb76630e364cd2652656c71/uv/private/pep517_whl/tools/build_helper.py#L167). The cross wrapper recognizes only literal -shared; macOS
    extensions normally use -bundle or -dynamiclib. With LLVM-style link flags, the earlier exe_link_flags branch also prevents adding -Wl,-undefined,dynamic_lookup, so linking C-extension modules fails on unresolved Python symbols.

uv/private/pep517_whl/tools/build_helper.py
jbedard: Codex:
> Darwin ARM target identity uses aarch64 where macOS reports arm64 (https://github.com/aspect-build/rules_py/blob/87f613b6df95de119fb76630e364cd2652656c71/uv/private/pep517_whl/tools/build_helper.py#L667). sitecustomize fakes platform.machine() as the
    raw Bazel CPU value, although wheel-tag code maps the same value to arm64. Packages branching on platform.machine() == "arm64" will take the wrong build path despite producing a tag that passes validation.

uv/private/pep517_whl/pep517_native_whl.bzl
xangcastle: 11 is the first one with apple silicon

xangcastle:
Closing per team decision: even split out of #1434, this slice still bundles too many behavior changes (frontend configuration, CC toolchain extraction, per-backend cross env, target-identity faking, wheel validation) without targeted tests for each. The review findings above confirmed it — several logic paths had no coverage that could have caught them.

The cross-compile capability will be rebuilt as a series of surgical PRs, each independently tested (unit tests for the helper logic, analysis tests for the rule changes, one e2e case per backend as it's enabled), using #1434 as the spike reference. The review feedback here — especially the exec-configuration/RBE discussion — feeds directly into that series.

PR #1462, opened 2026-08-18
perf: defer rule-group 0755 ownership to the mtree action

jbedard:
The 0755-vs-0644 decision for rule-group files was made at analysis time by flattening the source closure into a path dict captured by the mtree mapper. The tar action now receives the source closure as a path-string set (directories marked with a trailing slash and prefix-matched, so trees are never expanded and the files never become action inputs) and awk rewrites matching rows to 0755. Output is byte-identical.

### Test plan

### DIFF
--- a/py/private/modify_mtree.awk
+++ b/py/private/modify_mtree.awk
@@ -85,6 +85,33 @@ function decode_mtree_path(path) {
     return path
 }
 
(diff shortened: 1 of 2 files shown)


jbedard:
No point until other analysis stuff is gone

PR #1463, opened 2026-08-19
refactor: make build_helper.py import-safe, add unit test harness

xangcastle:
`build_helper.py` runs its CLI body at module level, so importing the module executes a build — which is why none of its helpers had unit tests (`memory_monitor_test` deliberately imports only `memory_monitor`). This moves the body under `main()` with a `__main__` guard and adds the missing coverage.

- **Pure reindentation**: `git diff -w` on the helper shows only the `def main() -> None:` line and the guard; no logic changes.
- **New `build_helper_test`**: table-driven unit tests for the pure helpers — `_absolutize_path`, `_resolve_compiler_path` (defaults, PATH lookup, flag-carrying values), `_local_cxx_companion` (gcc/g++, versioned clang, cross-prefixed drivers, missing companion), `_override_tool`, `_make_compiler_wrapper`, and `_legacy_metadata_conflicts_with_pyproject` (static vs dynamic vs legacy metadata shapes). The import at the top of the test file is itself the regression test for the guard.

The generated build tools invoke the helper as a script; the `__main__` guard keeps that entry point identical.

### Test plan

- New `//uv/private/pep517_whl/tests:build_helper_test` (27 cases)
- `bazel test //uv/private/pep517_whl/...`: the existing wheel-building integration tests re-ran against the refactored helper and pass
- `ruff check`, `pyright --pythonversion 3.10`, and `ty check` (the typecheck CI chain) pass locally on the touched files

### DIFF
--- a/uv/private/pep517_whl/tests/BUILD.bazel
+++ b/uv/private/pep517_whl/tests/BUILD.bazel
@@ -28,6 +28,13 @@ py_test(
 )
 
 py_test(
+    name = "build_helper_test",
+    srcs = ["build_helper_test.py"],
+    main = "build_helper_test.py",
+    deps = ["//uv/private/pep517_whl/tools"],
+)
+
+py_test(
     name = "local_cxx_companion_test",
     srcs = ["local_cxx_companion_test.py"],
     data = ["//uv/private/pep517_whl/tools:build_helper.py"],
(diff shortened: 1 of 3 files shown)


xangcastle:
@codex

xangcastle:
@codex

jbedard (approved):

PR #1464, opened 2026-08-19
feat(uv): validate the built wheel's platform tag against a requested target

xangcastle:
`build_helper` gains `--target-os` / `--target-cpu` and, when both are set, validates the produced wheel's PEP 427 platform tag (the last dash-separated component of the backend's filename) before renaming onto the fixed output:

- the tag must name the requested OS (`linux`/`macosx`/`win`) and CPU
- `-none-any` wheels are exempt — they carry no platform identity
- a tag naming the **exec host's** OS on a foreign-target build gets a dedicated error: that signature means the backend compiled for the machine it ran on, which is worth distinguishing from a generic mismatch
- CPU spellings follow what interpreters actually report: darwin `aarch64` → `arm64`, `arm` → `armv7l`, `x86` → `i686`

The check lives in a pure function (`_wheel_platform_error`, returns an error string or `None`, `host_os` injectable) so the full decision table is unit-tested — matching manylinux/musllinux/macosx tags, case normalization, both leak directions, wrong-OS and wrong-CPU, and the arm64/aarch64 and armv7l spellings.

Nothing passes the new flags yet; without them the validation short-circuits and existing builds behave identically.

### Test plan

- 12 new unit-test cases in `//uv/private/pep517_whl/tests:build_helper_test` (decision table + flag parsing defaults)
- `bazel test //uv/private/pep517_whl/...`: 10 passed, 2 skipped (platform-incompatible locally)
- `ruff check`, `pyright --pythonversion 3.10`, `ty check` pass on the touched files

### DIFF
--- a/uv/private/pep517_whl/tools/build_helper.py
+++ b/uv/private/pep517_whl/tools/build_helper.py
@@ -269,6 +269,77 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
         )
     )
 
+
+def _wheel_platform_identity(target_os: str, target_cpu: str) -> tuple[str, str]:
+    """The wheel platform-tag OS and CPU spellings for a Bazel OS/CPU constraint pair."""
+    if target_os == "darwin":
+        wheel_os = "macosx"
+        wheel_cpu = "arm64" if target_cpu == "aarch64" else target_cpu
+    elif target_os == "windows":
+        wheel_os = "win"
+        if target_cpu == "x86_64":
+            wheel_cpu = "amd64"
+        elif target_cpu == "aarch64":
+            wheel_cpu = "arm64"
+        else:
+            wheel_cpu = target_cpu
+    else:
+        wheel_os = target_os
+        if target_cpu == "x86":
+            wheel_cpu = "i686"
+        elif target_cpu == "arm":
+            wheel_cpu = "armv7l"
+        else:
+            wheel_cpu = target_cpu
+    return wheel_os, wheel_cpu
+
+
+def _wheel_platform_error(
+    wheel_filename: str,
+    target_os: str,
+    target_cpu: str,
+    host_os: str | None = None,
+) -> str | None:
+    """Check a built wheel's platform tag against the requested target platform.
+
+    Returns an error message when the tag names the wrong platform, None when
+    it matches or no target was requested. The platform tag is the last
+    dash-separated component of the filename (PEP 427); -none-any wheels are
+    exempt — they carry no platform identity. host_os is injectable for tests
+    and defaults to the running host.
+    """
+    if not target_os or not target_cpu:
+        return None
+    if wheel_filename.endswith("-none-any.whl"):
+        return None
+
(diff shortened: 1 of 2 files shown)


xangcastle:
@codex

xangcastle:
@codex

xangcastle:
@codex

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -269,6 +269,71 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
...
+    """The CPU spelling a wheel platform tag uses for a Bazel CPU constraint.
+
+    macOS reports arm64, never the constraint's aarch64; Windows tags are
+    win_amd64/win_arm64; armv7 CPython reports armv7l; ia32 is i686 in
+    linux wheel tags.
+    """
+    if target_os == "darwin" and target_cpu == "aarch64":
+        return "arm64"
+    if target_os == "windows":
+        return {"x86_64": "amd64", "aarch64": "arm64"}.get(target_cpu, target_cpu)
```
jbedard: Why the inline `{..}.get()` instead of just `if/else`?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -269,6 +269,71 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
         )
     )
 
+
+_WHEEL_OS_MAP = {"linux": "linux", "darwin": "macosx", "windows": "win"}
+
+
+def _expected_cpu_in_tag(target_os: str, target_cpu: str) -> str:
+    """The CPU spelling a wheel platform tag uses for a Bazel CPU constraint.
```
jbedard:
```suggestion
    """The python CPU wheel platform tag for the provided Bazel OS/CPU constraint.
```
?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -269,6 +269,71 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
...
+
+    macOS reports arm64, never the constraint's aarch64; Windows tags are
```
jbedard: This seems to be describing the entire implementation. Isn't one sentence without describing the implementation enough?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -269,6 +269,71 @@ def _legacy_metadata_conflicts_with_pyproject(worktree: str) -> bool:
         )
     )
 
+
+_WHEEL_OS_MAP = {"linux": "linux", "darwin": "macosx", "windows": "win"}
+
+
+def _expected_cpu_in_tag(target_os: str, target_cpu: str) -> str:
```
jbedard: Expected bazel CPU? Or Python CPU? Why does a CPU need to know the os? Why not a single method that converts both bazel os+cpu into python instead of 1 method and 1 constant?

jbedard (approved):

PR #1465, opened 2026-08-19
feat(uv): fail cross sdist builds with an explicit analysis error

xangcastle:
Building a native sdist for a target platform that no execution platform matches used to die with Bazel's opaque toolchain-resolution error (`no matching toolchains found for @aspect_rules_py//py/private/toolchain:native_build_toolchain_type`), leaving users to reverse-engineer what it meant. `pep517_native_whl` now detects the situation itself and fails with an actionable message naming both interpreter platforms.

**Detection** (analysis time, in the rule impl):
- When both the exec-configured and target-configured interpreters come from recognizable hermetic repos, their platform triples decide: equal → native, different → cross. This also *keeps a build native* when the triples match but the `NATIVE_BUILD_TOOLCHAIN` sentinel isn't registered for the platform — previously that combination failed resolution.
- When either identity is unrecognizable (custom interpreters), the sentinel's absence remains the cross signal, preserving today's semantics.

**Toolchain changes**: the sentinel, `EXEC_TOOLS_TOOLCHAIN`, and the C++ toolchain become `mandatory = False` in the `target` exec group (and the rule gains an optional target-level `PY_TOOLCHAIN` to read the target triple) so the rule's own error wins over resolution failures in any of them. Native builds resolve exactly as before — the full existing analysis/integration suite is the non-regression check.

Cross builds of native sdists still fail, but with an explicit error naming the execution and target interpreter platforms instead of a toolchain-resolution failure.

### Test plan

- `interpreter_platform_triple_test` (skylib unittest): PBS repo names, underscore-sanitized variants, rules_python-style repos, main-workspace paths, unrecognizable/custom interpreters
- `cross_decision_test`: full decision table (triples agree/disagree × sentinel present/absent)
- `cross_unsupported_test` (analysistest, `expect_failure`): transitions a minimal fixture to linux-aarch64 and asserts the new message; skipped on linux-aarch64 hosts where that target is native
- `bazel test //uv/...`: 167 passed, 2 skipped (platform-incompatible locally)

### DIFF
--- a/uv/private/pep517_whl/BUILD.bazel
+++ b/uv/private/pep517_whl/BUILD.bazel
@@ -19,6 +19,7 @@ bzl_library(
     srcs = ["pep517_native_whl.bzl"],
     deps = [
         ":common",
+        "//py/private/interpreter:versions",
         "//py/private/toolchain:types",
         "@bazel_lib//lib:resource_sets",
         "@rules_cc//cc:action_names_bzl",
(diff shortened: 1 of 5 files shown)


xangcastle:
@codex

xangcastle:
@codex

jbedard:
@codex

jbedard (approved):

PR #1469, opened 2026-08-20
feat(uv): support test-only Python packages

zbarsky-openai:
Allow uv.override_package(..., testonly = True) to attach test-only
dependencies or data to a resolved wheel without violating Bazel's
test-only dependency boundary.

Propagate the restriction through wheel installation, generated
dependency groups, project aliases, public hub aliases, and packages
that depend on the marked package. Keep raw wheel targets unrestricted
and distinguish test-only installs from production installs during
cross-project deduplication.

Document the new override and exercise a patched package carrying
test-only data through both standard and exposed-virtualenv tests.
Existing generated-repository snapshots remain unchanged for packages
without the override.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,13 @@ uv.override_package(
     exclude_glob = ["another_package/**/tests/**"],
 )
 
+# 3d. (Optional) Restrict a package and its dependents to test targets
+uv.override_package(
+    name = "pytest-postgresql",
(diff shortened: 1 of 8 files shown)


xangcastle (approved):

xangcastle:
Superseded by #1553 (rebased on current main; no functional changes).

PR #1470, opened 2026-08-20
feat(uv): parameterize the platform parent in gazelle_python_manifest

### LINKED ISSUE #1416: [Bug]: gazelle_python_manifest hard-codes @platforms//host, breaks cross-platform builds

xangcastle:
Fixes #1416. Supersedes #1468, carrying @sallustfire's commit unchanged (cherry-picked, authorship preserved) plus three follow-up commits.

## Base change (from #1468, by @sallustfire)

`gazelle_python_manifest`'s synthetic per-venv platforms hardcoded `parents = ["@platforms//host"]`, so the `platform_transition_filegroup` over the hub's wheel index replaced the caller's `--platforms` wholesale. That broke custom `--host_platform` setups (hermetic cc toolchains) and cross-compilation. The fix parameterizes the parent platform; the default preserves current behavior and is resolved as a `Label` in rules_py's own repository, so consumers need no `bazel_dep` on `platforms`.

## Follow-up commits

**`platform_parent` is a single label, not a list.** The list form failed at load time on any length other than one — promising plurality the underlying `platform()` rule does not support. Renamed and made singular before the API ships. The load-time check now rejects any non-string, non-Label value (lists, selects) with a message naming the parameter.

**Wheel selection is pinned under a real cross-compilation parent.** The e2e case's parent is now a fully-specified `linux_x86_64` platform carrying the `platform_libc`/`platform_version` flags — issue #1416's actual scenario, and host-invariant. The flags ride on the platform because their defaults are host-detected: without them, manylinux `config_setting`s never match off-linux and native packages silently degrade to sdist fallbacks (the other half of the config chimera the issue describes; this is also the documented recipe for real cross-compilation parents). A new analysis-time probe snapshots which wheel variant each hub dependency resolves to (`distribution: platform-class` lines — no versions or tag spellings, so lock bumps don't churn the golden). Native packages pin to `linux_x86_64`, pure ones to `any`, and sdist fallbacks surface explicitly. The manifest YAML golden is replaced by a `build_test`: the YAML content is platform-independent, so it only re-pinned another case's hub state; building the manifest end-to-end from linux-selected wheels is the actual signal.

**Gazelle manifest tooling moved to a `tools/` subpackage.** `generate.py` and `update.sh` move out of the package holding the public `defs.bzl` into `//uv/private/gazelle_manifest/tools` with absolute labels; implicit-attribute visibility is checked against the defining `.bzl`'s package, so `:generator` stays private.

## Test plan

- 37/37 e2e/cases tests pass (36 snapshot diff tests + the new `manifest_build_test`); goldens regenerated via `bazel run //:snapshots` on darwin_arm64 — the wheel-selection golden shows `linux_x86_64` wheels selected from a mac host.
- Regression detection verified by simulating the revert (macro ignoring `platform_parent`): the constraint probe flips to `default-parent` and the wheel probe flips to `macosx_*` wheels.
- `//uv/private/gazelle_manifest/tests:test` passes; buildifier clean on changed files.

---

- [x] Changes are visible to end-users: yes
- [x] Searched for relevant documentation and updated as needed: yes (`docs/uv.md` Gazelle section)
- [x] Breaking change (forces users to change their own code or config): no
- [x] Suggested release notes appear below: yes

`gazelle_python_manifest` now accepts `platform_parent`, letting the manifest's per-venv wheel selection parent onto a custom platform. Needed for repos with a custom `--host_platform` (hermetic C++ toolchains) and for cross-compilation; in the cross-compilation case, carry the `platform_libc`/`platform_version` flags on the parent platform.


xangcastle:
@codex

xangcastle:
@codex

e2e/cases/gazelle-platform-parents-1416/wheel_selection_probe.bzl
```diff
@@ -0,0 +1,48 @@
...
+the golden to exact versions or manylinux tag spellings, both of which churn
+on lock updates. Classification happens at analysis time from basenames, so
+the probe never triggers sdist builds for the transitioned platform.
+"""
+
+def _classify_platform_tag(basename):
+    tag = basename[:-len(".whl")].split("-")[-1]
+    if tag == "any":
+        return "any"
+    os = "unknown-os"
```
jbedard, line 15: Maybe drop the `-os` and `-cpu` so the result would just be `unknown_x86` or `linux_unknown`?

jbedard (approved):

PR #1471, opened 2026-08-20
fix(uv): dedup sdist build deps discovered by the configure tool

xangcastle:
## Problem

Build deps reported by the sdist configure tool (`extra_deps`) could duplicate the explicit deps of the generated build target. Because the explicit label (`@project__*//:pkg`) and the `available_deps` label (`@whl_install__*__pkg__*//:install`) live in different repos, they can resolve to different versions of the same package and collide in the `build_tool` venv when `package_collisions = "error"`.

## Fix

`_resolve_extra_deps` now skips `extra_deps` names already covered by the explicit deps (matched by normalized package name) and dedups resolved labels. The configure-tool contract doc is updated to reflect the skip/dedup behavior and the narrowed `fail()` condition.

## Test

New e2e case `uv-sdist-dup-build-deps`: `setuptools` locked at two versions (fork on `python_full_version`) and also provided via `default_build_dependencies`, so the explicit and discovered labels previously collided in the build_tool venv. `docopt` (setup.py-only sdist) exercises the inferred-build-dep path (`wheel` resolved from the lockfile).

### DIFF
--- a/e2e/MODULE.bazel
+++ b/e2e/MODULE.bazel
@@ -508,6 +508,23 @@ uv.override_package(
 use_repo(uv, "pypi-native-inputs")
 # }}}
 
+# For cases/uv-sdist-dup-build-deps
+# Regression: build deps discovered by the sdist configure tool must be
+# deduplicated by package name against the explicit deps. With setuptools
+# locked at two versions, the duplicate labels resolve to different versions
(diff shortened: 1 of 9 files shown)


xangcastle:
@codex

jbedard (approved):

PR #1473, opened 2026-08-20
fix(uv): fan out group dependencies over marker-gated locked versions

xangcastle:
## Problem

A dependency group may lock the same package at several versions gated by disjoint PEP 508 markers (platform/python forks, uv conflict routing). `_extract_lockfile_group_versions` collapsed the group's `dev-dependencies` entries into a single-preference map last-write-wins, so the walk wired one arbitrary version of the package for every platform.

## Changes

- Keep every `dev-dependencies` entry as a `(dep, marker)` candidate. Marker-gated candidates fan out over every locked version — in the seeds and in the reach-set walk — conjoining the requirement/edge markers with each candidate's lockfile marker. uv guarantees candidates within a group carry disjoint markers, so the `decide_marker` select() layer resolves exactly one per environment.
- Preserve the marker of single marker-gated candidates instead of promoting them to unconditional preferences.
- Keep transitive edge versions outside the candidate set as a fallback, so environments where no candidate marker holds still wire a version instead of silently dropping the package.
- Resolve requirement specs for fanned-out packages through a placeholder version map so legal spec forms (direct references, `!=` against post-releases) do not fall through to the specifier matcher and `fail()`.
- Expand each reach-set walk node exactly once instead of re-expanding duplicated seeds.
- Make `combine_markers` idempotent (`(m) and (m)` → `m`); uv mirrors the requirement marker into the lockfile entry, so this was inflating select arms in the common case.
- `fail()` loudly when two packages claim the same configuration via extra-only markers instead of silently taking the first writer — build-time selection cannot disambiguate uv conflict extras, and a hard error beats a wrong wheel.

## Testing

- New unit tests: `collect_activated_extras_group_multi_version_test` (direct multi-version fan-out) and `collect_activated_extras_group_marker_fallback_test` (single marker-gated candidate + transitive version outside the candidate set).
- New e2e case `uv-group-multi-version`: `six` locked at 1.16.0 (`sys_platform == 'linux'`) and 1.17.0 (`sys_platform != 'linux'`) in one group, asserted at runtime.
- `bazel test //uv/...` (86/86) and full e2e suite green locally (the only failure is the docker-daemon-dependent `interpreter-features-836`, unrelated).

### DIFF
--- a/uv/private/uv_project/repository.bzl
+++ b/uv/private/uv_project/repository.bzl
@@ -149,35 +149,37 @@ load("@aspect_rules_py//py:defs.bzl", "py_library")
             # to true. It's a configuration and locking failure for there to be
             # more than one package which resolves at this point. So we just jam all the configurations into a single select.
             for scc, markers in scc_cfgs.items():
+                scc_target = "//private/sccs:" + scc
+
                 if "" in markers:
                     if "//conditions:default" in cfg_arms:
                         fail("Configuration conflict! Package {} specifies two or more default package states!\n{}".format(package, pprint(cfgs)))
 
-                    cfg_arms["//conditions:default"] = "//private/sccs:" + scc
+                    cfg_arms["//conditions:default"] = scc_target
 
                 elif all([is_extra_only_marker(m) for m in markers.keys()]):
                     # Extra-only markers cannot be evaluated at build time; they
                     # are already resolved by the active venv. Treat the SCC as
                     # the default for this configuration.
-                    if cfg_arms.get("//conditions:default", "//private/sccs:" + scc) != "//private/sccs:" + scc:
+                    if cfg_arms.get("//conditions:default", scc_target) != scc_target:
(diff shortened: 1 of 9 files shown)


uv/private/uv_project/repository.bzl
```diff
@@ -159,15 +159,19 @@ load("@aspect_rules_py//py:defs.bzl", "py_library")
                     # Extra-only markers cannot be evaluated at build time; they
                     # are already resolved by the active venv. Treat the SCC as
                     # the default for this configuration.
-                    if "//conditions:default" not in cfg_arms:
-                        cfg_arms["//conditions:default"] = "//private/sccs:" + scc
+                    if cfg_arms.get("//conditions:default", "//private/sccs:" + scc) != "//private/sccs:" + scc:
```
jbedard: The `"//private/sccs:" + scc` is repeated 3 times within 3 lines, so put it in a var that has a name describing what it is?

jbedard:
@codex

jbedard:
@codex

jbedard:
Please open a PR for `main` with at least the tests from this, even if no fix is read yet.

jbedard (approved):

PR #1482, opened 2026-08-25
test(uv): frontend exec configuration must not inherit the target's platform flags

xangcastle:
Characterization first: three analysis tests pin what `cfg = config.exec(TARGET_EXEC_GROUP)` gives the PEP 517 frontend, via a probe rule that encodes the `platform_libc` value it was configured with into its executable's file name — the only channel visible at analysis time (tool runfiles arrive as an opaque middleman).

- **`frontend_exec_config_test`** — the frontend builds in an `-exec` output directory: the property remote execution depends on (the binary must match the worker, not the target).
- **`frontend_libc_leak_test`** — `--//uv/private/constraints/platform:platform_libc` set in the **target** configuration survives the exec transition and reaches the frontend, so the frontend's own Python build dependencies get selected with the *target's* libc. Pinned as **CURRENT BEHAVIOR (defect)**: the fix in this PR must flip this assertion explicitly.
- **`frontend_probe_plumbing_test`** — guards the probe channel itself so the leak test cannot pass vacuously (it already caught one dead channel: runfile-name markers never reach analysis-visible inputs).

The fix itself lands on top of these tests in this same PR; any approach that breaks the `-exec` assertion or leaves the leak assertion un-flipped is rejected by construction.

### Test plan

- The three new analysis tests above
- `bazel test //uv/private/pep517_whl/...`: 18 passed, 2 skipped (platform-incompatible locally)

### DIFF
--- a/uv/private/pep517_whl/tests/BUILD.bazel
+++ b/uv/private/pep517_whl/tests/BUILD.bazel
@@ -12,6 +12,13 @@ load(
     "exec_platform_steering_test",
 )
 load(
+    ":frontend_config_test.bzl",
+    "flag_probe",
+    "frontend_exec_config_test",
+    "frontend_libc_leak_test",
+    "frontend_probe_plumbing_test",
+)
+load(
     ":pep517_whl_test.bzl",
     "cross_decision_test",
     "execroot_collision_toolchain",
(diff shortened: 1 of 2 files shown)


xangcastle:
@codex

jbedard (approved):

PR #1488, opened 2026-08-26
feat(oci): every bsdtar compression filter, plus custom compressors

thesayyn:
Add support for alternate compression algorithms by opening up libarchive's full write-filter set: none, gzip, bzip2, xz, lzma, lzop, lz4, lrzip, zstd, compress. New `py_layer_compressor` covers anything else by piping the archive through a user-supplied program.

### DIFF
--- a/py/defs.bzl
+++ b/py/defs.bzl
@@ -20,6 +20,11 @@ load("@rules_python//python:packaging.bzl", _py_wheel = "py_wheel")
 load("@rules_python//python:pip.bzl", _whl_filegroup = "whl_filegroup")
 load("@rules_python//python:py_runtime.bzl", _py_runtime = "py_runtime")
 load("@rules_python//python:py_runtime_pair.bzl", _py_runtime_pair = "py_runtime_pair")
+load(
+    "//py/private:compression.bzl",
+    _PyLayerCompressorInfo = "PyLayerCompressorInfo",
+    _py_layer_compressor = "py_layer_compressor",
+)
 load("//py/private:providers.bzl", _PyWheelsInfo = "PyWheelsInfo")
 load(
     "//py/private:py_image_layer.bzl",
@@ -63,7 +68,9 @@ py_unpacked_wheel = _py_unpacked_wheel
 
 py_image_layer = _py_image_layer
 py_layer_tier = _py_layer_tier
+py_layer_compressor = _py_layer_compressor
 PyLayerTierInfo = _PyLayerTierInfo
+PyLayerCompressorInfo = _PyLayerCompressorInfo
 
 # The PyInfo provider used by rules_py
 PyInfo = _PyInfo
--- a/py/private/compression.bzl
+++ b/py/private/compression.bzl
@@ -1,0 +1,279 @@
+"""Compression codecs for `py_image_layer` layer tars.
+
+Layer tars are written by bsdtar, so the usable set is libarchive's write-filter
+set, mapped here onto bsdtar flags and file extensions. `py_layer_compressor`
+covers everything else via `--use-compress-program`.
+
+libarchive implements lzop and lrzip by shelling out to a same-named binary,
+which a sandboxed action is not guaranteed to have; `py_layer_compressor` is
+the hermetic answer there.
+
+The OCI image spec only defines tar, gzip and zstd layers, and rules_oci sniffs
+exactly those two magics — anything else is labelled an uncompressed tar and
+gets the *compressed* digest as its `diffid`, which is a broken image rather
+than a build error. So `oci` marks the codecs an image consumer can actually
+read, and the rules refuse the rest unless `allow_non_oci_layers` says the tars
+are headed somewhere other than an OCI image.
+"""
+
+PyLayerCompressorInfo = provider(
+    doc = "A custom layer compressor: a program bsdtar pipes the archive through.",
+    fields = {
+        "args": "tuple[str] — arguments appended after the program path.",
+        "executable": "File — the program bsdtar invokes via --use-compress-program.",
+        "extension": "str — output file extension, e.g. '.tar.br'.",
+        "files_to_run": "FilesToRunProvider — the program plus its runfiles, for the tar action's `tools`.",
+    },
+)
+
+# flag: selects the write filter. filter: the `--options` module accepting
+# `compression-level`, None when the filter has no level knob. oci: the OCI
+# image spec defines a layer format for it.
+ALGORITHMS = {
+    "bzip2": struct(flag = "--bzip2", filter = "bzip2", ext = ".tar.bz2", min_level = 1, max_level = 9, oci = False),
+    "compress": struct(flag = "--compress", filter = None, ext = ".tar.Z", min_level = 0, max_level = 0, oci = False),
+    "gzip": struct(flag = "--gzip", filter = "gzip", ext = ".tar.gz", min_level = 0, max_level = 9, oci = True),
+    "lrzip": struct(flag = "--lrzip", filter = "lrzip", ext = ".tar.lrz", min_level = 1, max_level = 9, oci = False),
+    # lz4 rejects level 0 outright ("Undefined option"), unlike gzip/xz/lzma.
+    "lz4": struct(flag = "--lz4", filter = "lz4", ext = ".tar.lz4", min_level = 1, max_level = 9, oci = False),
+    "lzma": struct(flag = "--lzma", filter = "lzma", ext = ".tar.lzma", min_level = 0, max_level = 9, oci = False),
+    "lzop": struct(flag = "--lzop", filter = "lzop", ext = ".tar.lzo", min_level = 1, max_level = 9, oci = False),
+    "none": struct(flag = None, filter = None, ext = ".tar", min_level = 0, max_level = 0, oci = True),
+    "xz": struct(flag = "--xz", filter = "xz", ext = ".tar.xz", min_level = 0, max_level = 9, oci = False),
+    "zstd": struct(flag = "--zstd", filter = "zstd", ext = ".tar.zst", min_level = -131072, max_level = 22, oci = True),
+}
+
+SUPPORTED_ALGORITHMS = sorted(ALGORITHMS.keys())
+OCI_ALGORITHMS = sorted([name for name, spec in ALGORITHMS.items() if spec.oci])
+
+# A custom compressor declares what it emits through its extension; only these
+# three land on a layer an OCI consumer can read.
+_OCI_EXTENSIONS = {spec.ext: name for name, spec in ALGORITHMS.items() if spec.oci}
+
+# libarchive's gzip filter stamps wall-clock into the header's MTIME field, so
+# identical inputs would yield different layer digests. `!timestamp` zeroes it.
+# No other write filter has an equivalent field.
+_EXTRA_OPTIONS = {
+    "gzip": ["gzip:!timestamp"],
+}
+
+DEFAULT_ALGORITHM = "gzip"
+DEFAULT_LEVEL = "6"
+
+def _is_integer(value):
+    digits = value[1:] if value.startswith("-") else value
+    return len(digits) > 0 and digits.isdigit()
+
+def _attr_suffix(attr_path):
+    """Parenthesized ` (py_layer_tier.compression["x"])` for an error message, or "" when unknown."""
+    return " (%s)" % attr_path if attr_path else ""
+
+def make_codec(algorithm, level, attr_path = ""):
+    """Resolve a builtin bsdtar algorithm + level into a codec struct.
+
+    Args:
+        algorithm: One of `SUPPORTED_ALGORITHMS`.
+        level: Compression level as a string, or "" to use libarchive's default.
+        attr_path: The attribute the values came from, named in error messages,
+            e.g. `py_layer_tier.compression["x"]`. Omit when there is no
+            user-visible attribute to point at.
+
+    Returns:
+        A codec struct consumed by `codec_tar_args` / `codec_tools`.
+    """
+    spec = ALGORITHMS.get(algorithm, None)
+    if spec == None:
+        fail("unknown compression algorithm %r%s. Supported: %s. For anything else, " %
+             (algorithm, _attr_suffix(attr_path), ", ".join(SUPPORTED_ALGORITHMS)) +
+             "declare a py_layer_compressor and point `compressors` at it.")
+
+    options = list(_EXTRA_OPTIONS.get(algorithm, []))
+    if level:
+        if not _is_integer(level):
+            fail("compression level must be an integer, got %r%s" % (level, _attr_suffix(attr_path)))
+        if spec.filter == None:
+            fail("the %r compressor takes no compression level%s; drop the level or " %
+                 (algorithm, _attr_suffix(attr_path)) + "pick an algorithm that has one.")
+        as_int = int(level)
+        if as_int < spec.min_level or as_int > spec.max_level:
+            fail("compression level %s is out of range for %r%s: expected %d..%d" %
+                 (level, algorithm, _attr_suffix(attr_path), spec.min_level, spec.max_level))
+        options.append("{}:compression-level={}".format(spec.filter, level))
+
+    return struct(
+        algorithm = algorithm,
+        flag = spec.flag,
+        options = ",".join(options) if options else None,
+        ext = spec.ext,
+        oci_compatible = spec.oci,
+        program = None,
+        program_format = None,
+        files_to_run = None,
+    )
+
+def codec_from_compressor(info):
+    """Resolve a `PyLayerCompressorInfo` into a codec struct.
+
+    Args:
+        info: The `PyLayerCompressorInfo` of a `py_layer_compressor` target.
+
+    Returns:
+        A codec struct consumed by `codec_tar_args` / `codec_tools`.
+    """
+    suffix = "".join([" " + arg for arg in info.args])
+    return struct(
+        algorithm = "custom",
+        flag = None,
+        options = None,
+        ext = info.extension,
+        # The program's bytes are opaque to us; its declared extension is the
+        # only claim we have about what an image consumer would find.
+        oci_compatible = info.extension in _OCI_EXTENSIONS,
+        program = info.executable,
+        # bsdtar takes `program arg...` as one argv entry and splits it itself.
+        program_format = "%s" + suffix,
+        files_to_run = info.files_to_run,
+    )
+
+def check_oci_compatible(codec, group_name, attr_path, allowed):
+    """Reject a codec no OCI image consumer can read, unless opted out of OCI.
+
+    Args:
+        codec: A codec struct from `make_codec` or `codec_from_compressor`.
+        group_name: The layer group the codec was selected for, named in the error.
+        attr_path: The attribute that selected it, e.g. `py_layer_tier.compression`.
+        allowed: True when the target set `allow_non_oci_layers`.
+    """
+    if codec.oci_compatible or allowed:
+        return
+    if codec.program != None:
+        detail = ("a py_layer_compressor declaring extension %r" % codec.ext)
+    else:
+        detail = "%r" % codec.algorithm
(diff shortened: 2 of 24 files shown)


xangcastle:
@codex

py/private/compression.bzl
```diff
@@ -0,0 +1,229 @@
...
+)
+
+# flag: selects the write filter. filter: the `--options` module accepting
+# `compression-level`, None when the filter has no level knob.
+ALGORITHMS = {
+    "bzip2": struct(flag = "--bzip2", filter = "bzip2", ext = ".tar.bz2", min_level = 1, max_level = 9),
+    "compress": struct(flag = "--compress", filter = None, ext = ".tar.Z", min_level = 0, max_level = 0),
+    "gzip": struct(flag = "--gzip", filter = "gzip", ext = ".tar.gz", min_level = 0, max_level = 9),
+    "lrzip": struct(flag = "--lrzip", filter = "lrzip", ext = ".tar.lrz", min_level = 1, max_level = 9),
+    "lz4": struct(flag = "--lz4", filter = "lz4", ext = ".tar.lz4", min_level = 0, max_level = 9),
```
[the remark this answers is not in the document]
thesayyn: Confirmed and fixed — `lz4` min_level is now 1.

Reproduced against the bsdtar the tar.bzl toolchain ships (libarchive 3.8.1), where lz4 is the only one of the level-taking filters that rejects 0:

```
lz4 level 0: tar: Undefined option: `lz4:compression-level'
lz4 level 1: OK
gzip/xz/lzma level 0: OK
```

So the shared 0..9 range would have passed analysis and failed the tar action. Added `lz4_zero_test` as a regression, and the codec table test now pins both `lz4`'s floor of 1 and `gzip`'s of 0 so the two don't get collapsed again.

---
_Generated by [Claude Code](https://claude.ai/code)_

py/private/py_image_layer.bzl
```diff
@@ -944,16 +979,40 @@ def _run_tar_action(ctx, bsdtar, bsdtar_files, tar_out, files_depset, map_each,
...
+    """Codecs the rule's own attrs pin, group name → codec."""
+    codecs = {}
+    for group_name, value in ctx.attr.group_compression.items():
+        codecs[group_name] = parse_compression(value, "py_image_layer.group_compression[%r]" % group_name)
+    for dep, group_name in ctx.attr.group_compressors.items():
+        if group_name in codecs:
+            fail(("group %r is configured in both py_image_layer.group_compression and " +
+                  "py_image_layer.group_compressors. A group gets its bytes from exactly " +
+                  "one compressor — drop one of the two entries.") % group_name)
+        codecs[group_name] = codec_from_compressor(dep[PyLayerCompressorInfo])
```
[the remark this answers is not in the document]
thesayyn: Confirmed and fixed, and it was worse than reported: the tier-level check you cite as the working precedent was **also** dead code.

Both loops ordered the checks the same way:

```starlark
if group_name in codecs:      # "configured in both ..."
if group_name in seen:        # "mapped to two compressors" -- unreachable
```

The first compressor for a group already puts it in `codecs`, so a second one always fell into the *first* branch and reported a `compression`/`compressors` conflict that doesn't exist. Writing the test for `py_image_layer` is what surfaced it — it failed with the wrong message rather than passing.

Both now test the duplicate first, then the cross-attribute conflict. Regression tests: `rival_compressors_test` (rule) and `rival_compressors_tier_test` (tier).

---
_Generated by [Claude Code](https://claude.ai/code)_

py/private/compression.bzl
```diff
@@ -0,0 +1,230 @@
...
+}
+
+DEFAULT_ALGORITHM = "gzip"
+DEFAULT_LEVEL = "6"
+
+def _is_integer(value):
+    digits = value[1:] if value.startswith("-") else value
+    return len(digits) > 0 and digits.isdigit()
+
+def _describe(where):
```
jbedard: This is not a very descriptive method name, argument names or docs

thesayyn: Fair — renamed and documented in `949516f`.

```starlark
def _attr_suffix(attr_path):
    """Parenthesized ` (py_layer_tier.compression["x"])` for an error message, or "" when unknown."""
    return " (%s)" % attr_path if attr_path else ""
```

`where` was the argument name in `make_codec` and `parse_compression` too, so it's `attr_path` in both now, with the Args entries saying what it's for (the attribute a bad value came from, named in the error) rather than just "context".

Output is unchanged, so the error-message tests pass untouched.

---
_Generated by [Claude Code](https://claude.ai/code)_

jbedard:
@zbarsky-openai does this meet your requirements? Please review

xangcastle (approved):

jbedard:
@tamird does this meet your requirements? Please review

tamird (commented):
In response to [the review request](https://github.com/aspect-build/rules_py/pull/1488#issuecomment-5449504681), I reviewed all 22 changed files at `949516f428f34ca5f28407a9a5a0365dbc306947` against integration base `1e35782c12eaff17028530d6f1d3f225677c169a`. Three issues are noted inline. These are source-traced findings; I did not run local builds or tests.

— Codex, reviewing on behalf of @tamird.

py/private/compression.bzl
```diff
@@ -0,0 +1,234 @@
...
+        "extension": "str — output file extension, e.g. '.tar.br'.",
+        "files_to_run": "FilesToRunProvider — the program plus its runfiles, for the tar action's `tools`.",
+    },
+)
+
+# flag: selects the write filter. filter: the `--options` module accepting
+# `compression-level`, None when the filter has no level knob.
+ALGORITHMS = {
+    "bzip2": struct(flag = "--bzip2", filter = "bzip2", ext = ".tar.bz2", min_level = 1, max_level = 9),
+    "compress": struct(flag = "--compress", filter = None, ext = ".tar.Z", min_level = 0, max_level = 0),
```
tamird: **[P2] Keep advertised OCI codecs compatible with the image consumer**

For example, `group_compression = {"default": ["bzip2"]}` now produces bzip2 bytes, but the repository's pinned `rules_oci` 2.2.7 has no bzip2 layer handling: its [descriptor generator](https://github.com/bazel-contrib/rules_oci/blob/bf3c073974562c8a8ab3ed2dae407b065b78a8dc/oci/private/descriptor.sh#L19) leaves these bytes with an empty compression value and the compressed digest as `diffid`, and [image.sh](https://github.com/bazel-contrib/rules_oci/blob/bf3c073974562c8a8ab3ed2dae407b065b78a8dc/oci/private/image.sh#L88) labels them as uncompressed tar. A successful layer build therefore produces an invalid OCI layer. The documented Brotli example has the same problem; changing the extension does not add consumer support. [OCI defines tar, gzip and zstd layer formats](https://github.com/opencontainers/image-spec/blob/147f9c13cedb47a0c4d9a11a222961073d585877/layer.md#L7).

Please keep OCI output to those supported formats (custom gzip/zstd implementations remain useful), or make non-OCI output a separate explicit contract. Add an OCI-image build/extraction regression: the new tests only check archive bytes, and the custom fixture actually emits gzip.

py/private/compression.bzl
```diff
@@ -0,0 +1,234 @@
...
+def codec_tar_args(codec, tar_args):
+    """Append the codec's filter selection to a bsdtar `--create` command line.
+
+    Args:
+        codec: A codec struct from `make_codec` or `codec_from_compressor`.
+        tar_args: The `ctx.actions.args()` object being built for bsdtar.
+    """
+    if codec.program != None:
+        tar_args.add("--use-compress-program")
+        tar_args.add(codec.program, format = codec.program_format)
```
tamird, line 169: **[P2] Preserve runfiles discovery when launching a custom compressor**

A Bazel `cc_binary` compressor using `Runfiles::Create(argv[0])` to read a dictionary from `data` cannot discover its staged runfiles through this invocation. The pinned libarchive 3.8.1 [strips the executable path from `argv[0]`](https://github.com/libarchive/libarchive/blob/9525f90ca4bd14c7b335e2f8c84a4607b0af6bdf/libarchive/archive_cmdline.c#L126) and [passes that basename argv to the child](https://github.com/libarchive/libarchive/blob/9525f90ca4bd14c7b335e2f8c84a4607b0af6bdf/libarchive/filter_fork_posix.c#L139). The tar action supplies no `RUNFILES_*` environment, so the [standard C++ runfiles fallback](https://github.com/bazelbuild/rules_cc/blob/6fd317b2ae0534a29db7085605b0262849e62f93/cc/runfiles/runfiles.cc#L443) looks for `<basename>.runfiles` in the execroot, not beside the executable under `bazel-out/...`. Adding `FilesToRunProvider` to `tools` stages the files but does not repair this lookup.

Please invoke the compressor through a launcher that preserves its executable/runfiles context, and test a tool that actually reads a runfile. The shell fixture has no data dependency and cannot catch this. This finding is source-traced, not a runtime reproduction.

py/private/py_image_layer.bzl
```diff
@@ -1364,6 +1447,8 @@ def py_image_layer(
         groups = {},
         group_execution_requirements = {},
         group_compress_levels = {},
+        group_compression = {},
+        group_compressors = {},
```
tamird, line 1471: **[P2] Preserve the existing positional macro arguments**

Inserting these parameters before the existing thresholds changes valid calls such as `py_image_layer("layers", ":app", {}, {}, {}, 200, 90)`. At the base, the final two arguments set `warn_remote_cache_threshold_mb` and `warn_layer_count`; here they populate `group_compression` and `group_compressors`, and the generated rule fails because those attributes require dictionaries. Positional values for the later platform/tier arguments shift as well.

Please append the new parameters after the existing parameters instead, and retain a positional-call compatibility test. This is a source-verified public-signature regression; I have not identified a downstream positional caller.

py/private/compression.bzl
thesayyn: runfiles is out of scope for this feature. We'll add support for that if somebody needs it.

py/private/py_image_layer.bzl
thesayyn: py_image_layer was never meant to be called without argument names. Will not do.

py/private/compression.bzl
thesayyn: Confirmed and fixed in `7d0641c`. I read the pinned `rules_oci` 2.2.7 sources and it's exactly as you describe — `descriptor.sh` sniffs two magics and nothing else:

```bash
if [[ $(... --read-bytes 2 "$archive") == " 1f 8b" ]]; then
    compression="gzip"; diffid=$(zstd --decompress --format=gzip <"$archive" | regctl digest)
elif zstd -t <"$archive" 2>/dev/null; then
    compression="zstd"; diffid=$(zstd --decompress --format=zstd <"$archive" | regctl digest)
fi
```

so a bzip2/xz/lzma/lz4/lrzip/lzop/`.Z` layer keeps `compression=` empty and `diffid=$digest`, and `image.sh` then emits `application/vnd.oci.image.layer.v1.tar` with the compressed digest in `rootfs.diff_ids`. Seven of the ten filters I added landed there.

I took your second option rather than dropping the filters, since the full libarchive set was the point of the change:

- `ALGORITHMS` now carries an `oci` bit (`none`, `gzip`, `zstd` — matching the spec).
- `py_layer_tier` and `py_image_layer` reject a non-OCI codec at analysis, naming the valid set, unless `allow_non_oci_layers = True` says the tars aren't headed for an image. So non-OCI output is now an explicit contract instead of a silent one.
- A `py_layer_compressor` is judged by its declared `extension`, since its bytes are opaque — `.tar`/`.tar.gz`/`.tar.zst` pass, anything else needs the same opt-in. The README's Brotli example was exactly the trap you point at; it's now a `pigz` example emitting `.tar.gz`, with the rule doc saying the extension decides OCI validity.

On the tests, both of your points were right:

- **The custom fixture emitted gzip.** It now writes its own container — a `LOLZ` magic followed by a gzip stream — so the bytes are provably not something bsdtar could have produced, and `assert_compression.py` strips the magic and unwraps it rather than stopping at a magic check.
- **No OCI-level coverage.** Added `oci_codec_media_type_test` in `e2e/cases/oci/py_image_layer`: it builds a real `oci_image` over a zstd first-party layer, an uncompressed interpreter layer and a gzip source layer, then walks the produced layout — `index.json` → manifest → config — and asserts every layer's `mediaType` is an OCI layer format *and* that each compressed layer's `diff_id` differs from its `digest`. That second assertion is the one that catches this class of bug: with bzip2 the two are equal, because rules_oci never decompressed it.

Five new analysis tests cover the guard itself (non-OCI algorithm, non-OCI compressor extension, the opt-in, and gzip/zstd/none passing untouched); 24 pass locally.

One caveat worth stating plainly: I could not execute the `e2e/cases` suite in my sandbox — its `container_structure_test` download is blocked by this environment's egress policy — so `oci_codec_media_type_test` is unrun and CI is its first execution. Everything under `py/tests` I ran, including forcing a re-run past the cache.

---
_Generated by [Claude Code](https://claude.ai/code)_

thesayyn: Follow-up on the regression test — the first version of it was wrong twice, and CI plus a negative test caught both. Now at `c378fe7`.

**1. Over-specified count.** I asserted exactly one gzip layer; CI reported three. The image has `base = "@ubuntu"`, so the manifest also carries the base's own gzip layers alongside the squashed pip layer. The counts are minimums now, since how many layers the base contributes isn't this rule's business.

**2. The test did not catch the bug it exists for.** More serious. I ran it against a synthetic layout shaped like the bzip2 failure and it passed:

```
bzip2-as-tar   -> (0, 'ok: all 2 layers are OCI formats ...')
```

Because from the manifest alone, a mislabelled bzip2 layer is *indistinguishable* from a genuine uncompressed one: rules_oci writes `mediaType: ...tar` and leaves `diff_id == digest`, which is exactly what an honest uncompressed layer looks like. The manifest is self-consistent and simply wrong about the bytes — so metadata checks can't see it. The test now reads each blob and asserts the bytes match the declared format (`ustar` at offset 257 for `...tar`, `1f 8b` for `+gzip`, `28 b5 2f fd` for `+zstd`):

```
healthy image  -> (0, 'ok: all 3 layers carry the bytes their mediaType claims, with matching diff_ids')
bzip2-as-tar   -> (1, "FAIL: layer sha256:8fdc... is labelled an uncompressed tar but its bytes are not —
                       expected b'ustar' at offset 257, got b'\x00\x00\x00\x00\x00'.")
stale diff_id  -> (1, 'FAIL: layer sha256:4749... is ...tar+gzip but its diff_id equals the compressed digest')
```

Those three cases are the synthetic-layout runs I used to check the assertions bite; the real `oci_codec_media_type_test` still runs only in CI, for the sandbox reason above.

---
_Generated by [Claude Code](https://claude.ai/code)_

py/defs.bzl
```diff
@@ -63,7 +68,9 @@ py_unpacked_wheel = _py_unpacked_wheel
 
 py_image_layer = _py_image_layer
 py_layer_tier = _py_layer_tier
+py_layer_compressor = _py_layer_compressor
```
tamird, line 71: **[P2] Regenerate the Python API reference for the new public API**

Current `main` added the Stardoc reference and freshness test in [#1511](https://github.com/aspect-build/rules_py/pull/1511). These new exports, the compressor provider, and the changed layer attributes need corresponding generated documentation. Integration CI for this head now fails [`//docs:update_docs_0_test`](https://github.com/aspect-build/rules_py/actions/runs/33436683192/job/99634883673) because `docs/api/py.md` differs from the generated `docs/py_api.md` (the other 337 root Bazel 8 tests pass).

Please update the branch onto current `main`, run `bazel run //docs:update_docs`, and commit the generated API reference. The README additions do not update this checked reference.

tamird (changes requested):
Re-reviewed all 23 changed files at `c378fe7204df6a646a07f59c14d548eb4c81db29` against merge base `1e35782c12eaff17028530d6f1d3f225677c169a`, following [Jason's request](https://github.com/aspect-build/rules_py/pull/1488#issuecomment-5449504681). One required integration fix is noted inline: update the generated Python API reference.

The explicit non-OCI opt-in and the regression test that checks an actual OCI image's layer bytes address the [earlier format-compatibility finding](https://github.com/aspect-build/rules_py/pull/1488#discussion_r3879953626). I found no additional functional issue. [Runfile-dependent compressors](https://github.com/aspect-build/rules_py/pull/1488#discussion_r3897984785) and [positional-call compatibility](https://github.com/aspect-build/rules_py/pull/1488#discussion_r3897990885) remain outside the scope described in the author's replies; I have not validated a downstream production workload.

All 24 `//py/tests/image-layer-compression:all` tests passed locally on macOS with Bazel 8.6.0. The [Bazel 8](https://github.com/aspect-build/rules_py/actions/runs/33436683192/job/99634883761) and [Bazel 9](https://github.com/aspect-build/rules_py/actions/runs/33436683192/job/99634883962) OCI/e2e CI jobs passed on this head. Root Bazel 8 CI fails only the generated-docs check; root Bazel 9 and the CLA check are still pending.

— Codex, reviewing on behalf of @tamird.

py/defs.bzl
thesayyn: Right on both counts — done in `c7c5691`.

Rebased onto current `main` (`de8af46`, which carries #1511) and ran `bazel run //docs:update_docs`; `docs/api/py.md` is the generated output, not hand-edited. It picks up:

- `py_layer_compressor` (rule + attribute table) and the `PyLayerCompressorInfo` provider,
- `py_layer_tier`'s new `compression` text, `compressors` and `allow_non_oci_layers`,
- `py_image_layer`'s `group_compression`, `group_compressors` and `allow_non_oci_layers`, plus the reworded `group_compress_levels`,
- the two added `PyLayerTierInfo` fields (`compressors`, `codecs`).

`//docs:update_docs_0_test` and `_1_test` both pass locally now, as do the 24 tests in `//py/tests/image-layer-compression` on the rebased tree.

Worth noting for anyone else hitting this: generating the docs needs a full analysis of `//py:defs`, so the diff test can only go stale silently on a branch that predates #1511 — which is what happened here.

---
_Generated by [Claude Code](https://claude.ai/code)_

tamird (approved):
**Correction:** This approval did not validate our monorepo requirements. See the [downstream validation follow-up](https://github.com/aspect-build/rules_py/pull/1488#pullrequestreview-5071797937); adoption and performance remain unvalidated.

---

LGTM at `c7c5691a227b3c8edf71d2e03d7a266b6d48b164`. Re-reviewed the complete change against `de8af46f8d95cfc2e6c3834aaeb23de0b5d6d164`, including the rebase's interaction with layer assembly. The compression changes are unchanged from the previous review, and I found no new issues in codec selection, custom-compressor wiring, the OCI guard, or regression coverage.

Bazel 8 and 9 root and OCI/e2e suites passed in [CI on this revision](https://github.com/aspect-build/rules_py/actions/runs/33438684777). I did not rerun local tests for this revision.

— Codex, reviewing on behalf of @tamird.

tamird (commented):
Correction to my earlier approval: I checked the upstream implementation but had not validated the downstream requirements behind Jason's request. That approval should not be treated as confirmation that this works in our monorepo.

Our current policy is zstd level 9 by default, with level 3 for frequently rebuilt source groups. This API supports explicit zstd settings for named groups, but retains the gzip level 6 fallback. Our existing compression patch also contains native/data layer-ownership changes, so this PR alone does not establish that the patch can be removed.

I have now run our existing compression/content/grouping integration test against the current pinned and patched dependency: it passed. The attempt against `c7c5691`, with compatibility patches and unchanged PR compression code, stopped before the test could run: the root requires `hermetic_launcher` 0.0.13 but the candidate resolves 0.0.14. The local test-only-package policy patch also needs rebasing. These are integration prerequisites, not defects demonstrated in this PR.

No new upstream blocker is established. Downstream adoption, patch retirement, and the expected service-image build performance remain unvalidated; I cannot yet answer the downstream compatibility question affirmatively.

Please dismiss my earlier approval while that downstream validation remains outstanding.

— Codex, reviewing on behalf of @tamird.

PR #1493, opened 2026-08-27
perf(py): native //py:python launcher via hermetic_launcher

xangcastle:
`//py:python` aliased rules_python's interpreter_binary, keeping a rules_python dependency alive just for a runfiles-safe way to exec the toolchain interpreter in actions. Reimplement it as a py_interpreter_launcher rule backed by hermetic_launcher (already a dependency, used by py_venv_exec): a small native binary that resolves the toolchain interpreter through the runfiles mechanism and execs it, forwarding extra arguments. Behavior is unchanged for consumers, and as a side effect the launcher now also works on Windows, where rules_python's shell-script wrapper is unsupported. 

### Test plan

### DIFF
--- a/py/BUILD.bazel
+++ b/py/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_lib//:bzl_library.bzl", "bzl_library")
+load("//py/private/interpreter:launcher.bzl", "py_interpreter_launcher")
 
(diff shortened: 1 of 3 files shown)


xangcastle:
@codex

jbedard (approved):

PR #1495, opened 2026-08-27
refactor(uv): replace pure-Starlark sha1 with fnv1a for repo name suffixes

xangcastle:
Drop the hand-rolled pure-Starlark SHA-1 implementation, which only existed to mint stable suffixes for wheel/sdist repository names. Since those names are truncated to 16 hex chars anyway and never serve a security or integrity purpose (the lockfile's own hash field remains the primary identifier), a 64-bit FNV-1a offers the same collision space at a fraction of the evaluation cost. Git-sourced sdists now derive their suffix from the pinned commit in the lockfile instead of hashing the URL, which also makes the resulting repo names meaningful when debugging. This deletes sha1.bzl along with its BUILD target and a stale dependency in uv_project that referenced it without using it.

### Test plan


xangcastle:
@codex

jbedard (approved):

PR #1496, opened 2026-08-28
refactor(uv): expand templates instead of .format() strings in uv_hub

xangcastle:
The hub repository rule was assembling its BUILD and .bzl files out of .format() strings, which meant doubling every brace in the generated Starlark and burying the skeletons inside the implementation.

This moves the static parts into five .tmpl files under uv/private/uv_hub/templates and expands them with repository_ctx.template, so the rule now only computes the dynamic bits: select arms, the group inversion, label lists.

Net effect is about ninety fewer lines, and the templates read as plain Starlark. Output is byte-for-byte identical, regenerating the snapshots produced an empty diff. Docstrings got a pass while I was in there: _hub_impl now documents the five artifacts and their invariants.

---

### Test plan


xangcastle:
@codex

uv/private/uv_hub/templates/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+package(default_visibility = ["//uv/private:__subpackages__"])
+
+exports_files([
+    "defs.tmpl",
```
jbedard, line 4: I think these should be named `*.tmpl.{bazel,bzl}` so we actually know what type of file they are, and an IDE will render them correctly etc

xangcastle: if I add .bzl at the end, gazelle will try to generate bzl_libraries

jbedard: We could tell gazelle to ignore `*.tmpl.*` maybe?

xangcastle: ok, let me try

PR #1503, opened 2026-08-28
fix(uv): preserve transitive dependency markers

finn-ball:
`uv lock` can omit a transitive edge marker when the condition is already guaranteed by the parent dependency. Preserve the markers that made each package reachable so platform-specific transitive versions do not become unconditional.

Propagation uses canonical minimal marker clauses to remain cycle-safe, and includes regressions for platform-split transitive versions and conditional dependency cycles.


xangcastle:
@codex

jbedard:
@codex review

jbedard (approved):

xangcastle (approved):

PR #1507, opened 2026-08-29
ci(uv): run cross-built wheels on native-arch runners

xangcastle:
CI and e2e only: no change under `uv/`; the public API is unchanged.

**Native-arch verification** (`.github/workflows/ci-workflows.yaml`): each crossbuild case exports `<case>_{amd64,arm64}_bundle` tarballs (wheel plus its standalone `test_<case>.py`, `e2e/crossbuild/tools/wheel_export.bzl`). Three pipelines cross-build on one runner, upload the bundles, and `pip install` + run them on a native runner of the target architecture, with no containers and no emulation in the verdict:

- darwin → linux/amd64 + linux/arm64 (gated to `main` and `*macos*` branches),
- ubuntu/amd64 → ubuntu/arm64,
- ubuntu/arm64 → ubuntu/amd64.

`test-all` gates on the two Linux pipelines and accepts `skipped` for the macOS one outside the gated branches. The bundle-building steps pass `--remote_download_outputs=toplevel`: `--config=ci` builds without the bytes, and with a warm cache the tarballs this job uploads were not materialized (the first run of this branch passed only because its cache was cold).

**New cases**, all built for linux/amd64 and linux/arm64 with `collect_wheels` (Tag metadata, ELF arch of every bundled `.so`) plus a native import test:

- `pycross-geohash`: C++ extension with no upstream wheel. On a macOS host, `test.sh` also cross-builds it for macOS amd64 (a manual `collect_wheels` with `check_elf = False`; the platform transition always resolves to os:macos, so `target_compatible_with` cannot tell hosts apart).
- `pycross-msgpack`: Cython-generated C extension via setuptools.
- `pycross-numpy`: meson-python with `config_settings = {"setup-args": ["-Dblas=none", "-Dlapack=none"]}` (#1524); the hermetic sandbox has no BLAS.
- `pycross-psutil`: plain setuptools C extension.
- `pycross-compile_pyc`: regression guard for cross-arch `whl_install` pyc precompilation.

The Rust cases stay out of this PR on purpose: `pycross-rust` keeps building as on `main` and gets no export, so the pipelines never depend on the Rust toolchain work in #1526. Once that lands, the Rust cases join the native runs with a one-line `pycross_wheel_export` each.

The macOS `smoke` job is removed: the root and `e2e/cases` suites already run on Linux for every PR, and the crossbuild cases are exercised by the darwin cross-build pipeline above, so it tripled macOS minutes without covering anything new.

`collect_wheels` gains `tags` and `check_elf` for the macOS collection; `pycross-setuptools` shares its setproctitle wheel label between `collect_wheels` and the export instead of repeating it.

---

### Test plan

- New test cases added
- `cd e2e/crossbuild && bazel test //... && bash test.sh` passes locally.
- CI: the three build/run pipelines plus `test-all`.


xangcastle:
@codex

e2e/crossbuild/pycross-bcrypt/setup.MODULE.bazel
```diff
@@ -0,0 +1,38 @@
...
+    hub_name = "pypi_pycross_bcrypt",
+    lock = "//pycross-bcrypt:uv.lock",
+    pyproject = "//pycross-bcrypt:pyproject.toml",
+)
+uv.override_package(
+    name = "bcrypt",
+    env = {
+        "CARGO": "$(CARGO)",
+        "RUSTC": "$(RUSTC)",
+        "RULES_PY_RUST_HOST_SYSROOT": "$(RUST_HOST_SYSROOT)",
```
jbedard: I thought `RULES_PY_RUST_HOST_SYSROOT` was a secret internal var?

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -1015,6 +1015,36 @@ def _build_backend(pyproject_data: dict[str, object] | None) -> str | None:
     return backend if isinstance(backend, str) else None
 
 
+_SETUPTOOLS_BACKENDS = (
```
jbedard: This seems like a lot more then just a CI commit if we are doing this stuff?

.github/workflows/ci-workflows.yaml
```diff
@@ -153,19 +153,215 @@ jobs:
...
+                    tar -xzf "${bundle}" -C "${tmp}"
+                    (cd "${tmp}" && pip install *.whl && python "test_${case_name}.py") || status=1
+                    rm -rf "${tmp}"
+                    echo "::endgroup::"
+                done
+                exit "${status}"
+
+    crossbuild-wheels-amd64:
+        name: build wheels on amd64
+        runs-on: ubuntu-22.04-32core
```
jbedard, line 195: Do we need the 32 cores? 😅

.github/workflows/ci-workflows.yaml
```diff
@@ -153,19 +153,215 @@ jobs:
...
-                fi
-                if [ "${{ needs.typecheck.result }}" != "success" ]; then
-                    echo "Typecheck did not succeed: ${{ needs.typecheck.result }}"
-                    exit 1
-                fi
+                for result in \
+                    "typecheck=${{ needs.typecheck.result }}" \
+                    "test=${{ needs.test.result }}" \
+                    "crossbuild-run-amd64-built=${{ needs['crossbuild-run-amd64-built'].result }}" \
+                    "crossbuild-run-arm64-built=${{ needs['crossbuild-run-arm64-built'].result }}"; do
```
jbedard: Can't we do something such as:
```
for name in [...]:
   result = needs[name].result
```
?

.github/workflows/ci-workflows.yaml
```diff
@@ -153,19 +153,215 @@ jobs:
...
+                    if [ "${result#*=}" != "success" ]; then
+                        echo "Required job did not succeed: ${result}"
+                        exit 1
+                    fi
+                done
+                # The darwin pipeline is gated to main and '*macos*' branches:
+                # skipped is expected elsewhere; anything that ran must pass.
+                case "${{ needs['crossbuild-run-darwin-built'].result }}" in
+                    success|skipped) ;;
```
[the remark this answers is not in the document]
xangcastle: Fixed in bb0b67ab

.github/workflows/ci-workflows.yaml
xangcastle: Done in bb0b67ab: the step iterates over toJSON(needs) with jq, so every job in needs is checked without listing it

jbedard (approved):

PR #1512, opened 2026-08-31
feat(py): add per-target free-threaded interpreter selection

thorvald-oai:
## What

Allow `py_binary`, `py_test`, and `py_venv` to select free threading per target with `freethreaded = True|False`. The default `None` inherits the Aspect free-threading flag.

Keep rules_python's native-extension ABI setting synchronized within the selected Python dependency graph. Runtime `data` edges restore both original caller settings independently, including across nested terminal overrides.

Extend the existing transition and exposed-venv tests and document the attribute. Existing interpreter-version support and default runtime selection are preserved.

## Why

Mixed GIL-enabled and free-threaded targets need consistent interpreter and native-extension settings without changing a build-wide flag. Preserving the original settings on runtime data also lets shared data reuse one configuration across those targets.

Before this change, setting only Aspect's `freethreaded` flag to `True` leaves `rules_python`'s `py_freethreaded` at its default `"no"`. A binary can therefore use a free-threaded interpreter while its nanobind dependency is compiled without `NB_FREE_THREADED`. With `freethreaded = True` on the binary, this PR sets both flags together so the interpreter and native extension use the same GIL mode.

### DIFF
--- a/docs/api/py.md
+++ b/docs/api/py.md
@@ -693,7 +693,7 @@ ORed, `*` is fnmatch). Unknown args are rejected rather than ignored.
 <pre>
 load("@aspect_rules_py//py:defs.bzl", "py_venv")
 
-py_venv(<a href="#py_venv-kwargs">**kwargs</a>)
+py_venv(<a href="#py_venv-freethreaded">freethreaded</a>, <a href="#py_venv-kwargs">**kwargs</a>)
 </pre>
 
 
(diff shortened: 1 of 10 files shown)


jbedard:
@codex

jbedard (approved):

jbedard:
@codex

jbedard:
@codex

PR #1519, opened 2026-09-02
feat(uv)!: anchor post-install patches at site-packages, default strip to 1

jbedard:
Post-install patches were applied from the install prefix, so every patch header had to spell out `lib/python3.X/site-packages/` and broke on a toolchain bump. Apply them from site-packages instead, matching the `exclude_glob` convention. Prefix data files and `bin/` are no longer patch-reachable; the data-file guard already forbade changing that set.

Both `pre_build_patch_strip` and `post_install_patch_strip` now default to 1, matching `git diff` output, so the common case needs no strip attr. The "strip requires patches" checks compare against the default instead of truthiness.

BREAKING CHANGE: existing post-install patches must drop the `lib/pythonX.Y/site-packages/` prefix from their headers or set `post_install_patch_strip = 4`. Callers relying on an implicit strip of 0 must set it explicitly.

### Test plan

- New test cases added


jbedard:
@codex

jbedard:
@codex

jbedard:
@codex

jbedard:
@tamird and @zbarsky-openai please review this breaking API change. Ideally this should simplify patches within your repo, please report any complications

xangcastle (approved):

tamird (commented):
Reviewed against our monorepo patch consumers. No blocking source finding; anchoring at site-packages removes real Python-minor duplication for us.

Adoption needs a coordinated migration: all 19 current post-install overrides explicitly use strip=1 and their patch headers contain, or generate, lib/pythonX.Y/site-packages. Those callers will fail unchanged with the new working directory. We can make the headers site-packages-relative and delete the minor-specific header generation, or temporarily use strip=4. The three pre-build overrides already use strip=1 and remain sdist-relative.

For generated patches, only the header-path rewriting becomes unnecessary. In particular, our pytest-postgresql patch still needs its hermetic executable/runfile substitutions. I found no current post-install hunk that requires patching prefix bin/share files.

This is source-level compatibility review. Our pinned revision/local patch stack still needs integration work; no monorepo build or performance result is claimed.

PR #1520, opened 2026-09-03
fix(py): capture flag baselines only when a terminal changes a flag

jbedard:
Previously every `py_venv` transition wrote the four baseline shadow flags, so the `deps` of a terminal with no `python_version`, `freethreaded` or `dep_group` override landed in a configuration distinct from the caller's. Any cc/proto library shared between a Python terminal and a non-Python one was analyzed and built twice.

A baseline now records the inherited value only when the transition actually changes its flag. Terminals that keep every inherited value are no-ops and share the caller's configuration; overrides equal to the inherited value are no-ops too.

This may have been a regression introduced by https://github.com/aspect-build/rules_py/pull/1294

Python terminals with no `python_version`/`freethreaded`/`dep_group` override no longer fork their deps into a second configuration; baselines are captured only when a flag actually changes.

### Test plan

### DIFF
--- a/py/private/transitions.bzl
+++ b/py/private/transitions.bzl
@@ -1,10 +1,10 @@
 """Common transition implementation used by the various terminals."""
 
-DEP_GROUP_FLAG = "@aspect_rules_py//uv/private/constraints/dep_group:dep_group"
+_DEP_GROUP_FLAG = "@aspect_rules_py//uv/private/constraints/dep_group:dep_group"
 _DEP_GROUP_BASELINE_FLAG = "@aspect_rules_py//uv/private/constraints/dep_group:baseline"
 
 # Our own python_version flag, replacing the rules_python one.
-PYTHON_VERSION_FLAG = "@aspect_rules_py//py/private/interpreter:python_version"
+_PYTHON_VERSION_FLAG = "@aspect_rules_py//py/private/interpreter:python_version"
 _PYTHON_VERSION_BASELINE_FLAG = "@aspect_rules_py//py/private/interpreter:baseline_python_version"
 
 # rules_python's flag, kept for backward compatibility during migration.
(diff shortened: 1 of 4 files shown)


xangcastle (approved):

jbedard:
@tamird please review. This fixes a regression that a robot claims https://github.com/aspect-build/rules_py/pull/1294 caused where more transitions are created then necessary. I want to ensure this does not regress or cause new issues for you, especially related to #1294

py/private/transitions.bzl
```diff
@@ -61,84 +54,49 @@ def _freethreaded_available(version):
...
-    acc[_RPY_VERSION_BASELINE_FLAG] = _baseline(
-        settings,
-        _RPY_VERSION_BASELINE_FLAG,
-        settings[_RPY_VERSION_FLAG],
-    )
-    if attr.python_version:
-        version = str(attr.python_version)
+    if attr.freethreaded:
+        freethreaded = attr.freethreaded == "true"
     else:
```
tamird, line 60: [P1] Keep freethreaded optional for direct transition callers

Our monorepo has an existing custom rule that applies python_transition to a binary edge and declares python_version and dep_group, but no freethreaded attribute. Previously getattr(attr, "freethreaded", "") let that caller inherit the mode; this unconditional field access now fails during transition evaluation, before its dependencies can be analyzed.

Please retain the optional getattr lookup here. The existing reset-data-edges terminal fixture always declares freethreaded, so omitting its value does not cover a caller that lacks the attribute itself; that case belongs in the existing analysis coverage.

tamird (commented):
Monorepo-aware review: capturing only changed flags preserves the #1294 contract—nested terminals retain the first inherited baseline, and data edges restore the original settings and clear the shadows. Our aligned default flags should also benefit from eliminating the baseline-only configuration split.

There is one downstream compatibility regression in the inline finding: a direct transition caller without a freethreaded attribute now fails. This is a source review against our actual callers; I have not run a monorepo integration or performance check.

py/private/transitions.bzl
jbedard: https://github.com/aspect-build/rules_py/pull/1531

tamird: Verified merged #1531 (b85a957f): it restores the optional getattr lookup, and the new fixture actually omits the freethreaded attribute while checking inherited mode and shared data reset. That addresses the compatibility case reported here. Monorepo upgrade execution remains separate from this source verification.

PR #1524, opened 2026-09-03
feat(uv): config_settings on uv.override_package for PEP 517 backends

xangcastle:
Adds `config_settings` to `uv.override_package`, the PEP 517 way to pass backend options to a source build:

```starlark
uv.override_package(
    lock = "//:uv.lock",
    name = "numpy",
    config_settings = {"setup-args": ["-Dblas=none", "-Dlapack=none"]},
)
```

**Flow.** `sdist_build` renders the attribute into the generated `pep517_whl` / `pep517_native_whl` call; the rules pass one `--config-setting key=value` per listed value (keys sorted, so the action key is stable); `build_helper.py` hands them to `python -m build` as `-C key=value`. `build` folds repeated keys into a list, so package settings cannot collide with the `-C` the cross branch adds for meson/CMake. The `setup.py bdist_wheel` fallback path rejects them with an explicit error, since no frontend is involved there.

The keys are backend-specific (`setup-args` for meson-python, `cmake.define.<VAR>` for scikit-build-core, `--build-option` for setuptools), so this is one generic attribute instead of one environment variable per backend. Applies to pure and native source builds.

**Generated BUILD escaping.** `config_settings`, `env` and `toolchains` render through `repr()`: keys and values are free-form strings owned by the user or the backend, so a quote or backslash in one (a `CFLAGS` with `-DMSG="hi"`) must land in the external repo's `BUILD.bazel` as a valid Starlark literal. `env` previously interpolated raw.

**Tests.**
- `config_settings_{pure,native}_args_test`: analysis tests pin the exact argv on both rules; `config_settings_absent_test` pins that an unset attribute adds nothing.
- `config_settings_test`: a `build_test` drives the in-tree PEP 517 backend fixture, which compares the `config_settings` it receives against `EXPECTED_CONFIG_SETTINGS` (a single value arrives as `str`, repeated keys as `list`).
- `config_settings_attr_test`, `env_attr_test`: unit tests for the BUILD template rendering (sorted keys, Starlark literals with escaping, empty when unset).
- Existing `sdist_build` snapshots are unchanged.

Docs: new "Backend config settings" section in `docs/uv.md`.

### DIFF
--- a/uv/private/extension/defs.bzl
+++ b/uv/private/extension/defs.bzl
@@ -274,6 +274,7 @@ def _parse_projects(module_ctx, hub_specs):
                 override.extra_data or
                 override.toolchains or
                 override.env or
+                override.config_settings or
                 override.monitor_memory or
                 override.resource_set != "default"
             )
@@ -524,6 +525,7 @@ def _parse_projects(module_ctx, hub_specs):
                         console_scripts = sbuild_console_scripts,
                         resource_set = pkg_override.resource_set,
                         env = pkg_override.env,
+                        config_settings = pkg_override.config_settings,
                         error = "uv.override_package() for '{}=={}' in lock '{}': build-only attributes require a source distribution, but the lock record has only wheels: {{}}".format(
                             package["name"],
                             package["version"],
@@ -581,11 +583,13 @@ def _parse_projects(module_ctx, hub_specs):
                     # they don't replace them. Empty == no augmentation.
                     extra_toolchains = []
                     extra_env = {}
+                    config_settings = {}
                     monitor_memory = False
                     resource_set = "default"
                     if pkg_override:
                         extra_toolchains = [str(t) for t in pkg_override.toolchains]
                         extra_env = pkg_override.env
+                        config_settings = pkg_override.config_settings
                         monitor_memory = pkg_override.monitor_memory
                         resource_set = pkg_override.resource_set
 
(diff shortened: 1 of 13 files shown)


xangcastle:
@codex

uv/private/extension/defs.bzl
```diff
@@ -1016,6 +1023,10 @@ _override_package_tag = tag_class(
             default = {},
             doc = "Extra environment variables merged into the build action's `env` dict. Values may reference $(VAR) make-variables sourced from extra `toolchains` listed above. Prefix an execroot-relative path with `$(EXECROOT)/` so it remains valid after the backend changes into the unpacked source tree. Omit CC/CXX/AR/LD/STRIP to use the configured C++ action tools.",
         ),
+        "config_settings": attr.string_list_dict(
+            default = {},
+            doc = "PEP 517 config settings handed to the build backend through the frontend's `-C key=value`, one flag per listed value; a key with several values reaches the backend as a list. Backend-specific by nature, e.g. `{\"setup-args\": [\"-Dblas=none\"]}` for meson-python or `{\"cmake.define.FOO\": [\"1\"]}` for scikit-build-core. Applies to pure and native source builds.",
```
jbedard: Does the user of this rule care about the `-C` part? Is that public or just our internal wiring?

jbedard (approved):

uv/private/extension/defs.bzl
xangcastle: Internal wiring, and the doc was leaking it. I documented that in my last commit

PR #1525, opened 2026-09-03
feat(uv): map well-known toolchain make-variables into the sdist build env

xangcastle:
`pep517_native_whl` reads the make-variables of the toolchains listed on it and sets the env keys `build_helper.py` consumes, so listing a toolchain is enough:

```starlark
uv.override_package(
    lock = "//:uv.lock",
    name = "jpype1",
    toolchains = ["@bazel_tools//tools/jdk:current_java_runtime"],
)
```

| make-variable | env key | source |
|---|---|---|
| `CARGO`, `RUSTC` | `CARGO`, `RUSTC` | rules_rust `current_rust_toolchain` |
| `RUST_HOST_SYSROOT` | `RULES_PY_RUST_HOST_SYSROOT` | exec-configured rust sysroot layer |
| `JAVA`, `JAVABASE` | `JAVA`, `JAVA_HOME` | Java runtime |
| `ANT_HOME`, `ANT_BIN_DIR` | `ANT_HOME`, `RULES_PY_ANT_BIN_DIR` | Ant layer |

The table mirrors the set `build_helper.py` already absolutizes. Explicit `env` entries win over derived values, and other make-variables still need an explicit `"FOO": "$(FOO)"`.

**Why.** Until now the only way to feed the helper's `RULES_PY_RUST_HOST_SYSROOT` / `RULES_PY_ANT_BIN_DIR` was for users to spell those internal names in `uv.override_package(env = ...)`, next to `"CARGO": "$(CARGO)"`-style boilerplate. The `RULES_PY_*` names go back to being a contract between the rule and the helper.

**Consumers.** `e2e/cases/uv-sdist-jdk-build`, `e2e/crossbuild/pycross-rust` and `e2e/crossbuild/pycross-jdk` drop their `env` dicts and keep only `toolchains`; the jpype1 BUILD snapshot loses the same four lines. Every table entry has one of these as its consumer.

**Tests.**
- `derived_env_test`: `JAVA`/`JAVA_HOME` derive from the Java runtime with no `env`; the make-variable name `JAVABASE` does not leak as an env key.
- `derived_env_explicit_wins_test`: an explicit `JAVA_HOME` overrides the derived value while `JAVA` stays derived.
- `explicit_env` end-to-end fixture (`python_env_test`): drops its `$(JAVA)`/`$(JAVABASE)` lines, so the in-tree PEP 517 backend now checks that the *derived* paths arrive absolute and existing.
- `pycross-rust` and `pycross-jdk` cross-build green locally with derived env only.

Docs: "Build-time toolchains" section in `docs/uv.md`; `toolchains` attr doc names the derived set.

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -379,6 +379,27 @@ time, so the complete source-built wheel still participates in the normal
 An explicit `console_scripts = {}` suppresses all detected scripts, which is
 useful when a pre-build patch removes stale entry-point metadata.
 
+### Build-time toolchains
+
+A native sdist build may need tools beyond the C++ toolchain: a JDK for JNI
+extensions, cargo and rustc for Rust extensions, Ant. List them on the
+package's override and their well-known make-variables reach the build
+environment on their own:
+
+```starlark
+uv.override_package(
+    lock = "//:uv.lock",
+    name = "jpype1",
+    toolchains = ["@bazel_tools//tools/jdk:current_java_runtime"],
+)
+```
+
+`$(JAVA)` and `$(JAVABASE)` arrive as `JAVA` and `JAVA_HOME`; `$(CARGO)`,
+`$(RUSTC)` and `$(RUST_HOST_SYSROOT)` as the variables the Rust build path
+reads; `$(ANT_HOME)` and `$(ANT_BIN_DIR)` likewise. Any other make-variable a
+toolchain exports still needs an explicit `env` entry (`"FOO": "$(FOO)"`),
+and an explicit entry always wins over the derived value.
+
 ## Best practices
 
 **Consolidate your hubs**. In `rules_python`, environments with multiple depsets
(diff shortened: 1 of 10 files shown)


xangcastle:
@codex

xangcastle:
@codex review

docs/uv.md
```diff
@@ -379,6 +379,27 @@ time, so the complete source-built wheel still participates in the normal
...
+```starlark
+uv.override_package(
+    lock = "//:uv.lock",
+    name = "jpype1",
+    toolchains = ["@bazel_tools//tools/jdk:current_java_runtime"],
+)
+```
+
+`$(JAVA)` and `$(JAVABASE)` arrive as `JAVA` and `JAVA_HOME`; `$(CARGO)`,
+`$(RUSTC)` and `$(RUST_HOST_SYSROOT)` as the variables the Rust build path
```
jbedard, line 398: Should give an example of the toolchain that the rust ones come from?

jbedard:
@codex

xangcastle:
@codex

jbedard (approved):

PR #1526, opened 2026-09-03
feat(uv): build Rust sdists with the project's rules_rust or rules_rs toolchain

xangcastle:
Rust sdists (maturin, setuptools-rust) build with the Rust toolchain the project registers. rules_py depends on neither rules_rust nor rules_rs and works with both.

**API**
- `uv.rust_toolchain(toolchain = "@rules_rust//rust/toolchain:current_rust_toolchain", lock = ...)`: module-wide or per lock, the lock wins. sdist_build wires it, plus an exec-configured `rust_host_sysroot` layer, into every sdist whose backend is maturin, or setuptools with setuptools-rust declared in its build requirements. Such an sdist with no toolchain fails at repository generation naming the declaration to add. Stray `.rs` files (zstandard's optional extension) neither wire nor demand one.
- `uv.override_package(cargo_lock = ...)`: a Cargo.lock for an sdist that ships none. Crates are vendored from the lock at repository generation and cargo builds offline; git and path sources are rejected.
- `bazel run @sdist_build__<project>__<pkg>__<ver>//:cargo_lock` writes that lock with the project's toolchain.

**Build**
- Cross: the target toolchain's sysroot merged with the exec platform's rust-std, so build scripts and proc-macros compile against the exec std. cc-rs finds the wired C toolchain under `CC_<triple>`, `CXX_<triple>`, `AR_<triple>` and `RANLIB_<triple>`.
- Reproducible wheels: `--remap-path-prefix` for sandbox and execroot, `-C codegen-units=1` and a host-independent `-C metadata=` for target crates, maturin's SBOM off. `MATURIN_NO_INSTALL_RUST=1` on every build.

**e2e/crossbuild**: pycross-rust (rules_rust), pycross-rust-rs (rules_rs, lock-scoped), pycross-tiktoken (setuptools-rust, `cargo_lock`); their wheels run on native amd64 and arm64 runners. `test.sh` aggregates per-case scripts; pycross-tiktoken's runs `:cargo_lock`. The BUILD every sdist_build repository generates is snapshotted under `snapshots/`, so which sdists get toolchains, the Rust layer or vendored crates is a reviewable diff.

---

New `uv.rust_toolchain()`, `uv.override_package(cargo_lock = ...)` and `:cargo_lock` targets. Rust sdists no longer need a per-package `uv.override_package(toolchains = [...])`.

### Test plan

- Analysis, Starlark and helper unit tests for the wiring, the vendoring, the rustc wrapper and the lock generator.
- `cd e2e/crossbuild && bazel test //... && bash test.sh` passes locally with the sandbox network blocked, cross-building pydantic-core (rules_rust and rules_rs) and tiktoken for linux/amd64 and linux/arm64 from a macOS host.
- The same commit checked out at two paths builds byte-identical cross pydantic-core wheels.
- `docs/uv.md` updated.


xangcastle:
@codex review

uv/private/sdist_build/repository.bzl
```diff
@@ -409,6 +473,10 @@ sdist_build = repository_rule(
         ),
         "pre_build_patches": attr.label_list(default = []),
         "pre_build_patch_strip": attr.int(default = 1),
+        "rust_toolchain": attr.string(
```
jbedard: Why not a label?

uv/private/extension/defs.bzl
```diff
@@ -959,6 +962,13 @@ _project_tag = tag_class(
             mandatory = True,
             doc = "The `uv.lock` pinning this project's dependency graph.",
         ),
+        "rust_toolchain": attr.label(
+            mandatory = False,
+            doc = "A rules_rust `current_rust_toolchain`-style target. When set, every sdist in " +
```
jbedard: What about `rules_rs`? Do we need a test for both?

docs/uv.md
```diff
@@ -400,6 +400,24 @@ reads; `$(ANT_HOME)` and `$(ANT_BIN_DIR)` likewise. Any other make-variable a
 toolchain exports still needs an explicit `env` entry (`"FOO": "$(FOO)"`),
 and an explicit entry always wins over the derived value.
 
+Rust needs one more step, because build scripts and proc-macros always run on
```
jbedard: Isn't that true for all sdist build scripts? They run on the exec platform, but may cross-compile to a target platform?

uv/private/extension/defs.bzl
xangcastle: yes sir, and rules_rs should be by default

e2e/crossbuild/MODULE.bazel
```diff
@@ -14,6 +14,7 @@ bazel_dep(name = "bazel_lib", version = "3.0.0")
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "llvm", version = "0.8.3")
 bazel_dep(name = "rules_rust", version = "0.68.0")
```
jbedard, line 16: Can we drop this?

jbedard: No, because we're testing both within this MODULE, right?

xangcastle: on it

xangcastle: yes confirmed, we need those toolchains

e2e/crossbuild/pycross-rust-rs/setup.MODULE.bazel
```diff
@@ -3,14 +3,19 @@
...
+it fetches, so the lock-scoped `uv.package_toolchains()` below points at that
+repository's `current_rust_toolchain`, overriding the module-wide rules_rust
+one for this project only: the wiring is a provider contract, not a dependency
 on one ruleset's toolchain layout. The two rulesets coexist here because each
 carries its own `rules_rust` repository and therefore its own toolchain type.
 """
 
 uv = use_extension("@aspect_rules_py//uv:extensions.bzl", "uv")
 uv.declare_hub(hub_name = "pypi_pycross_rust_rs")
+uv.package_toolchains(
```
jbedard, line 15: If it's the toolchains for a **package** shouldn't we (optionally?) have a `package` target? I think like `override_package`...

uv/private/extension/defs.bzl
```diff
@@ -1087,12 +1121,37 @@ project locks declared by the same module. Specifying `target` requires `lock`
 and is mutually exclusive with all other modification attributes.""",
 )
 
+_package_toolchains_tag = tag_class(
+    attrs = {
```
jbedard, line 1193: Should this have `toolchains` for the generic toolchains like cc/java? Or we could even have `uv.package_rust_toolchain()` vs `uv.package_toolchains()`...

e2e/crossbuild/pycross-rust-rs/setup.MODULE.bazel
xangcastle: Fixed in bc446420: `name` (optional, like `override_package`) scopes the declaration to one package, precedence package-in-lock > package > lock > module.

uv/private/extension/defs.bzl
xangcastle: Fixed in bc446420: the tag now also carries the generic `toolchains` list (`pycross-jdk` moves its JDK/Ant to it); `rust_toolchain` stays its own attribute since sdist_build wires it only into Rust builds with the exec-sysroot layer.

docs/uv.md
```diff
@@ -382,23 +382,29 @@ useful when a pre-build patch removes stale entry-point metadata.
...
-package's override and their well-known make-variables reach the build
+extensions, Ant, cargo and rustc for Rust extensions. Declare them once with
+`uv.package_toolchains()` and their well-known make-variables reach the build
 environment on their own:
 
 ```starlark
-uv.override_package(
+uv.package_toolchains(
     lock = "//:uv.lock",
-    name = "jpype1",
```
jbedard, line 392: Don't we still only want to add these toolchains to this package?

e2e/crossbuild/pycross-rust-rs/setup.MODULE.bazel
jbedard: This `package_toolchains()` still doesn't specify the package, should it?

e2e/crossbuild/pycross-tiktoken/setup.MODULE.bazel
```diff
@@ -0,0 +1,35 @@
...
+    lock = "//pycross-tiktoken:uv.lock",
+    pyproject = "//pycross-tiktoken:pyproject.toml",
+)
+
+# tiktoken's sdist ships no Cargo.lock, so nothing pins its crates; this one
+# (`cargo generate-lockfile` on the extracted sdist) lets sdist_build vendor
+# them and the build run offline like every other case here.
+uv.override_package(
+    name = "tiktoken",
+    cargo_lock = "//pycross-tiktoken:Cargo.lock",
```
jbedard, line 28: Isn't this part of rust toolchain setup and should be in the `.package_toolchains()`?

jbedard:
@tamird please review the primary rust sdists API here. Also review the API proposals made in PR comments as well as any other sdists and crossbuild related issues you find

e2e/crossbuild/pycross-rust-rs/setup.MODULE.bazel
xangcastle: Dropped the package scope altogether: it is `uv.rust_toolchain(toolchain, lock)` now, module or lock only. Detection decides which sdists get it, so no package name is needed here. 2957fe99

uv/private/extension/defs.bzl
xangcastle: Split as you suggested in 2957fe99: `uv.rust_toolchain()` for Rust, generic toolchains stay on `uv.override_package(toolchains = ...)`.

docs/uv.md
xangcastle: Agreed. Generic toolchains are per package again (`override_package`), so a JDK is an input of the one build that uses it. 2957fe99

e2e/crossbuild/pycross-tiktoken/setup.MODULE.bazel
xangcastle: Kept on `override_package`: a Cargo.lock pins one sdist version's crates, a build input like `pre_build_patches`, not a toolchain. `bazel run @sdist_build__…//:cargo_lock` now generates it and prints this exact declaration.

docs/uv.md
```diff
@@ -382,23 +382,127 @@ useful when a pre-build patch removes stale entry-point metadata.
...
+
+Rust sdists (maturin, setuptools-rust) build with the Rust toolchain your
+module registers. Fetching rustc and cargo is the Rust rulesets' job, and
+rules_py depends on neither of them, so point `uv.rust_toolchain()` at the
+toolchain once and every Rust sdist of every `uv.project()` in the module is
+wired to it. Either ruleset works:
+
+- [rules_rust](https://github.com/bazelbuild/rules_rust):
+  `toolchain = "@rules_rust//rust/toolchain:current_rust_toolchain"`
+- [rules_rs](https://github.com/hermeticbuild/rules_rs): its toolchains are
```
jbedard: > its toolchains are rules_rust `rust_toolchain` instances declared in the patched `rules_rust`

Is that true? Yes or no it sounds like more information then is required here...

xangcastle: Yes, verified. Agreed it doesn't belong in user docs; trimmed.

docs/uv.md
```diff
@@ -382,23 +382,126 @@ useful when a pre-build patch removes stale entry-point metadata.
...
+  ```starlark
+  rules_rust_rs = use_extension("@rules_rs//rs:rules_rust.bzl", "rules_rust")
+  use_repo(rules_rust_rs, rules_rust_rs = "rules_rust")
+  ```
+
+  then `toolchain = "@rules_rust_rs//rust/toolchain:current_rust_toolchain"`.
+
+```starlark
+uv.rust_toolchain(
+    toolchain = "@rules_rust//rust/toolchain:current_rust_toolchain",
```
jbedard, line 429: Doesn't this add the rust toolchain logic to every single package? I feel like we should never recommend that or put it in examples and should always have both a package `name` and uv `lock` in the examples...?

uv/private/sdist_build/repository.bzl
```diff
@@ -159,6 +162,171 @@ def _config_settings_attr(config_settings):
...
+    Returns:
+        struct(load_stmt, target, toolchains): the `load()` lines and the
+        `rust_host_sysroot(...)` and `cargo_lock_generator(...)` targets to
+        splice into the BUILD (both "" when not wired), and the final
+        `toolchains` list.
+    """
+    if not (rust_toolchain and _is_rust_build(inspection)):
+        return struct(load_stmt = "", target = "", toolchains = list(toolchains))
+    target = """
+rust_host_sysroot(
```
tamird, line 339: Please tie the Rust tools and host std to the execution platform selected for `PySdistNativeBuild`. These are ordinary target dependencies: `current_rust_toolchain` resolves in its own rule, and `rust_host_sysroot.actual` uses that rule's default `cfg = "exec"`, while the wheel action runs in `TARGET_EXEC_GROUP`. The Rust toolchain type never participates in that group's selection. With multiple registered execution platforms, the Rust selection can therefore supply a cargo/rustc binary or host std for a different platform than the one running the wheel action.

[Bazel resolves each execution group independently](https://bazel.build/extending/exec-groups#defining-execution-groups). Please make these selections share the wheel action's execution platform and cover a case where the default and wheel execution platforms differ.

uv/private/sdist_build/repository.bzl
```diff
@@ -159,6 +162,171 @@ def _config_settings_attr(config_settings):
...
+)
+""".format(
+        toolchain = repr(rust_toolchain),
+        src = repr(src),
+        output = "\n    output = {},".format(repr(lock_output)) if lock_output else "",
+    )
+    return struct(
+        load_stmt = _RUST_LAYER_LOAD + _CARGO_LOCK_LOAD,
+        target = target,
+        toolchains = [rust_toolchain, ":rust_host_sysroot"] + [t for t in toolchains if t != rust_toolchain],
```
tamird: This does not provide a safe package-level escape hatch once a module or lock default exists. `_collect_toolchain_inputs_and_vars` merges variables in list order, so a package's different Rust toolchain overwrites `CARGO`, `RUSTC`, and `RUST_SYSROOT`, but the generated `RUST_HOST_SYSROOT` still comes from the default toolchain. A cross build can then combine one Rust version's compiler with another version's host std.

Please make a package-specific Rust selection replace the complete compiler/sysroot selection, or reject this ambiguous combination. General extra toolchains can remain additive. This is the concrete reason I think the API needs a package-level exception even if the usual declaration remains module-wide.

uv/private/sdist_build/repository.bzl
```diff
@@ -24,6 +24,9 @@ def _write_context_file(repository_ctx, available_deps):
         "deps": [str(d) for d in repository_ctx.attr.deps],
         "available_deps": available_deps,
     }
+    if repository_ctx.attr.cargo_lock:
+        # A user-supplied lock replaces whatever the sdist ships (usually nothing).
+        context["cargo_lock"] = str(repository_ctx.path(repository_ctx.attr.cargo_lock))
```
tamird, line 29: Please watch this Cargo.lock before passing its path to the configure process. In Bazel 9, [`repository_ctx.path(label)` no longer watches the file by default](https://github.com/bazelbuild/bazel/blob/9.0.0/src/main/java/com/google/devtools/build/lib/packages/semantics/BuildLanguageOptions.java#L289-L300), and the Python read through `_configure_context.json` is opaque to Bazel.

After `:cargo_lock` regenerates the file in place, the label is unchanged, so this repository can retain the old vendored crates while the wheel action receives the new lock. The documented regeneration workflow then fails offline until something else forces a repository refetch. A regression should update the lock contents at the same label and verify that the vendored graph changes.

uv/private/pep517_whl/cargo_lock.bzl
```diff
@@ -0,0 +1,155 @@
...
+case "$sdist" in
+    *.zip) unzip -q "$sdist" -d "$work" ;;
+    *) tar -xf "$sdist" -C "$work" ;;
+esac
+
+# The shallowest Cargo.toml is the crate (maturin: top level; setuptools-rust:
+# a subdirectory such as src/_bcrypt), the same choice the build makes when
+# it places the lock next to the manifest.
+manifest=""
+for depth in 1 2 3 4 5 6 7 8; do
```
tamird, line 44: The shallowest Cargo.toml is not necessarily the manifest the backend builds. Maturin explicitly supports [`tool.maturin.manifest-path`](https://github.com/PyO3/maturin/blob/v1.15.0/guide/src/config.md), and [Cargo keeps the lock at the selected workspace's root](https://doc.rust-lang.org/cargo/reference/workspaces.html). An sdist with two independent workspaces and a deeper selected extension will generate a lock for the wrong graph here; `_find_cargo_lock` and `_inject_cargo_lock` make similar independent guesses.

Please carry the backend-selected manifest and Cargo workspace identity through detection, generation, and injection. If the singular `cargo_lock` attribute cannot represent a package's build layout, reject that layout explicitly instead of choosing by path depth.

uv/private/pep517_whl/cargo_lock.bzl
```diff
@@ -0,0 +1,155 @@
...
+    manifest="$(find "$work" -mindepth "$depth" -maxdepth "$depth" -name Cargo.toml -type f | LC_ALL=C sort | head -n 1)"
+    [ -n "$manifest" ] && break
+done
+if [ -z "$manifest" ]; then
+    echo "error: {sdist_name} has no Cargo.toml; nothing to lock" >&2
+    exit 1
+fi
+
+"$cargo" generate-lockfile --manifest-path "$manifest"
```
tamird, line 53: This resolves the pristine sdist, while the wheel action first applies `pre_build_patches`. If an existing package override patches Cargo.toml to add or upgrade a dependency, `:cargo_lock` writes the old graph and repository generation vendors it; the actual patched build then cannot use that lock offline.

Please generate from the same patched source and relevant backend configuration as the wheel build. A useful regression is a pre-build patch that changes one Cargo dependency and a generated lock that is then consumed by the offline build.

uv/private/sdist_configure/detect_native.py
```diff
@@ -402,6 +404,46 @@ def _extract_names(key: str) -> list[str]:
...
+    Packages without a `source` are the workspace's own crates (the root
+    package, path dependencies) and are already in the sdist; everything else
+    is what cargo would otherwise fetch at build time.
+    """
+    if tomllib is None:
+        return []
+    try:
+        data = tomllib.loads(content)
+    except tomllib.TOMLDecodeError:
+        return []
```
tamird, line 419: Please propagate this parse failure instead of returning a valid-looking empty graph. `_sdist_build_impl` only forwards `cargo_lock` and enables vendoring when `cargo_crates` is truthy, so a malformed explicitly supplied lock is silently discarded. The build can then use the sdist's original lock or resolve online when it ships none, despite the user's override.

Keep malformed input distinct from a valid lock with no registry crates, and ensure an explicitly supplied lock is either consumed or rejected with an actionable error.

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -921,15 +919,145 @@ def _generate_cmake_toolchain_file(
...
+        else:
+            option = None
+            if arg in ("-C", "--codegen") and i + 1 < len(args):
+                option = args[i + 1]
+            elif arg.startswith("-C") and "=" in arg:
+                option = arg[2:]
+            if option is not None and option.split("=", 1)[0] in PROFILE_OPTIONS:
+                codegen.append(option)
+    identity += sorted(crate_types) + sorted(cfgs) + sorted(codegen)
+    metadata = "metadata=" + hashlib.sha256("\\0".join(identity).encode()).hexdigest()[:16]
```
tamird, line 978: Please preserve Cargo's crate metadata rather than reconstructing its private identity scheme here. Cargo deliberately hashes the package source and the metadata of its dependency graph to distinguish compilation units ([implementation](https://github.com/rust-lang/cargo/blob/0.91.0/src/cargo/core/compiler/build_runner/compilation_files.rs)). This replacement includes neither, so same-name/version units from different sources with the same features, profile, and target receive the same Rust crate identity, removing Cargo's collision protection.

If a host path still affects target metadata, that needs a fix in Cargo or an upstream-supported remapping that retains the complete unit identity. The earlier [#1537 evidence](https://github.com/aspect-build/rules_py/pull/1537) already reported matching wheels from two checkout paths without rewriting metadata; please distinguish any remaining cross-host problem from that result.

uv/private/pep517_whl/tools/build_helper.py
```diff
@@ -1006,6 +1150,92 @@ def _configure_cargo_cross_env(build_env: dict[str, str], tmpdir: str, target_os
...
+    """
+    pyproject = path.join(worktree, "pyproject.toml")
+    if not path.exists(pyproject):
+        return False
+    with open(pyproject, encoding="utf-8") as f:
+        content = f.read()
+    if "[tool.maturin.sbom]" in content:
+        return False
+    with open(pyproject, "a", encoding="utf-8") as f:
+        f.write(_MATURIN_SBOM_OFF)
```
tamird, line 1173: Please preserve Maturin's SBOM defaults. `main()` calls this for every Maturin build, including the existing manually wired path, so an omitted table now disables the Rust SBOM and, when repair copies libraries, the auditwheel SBOM that Maturin [generates by default](https://github.com/PyO3/maturin/blob/v1.15.0/guide/src/sbom.md). This removes dependency records that those wheels would otherwise contain.

The unstable sandbox paths should be fixed or normalized at the SBOM producer. An explicit package opt-out can remain available, but selecting the project's Rust toolchain should not implicitly remove this provenance.

tamird (changes requested):
A module or lock default for Rust seems reasonable: the generated BUILD attaches it only to detected Rust sdists, so I do not think every declaration needs a package name. A package exception does need to replace a complete Rust selection, and that selection must use the wheel action's execution platform.

The lock API also needs one consistent relationship between the patched source, backend-selected Cargo workspace, lock contents, and vendored graph. The inline findings identify where those currently diverge. I would also keep Cargo's crate identity and Maturin's SBOM defaults while addressing reproducibility in their owning implementations.

uv/private/sdist_build/repository.bzl
xangcastle: Fixed in 062334ad: the project rust toolchain type now joins TARGET_EXEC_GROUP's resolution via a generated rule (pep517_rust_native_whl).

docs/uv.md
xangcastle: Fixed in ae5bb410: examples lead with the lock-scoped form now; the module-wide one stays documented for single-toolchain modules. It only ever wires maturin/setuptools-rust builds — no other package is touched.

uv/private/sdist_configure/detect_native.py
xangcastle: Fixed in 627f4158: a malformed lock — declared or shipped — now fails repository generation naming the file and the fix; a valid lock with no registry crates stays meaning workspace-only.

uv/private/sdist_build/repository.bzl
xangcastle: Fixed in f47ba0cd: the declared lock is watched before the tool reads it; the tiktoken script case now updates the lock at the same label and asserts the vendored graph follows, both directions.

uv/private/pep517_whl/cargo_lock.bzl
xangcastle: Fixed in 2ee807b3: the configure tool resolves the manifest once (maturin manifest-path, pyproject sibling, or the sole setuptools-rust one) and rejects unrepresentable layouts instead of guessing by depth; injection asks cargo for the workspace root.

uv/private/pep517_whl/cargo_lock.bzl
xangcastle: Fixed in 2ee807b3: :cargo_lock applies the package pre_build_patches before resolving; the tiktoken case patches cfg-if into Cargo.toml and asserts the generated lock carries it for the offline build.

uv/private/pep517_whl/tools/build_helper.py
xangcastle: Keeping this one, with evidence: spoofing only the host line of rustc -vV changes the -C metadata cargo passes (same toolchain, same crate), so two executor platforms mangle different symbols into the same wheel. Cargo's own hash is also source-blind (an edited lib.rs keeps it), so the replacement preserves what cargo distinguished. f68b2b83 folds the crate manifest into the identity to cover patched forks.

PR #1527, opened 2026-09-03
feat(py)!: opt-in console-script wrappers via include_console_scripts

jbedard:
Wheel console-script wrappers (`<venv>/bin/<name>`) and `bin/activate` no longer land in a py_binary/py_test's runfiles by default. A `py_venv` still generates both for `bazel run :name.venv`, but keeps them out of `VirtualenvInfo.runtime_runfiles`; launchers pull the wrappers in only with `include_console_scripts = True`. Internal venvs skip wrapper generation and console-script collision checks entirely when the flag is off.

BREAKING CHANGE: `py_binary`/`py_test` no longer put wheel console scripts on `PATH`. Set `include_console_scripts = True` if packages are invoked as subprocesses by name.

### Test plan

- New test cases added


py/private/py_venv/tests/snapshots/many_scripts.bin.tree.snap
```diff
@@ -1,191 +1,6 @@
-FILE bin/activate
```
jbedard, line 1: Notice the `bin/activate` script is also gone

jbedard: Same with the OCI image listing tests

xangcastle (approved):

PR #1528, opened 2026-09-03
test: add venv console-script snapshot tests

jbedard:
### Test plan

- New test cases added

### DIFF
--- a/py/private/py_venv/tests/snapshots/many_scripts.bin.tree.snap
+++ b/py/private/py_venv/tests/snapshots/many_scripts.bin.tree.snap
@@ -1,0 +1,267 @@
+FILE bin/activate
+---
+# Adapted from CPython Lib/venv/scripts/common/activate
+
+deactivate () {
+    # reset old environment variables
+    if [ -n "${_OLD_VIRTUAL_PATH:-}" ] ; then
+        PATH="${_OLD_VIRTUAL_PATH:-}"
+        export PATH
+    fi
+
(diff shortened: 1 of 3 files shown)


py/private/py_venv/tests/snapshots/many_scripts.bin.tree.snap
```diff
@@ -0,0 +1,267 @@
...
+case $0 in
+    */*) script_dir=${0%/*} ;;
+    *) script_dir=. ;;
+esac
+exec "${script_dir}/python" -c 'import sys; from importlib import import_module; from operator import attrgetter; sys.argv[0] = "pytest"; sys.exit(attrgetter("console_main")(import_module("pytest"))())' "$@"
+---
+LINK bin/python -> ../../../../../../../+python_interpreters+python_3_13_x86_64_unknown_linux_gnu/bin/python3
+LINK bin/python3 -> python
+LINK bin/python3.13 -> python
+FILE bin/sqlformat
```
jbedard, line 180: Note `sqlformat` which is not even a direct dep

xangcastle (approved):

py/private/py_venv/tests/snapshots/many_scripts.bin.tree.snap
xangcastle: something to investigate for sure

jbedard: I think it is correct, when creating these `bin` things we might want _all_ of them, even indirect. I'm just pointing out that the test covers indirect deps like this :+1:

PR #1529, opened 2026-09-04
fix(uv): exclude abi3 wheels from free-threaded Python

thorvald-oai:
## What and why

Require `freethreaded=false` as well as the existing Python >=3.3 condition for the ordinary `abi3` selector. Ordinary stable-ABI wheels are not compatible with free-threaded CPython; accepting them can suppress the sdist fallback when no compatible free-threaded wheel exists.

Keep ordinary `abi3` forward compatibility and the separate `abi3t` selector unchanged.

Extend the existing cryptography fixture with paired analysis tests for Python 3.14: ordinary CPython selects its `cp311-abi3` wheel, while free-threaded CPython selects the sdist build target. The fixture has no `cp314t` wheel that could mask the bug. No dependency or lockfile changes, and no source compilation is claimed by these selection tests.

## Validation

- Before the fix, a configured query reproduced free-threaded Python 3.14 incorrectly selecting the `cp311-abi3` wheel. After the fix, ordinary Python 3.14 still selects that wheel.
- The full `e2e/cases` jobs passed on [Bazel 8](https://github.com/aspect-build/rules_py/actions/runs/33894929078/job/101095553582) and [Bazel 9](https://github.com/aspect-build/rules_py/actions/runs/33894929078/job/101095553786) at `5f984ed9`. Both new analysis tests passed, alongside the existing ordinary-Python import tests and Python 3.14t/3.15t wheel-tag tests.
- Upstream Buildifier and Gazelle checks passed. Local test execution was blocked by timeouts fetching the existing `rules_coreutils` and `ape` 1.0.1 archives; the execution results above come from upstream CI.


thorvald-oai:
@codex review

jbedard (approved):

PR #996, opened 2026-05-11
fix(uv): recognize generic linux wheel platform tags

jbedard:
### Test plan

- Covered by existing test cases
- New test cases added

PR #1532, opened 2026-09-08
fix(uv): recognize generic linux wheel platform tags (1.x backport of #996)

xangcastle:
Backport of #996 (2aff0bee) to the 1.x maintenance branch.

## Problem

1.x deliberately rejects bare-linux wheel platform tags (`linux_x86_64`, `linux_armv7l`, ...) in `supported_platform()`. Wheels published without `auditwheel repair` (internal registries, PyTorch CPU wheels) generate no select arm; if the package has no sdist it resolves to `@platforms//:incompatible` and the build breaks.

## Changes

- Code is a byte-identical port of #996 (zero drift between 1.x and the commit's parent on the touched files): `GENERIC_LINUX_ARCHES`, `linux_*` acceptance in `supported_platform()`, `generate_linux()` config_settings (os+cpu only, no libc/version constraint), `armv6l/armv8l → arm` mangling, and removal of the `linux_armv7l → manylinux_2_17_armv7l` FIXME alias.
- `constraints/platform/test.bzl`: rejection assert inverted; new `generic_linux_arches_test` and `bare_linux_tags_test`; `linux_sparc64` remains rejected.
- e2e `cases/uv-bare-linux-wheels` ported from #996 and adapted to the 1.x module wiring (dedicated hub, `default_build_dependencies` on `uv.project`). The uv_hub snapshot goldens updated by #996 do not exist on 1.x; equivalent coverage lives in the case's genquery+grep assertions.
- `select_key` needed no change: bare-linux ranks `(0, 0)`, below any manylinux/musllinux, so it only matches when no better wheel exists (same emergent behavior as main).

## Observable behavior

A bare-linux wheel now matches on musl hosts and glibc < 2.17 as well (config_setting carries no libc constraint), mirroring pip/uv (`packaging.tags` includes the host's bare tag). It acts strictly as a fallback and never shadows a compatible manylinux/musllinux wheel.

## Test plan

- Red without the code fix: `//cases/uv-bare-linux-wheels:test` fails — `linux_armv7l` constraint absent from `deps(@pypi-bare-linux-wheels//grpcio)`.
- Green with the fix: e2e passes; `bazel test //uv/...` 89/89; regressions `uv-platform-filter-844`, `uv-dep-hashes` (both), `uv-abi3-compat-853` all pass.

### DIFF
--- a/e2e/MODULE.bazel
+++ b/e2e/MODULE.bazel
@@ -383,6 +383,22 @@ uv.project(
 )
 # }}}
 
+# For cases/uv-bare-linux-wheels
+# Regression: grpcio ships bare linux_armv7l wheels that must survive lockfile parsing.
+# {{{
+uv.declare_hub(hub_name = "pypi-bare-linux-wheels")
+uv.project(
+    default_build_dependencies = [
+        "build",
+        "setuptools",
+    ],
+    hub_name = "pypi-bare-linux-wheels",
+    lock = "//cases/uv-bare-linux-wheels:uv.lock",
+    pyproject = "//cases/uv-bare-linux-wheels:pyproject.toml",
+)
+use_repo(uv, "pypi-bare-linux-wheels")
+# }}}
+
 # For cases/pytest-mock-530
 # Regression: pytest-mock plugin discovery with py_test and py_venv_test.
 # {{{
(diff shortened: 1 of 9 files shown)


xangcastle:
@codex

jbedard (approved):

PR #1533, opened 2026-09-08
fix(uv): reject malformed python/abi wheel tags via allowlist

xangcastle:
Completes the tag-validation symmetry started by #844: platform tags have had a strict allowlist since then, but python tags were only prefix-checked (`cp311_2` passed) and abi tags were not filtered at all. A single wheel with a malformed tag in the lockfile — e.g. renamed in a private registry — breaks analysis far beyond that wheel.

## Problem

Two distinct failure modes, both reproduced by the new e2e case:

1. **Malformed python tag** (`cp311_2-...`): crashes the whl_install repository fetch before any select is generated — `Error in int: invalid base-10 literal: "11_2"` in `select_key` (`whl_install/repository.bzl`).
2. **Malformed abi tag only** (`cp311-cp311_2-...`): the generated config group name collides with a *private intermediate alias* that bazel_skylib's `selects.bzl` creates for the legitimate `cp311` group chain (`name + "_" + i`), leaving the entire `pip_configurations` BUILD in error — every package of every hub breaks.

## Changes

Single source of data for both filtering and generation, mirroring #844:

- `constraints/defs.bzl`: `FOREIGN_INTERPRETER_PREFIXES` + `foreign_interpreter_tag()` — foreign interpreters (pp/pypy/graalpy/ip/jy) skip silently; anything else unsupported warns.
- `constraints/python/defs.bzl`: `PYTHON_TAGS` derived from INTERPRETERS×MAJORS×MINORS (exactly what `python/macro.bzl` generates); `supported_python()` is now membership, not a prefix check; `filter_python_tags()` warns per rejected tag.
- `constraints/abi/defs.bzl` (new): `ABI_FEATURES`/`ABI_FEATURE_SUFFIXES`/`ABI_TAGS` (+`none`, `abi3`), `supported_abi()`, `filter_abi_tags()`.
- `constraints/abi/macro.bzl`: refactored to generate from that same data — verified by query that the 1438 generated targets are unchanged.
- Both consumers filter before generating: `extension/lockfile.bzl` (hub configurations) and `whl_install/repository.bzl` (select arms), so hub and arms stay consistent.

A rejected wheel is skipped with an actionable warning, never a hard failure:

```
WARNING: iniconfig-2.3.0-cp311-cp311_2-manylinux_2_17_x86_64.whl: abi tag 'cp311_2' does not match any supported configuration; skipping this wheel as an install candidate. This usually means the wheel file was renamed with a malformed tag — fix the filename in the registry and re-lock.
```

If the malformed wheel was the only candidate, the existing "no compatible wheel" diagnostics take over.

## Test plan

- New e2e `cases/uv-malformed-abi-tag`: crafted uv.lock carrying a real `py3-none-any` wheel plus two synthetic malformed entries (never fetched). Red without the fix reproduces both failure modes above; green with the fix resolves via the valid wheel and asserts `cp311_2` never leaks into the graph.
- New unit suites `constraints/python/test.bzl` and `constraints/abi/test.bzl`: `cp311_2`/`cp999`/`cp321`/`py4`/`cp3`/`cp311td` rejected; `none`/`abi3`/`cp313t`/`cp312dmtu` accepted; foreign interpreters silent.
- `bazel test //uv/...` 95/95; e2e regressions: uv-platform-filter-844, uv-abi3-compat-853, freethreaded-805, uv-dep-hashes (×2), uv-no-sdist-754, windows-crossbuild-837, uv-deps-650 (×4) all pass; buildifier clean.


uv/private/whl_install/repository.bzl
```diff
@@ -119,19 +120,19 @@ def _whl_install_impl(repository_ctx):
     for whl, target in prebuilds.items():
         parsed = parse_whl_name(whl)
 
+        # Escape hatch for ignoring unsupported interpreters and abis
+        whl_python_tags = filter_python_tags(whl, parsed.python_tags)
+        whl_abi_tags = filter_abi_tags(whl, parsed.abi_tags)
+
         # FIXME: Make it impossible to generate absurd combinations such as
```
jbedard, line 127: Is this fixed or still an issue?

jbedard (approved):

PR #1537, opened 2026-09-10
feat(uv): reproducible rustc wrapper and cc-rs environment for Rust sdists

xangcastle:
Stacked on #1526. Helper-only: the rustc wrapper and the cargo environment that PR introduces.

**Reproducible rustc wrapper.** The wrapper only injected `--sysroot`, so a Rust wheel carried the sandbox and execroot paths of the host that built it (debug info, panic locations) and LLVM's per-host module ids in symbol names. The wheel action itself hits the remote cache, but its bytes differ per host, so every downstream action (`whl_install`, venvs) misses. The wrapper now passes `--remap-path-prefix` for the sandbox root and the execroot and `-C codegen-units=1` for target crates; build scripts and proc-macros, which never reach the wheel, keep cargo's codegen. Native builds treat every crate as target. maturin's CycloneDX SBOM is turned off in the extracted `pyproject.toml` unless the sdist configures `[tool.maturin.sbom]` itself: it records every crate as `path+file:///<sandbox>/...`, and the sandbox id differs per action, so no two builds agreed while it was on (maturin 1.15 has no flag or variable for it). Rewriting cargo's `-C metadata=` hash is left for a follow-up: it needs a host-independent toolchain identity, and the changes here already make two checkouts produce identical wheels (see test plan).

**cc-rs environment.** Crates with C or C++ sources (`ring`, `zstd-sys`) build them through cc-rs, which looks up `CC_<triple>`, `CXX_<triple>`, `AR_<triple>` and `RANLIB_<triple>` (dashed and underscored spellings) before falling back to the PATH; only `CARGO_TARGET_<TRIPLE>_LINKER` was set. Cross builds now export the wired C toolchain under those names, with a ranlib wrapper over `ar s` for toolchains that ship none (shared with the CMake toolchain file). The cargo linker stays the C driver: linking with `c++` would give every Rust extension an implicit libstdc++ dependency, not only those with C++ sources.

Tests: the generated wrapper is run around an argv-echoing fake rustc (remap flags, codegen-units gating on `--target`, native mode, cargo's arguments untouched); the cc-rs variables are checked directly and through the cross env.

---

### Test plan

- New test cases added (see above).
- `cd e2e/crossbuild && bazel test //pycross-rust/... //pycross-rust-rs/... //pycross-tiktoken/...` passes offline (native, linux/amd64 and linux/arm64 from a macOS host).
- Reproducibility: the same commit checked out at two paths, hence two output bases and sandbox ids, builds byte-identical cross pydantic-core wheels:

  | wheel | sha256 (both checkouts) |
  |---|---|
  | `pydantic_core-2.47.0-cp312-cp312-linux_x86_64.whl` | `0c5d3027e6be866c0005247c277263560450fe6a57c2d01aa0665d718217a723` |
  | `pydantic_core-2.47.0-cp312-cp312-linux_aarch64.whl` | `b9c9af30a4faae3ebc6781a304307722320627cb1c84de9b554ae9c94df20729` |

  Before this PR the `.so` differed by sandbox paths and, with the wrapper alone, only the SBOM and its `RECORD` line still differed.

### DIFF
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -463,8 +463,9 @@ instead of fetching rustc inside the build.
 The rustc cargo runs is a wrapper that also makes the extension independent
 of where the action ran: sandbox and execroot paths are remapped out of the
 binaries (`--remap-path-prefix`) and target crates compile as one codegen
-unit, so the wheel's bytes match across hosts and downstream actions hit the
-cache. Crates that compile C or C++ through cc-rs (`ring`, `zstd-sys`) find
+unit, and maturin's SBOM, which records sandbox paths, is turned off unless
+the sdist configures it; the wheel's bytes then match across hosts and
+downstream actions hit the cache. Crates that compile C or C++ through cc-rs (`ring`, `zstd-sys`) find
 the wired C toolchain under `CC_<triple>`, `CXX_<triple>`, `AR_<triple>` and
 `RANLIB_<triple>` instead of whatever is on the PATH.
 
--- a/docs/uv.md
+++ b/docs/uv.md
@@ -460,6 +460,14 @@ maturin is told not to download a Rust toolchain of its own
 (`MATURIN_NO_INSTALL_RUST=1`): a Rust sdist with no toolchain wired fails
 instead of fetching rustc inside the build.
 
+The rustc cargo runs is a wrapper that also makes the extension independent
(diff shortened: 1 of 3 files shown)


xangcastle:
Folded into #1526 as commits 0c8eae12 and 5e93608e; same tree, same evidence.

PR #1538, opened 2026-09-11
feat(py): propagate type stubs and carry them into venvs

### LINKED ISSUE #558: [FR]: Support for pyi files?
### What is the current behavior?

Hi, 

rules_python has support for `pyi_srcs`. I saw that however aspect rules_py does not have that. 
For the time being I am adding the `.pyi` files to the `data` field, but I wonder if adding `pyi_srcs` would be helpful? 

### Describe the feature

It would also be nice to allow `aspect configure` to understand `.pyi` files too and put them in the same library too. 

Thank you!

xangcastle:
Builds on #1523 (its commit is kept as the first commit here, authorship intact) and makes `transitive_pyi_files` mean the same thing on both sides of the ruleset boundary.

Closes #558.

### Changes

- `.pyi` files listed in `srcs` are classified by extension into `PyInfo.transitive_pyi_files`, mirroring rules_python where stubs never count as runtime sources. `transitive_sources` no longer carries them, including through virtual-dependency resolutions. No new `pyi_srcs` attribute.
- Venvs that include sources and launchers put both depsets in runfiles. Stubs that only exist in a foreign provider (protobuf's `py_proto_library` `_pb2.pyi`, a rules_python `pyi_srcs`) now land beside the modules they annotate, so a type checker pointed at the venv resolves them. Neither rules_python nor protobuf put stubs in their own runfiles; the venv is where rules_py has to do it.
- Under `--@aspect_rules_py//py:emit_rules_python_providers` the emitted rules_python `PyInfo` also populates `direct_pyi_files`.
- Drops the redundant resolution wrapper and `has_py_info` guards; the interop accessor already returns an empty depset for targets carrying neither provider.

`PyInfo.transitive_sources` no longer lists `.pyi` files from `srcs`; they move to `PyInfo.transitive_pyi_files`. Runfiles of venvs and launchers gain stubs that come from rules_python deps.

### Test plan

- `//py/tests/type-stubs`: analysis coverage for first-party, foreign, virtual-resolution, launcher, venv and compatibility-provider shapes, plus a `py_test` asserting the stubs sit in runfiles at execution time.
- `e2e/rules-python-interop`: a rules_python library with `pyi_srcs` consumed by a rules_py `py_library` and `py_test`.
- `e2e/rules-python-protobuf`: protobuf's `py_proto_library` generating `greeting_pb2.pyi`, propagated through rules_py and present in runfiles.
- `e2e/rules-python-provider-compat`: rules_python's `PyInfo` builder merging a rules_py stub under the compatibility layer.
- Full `//...` plus the three e2e workspaces and `test.sh` pass locally.


xangcastle:
@codex

py/private/py_venv/py_venv.bzl
```diff
@@ -124,7 +124,9 @@ def _venv_providers(ctx, venv, venv_only, executable = None, include_sources = F
     """Providers emitted by both the executable and lib variants."""
     runfiles = venv.runtime_runfiles.merge(ctx.runfiles(files = venv_only))
     if include_sources:
-        runfiles = runfiles.merge(ctx.runfiles(transitive_files = venv.transitive_sources))
+        runfiles = runfiles.merge(ctx.runfiles(transitive_files = depset(
```
jbedard, line 127: I don't think PYI should be in the runfiles at all...

py/private/py_library.bzl
```diff
@@ -46,13 +69,18 @@ def _make_virtual_depset(ctx):
     )
 
 def _make_resolved_virtual_depset(target):
-    transitive = [target[DefaultInfo].files]
+    # A resolution target's default outputs stand in for its sources (a wheel
+    # install tree has no PyInfo of its own). Stubs among them belong to the
+    # pyi depset collected alongside, not to the runtime sources.
+    direct = _runtime_sources(target[DefaultInfo].files.to_list())
```
jbedard, line 75: We can't call `.to_list()`

py/private/py_info.bzl
```diff
@@ -11,7 +11,8 @@ deps to build the eventual venv or wheel.
 RulesPyInfo = provider(
     doc = "Python source, import-path, and virtual-dependency information for a target's dependency closure.",
     fields = {
-        "transitive_sources": "depset[File] — postorder depset of first-party `.py` sources in the transitive closure.",
+        "transitive_sources": "depset[File] — postorder depset of first-party runtime sources in the transitive closure; `.pyi` stubs are excluded.",
+        "transitive_pyi_files": "depset[File] — postorder depset of `.pyi` type stubs in the transitive closure: stubs listed in `srcs` plus those carried by deps of either ruleset.",
```
jbedard, line 15: I think we can drop the "stubs listed in `srcs` plus those carried by deps of either ruleset." - that is an implementation detail of rules, and I don't think we should mention "other rulesets" in these docs

jbedard (approved):

PR #1544, opened 2026-09-17
test(interpreter): make the stray-bytecode e2e portable and cache-isolated

xangcastle:
The `interpreter-toolchain-settings` e2e added in #1540 plants bytecode inside a fetched interpreter repo and cleans up on exit.

- The cleanup used `rmdir --ignore-fail-on-non-empty`, a GNU-only flag that BSD `rmdir` rejects, so on macOS the planted `__pycache__` directories were left behind. Each directory is now removed individually, tolerating a failure.
- With the repo contents cache (default since Bazel 9) `external/<repo>` is a symlink into a cache shared by every workspace of the user, so the planted files were visible to unrelated builds during the run and would survive a killed shell. The workspace `.bazelrc` now disables that cache so the repo is fetched into its own output base. The flag exists on Bazel 8 as well, where the cache is already off by default.

### Test plan

- `e2e/interpreter-toolchain-settings/test.sh` on macOS with a fresh output base under Bazel 8.6 and 9.2: the interpreter repo is a plain directory, the script passes, and no planted file remains in the output base or the shared cache.


xangcastle:
@codex

jbedard (approved):

PR #1546, opened 2026-09-17
ci: connect builds to Aspect Cloud

joshcbarnes:
Points CI at Aspect Cloud's remote cache and BES. The jobs run on stock GitHub-hosted runners, which have no cache of their own, and nothing here named an endpoint.

Draft: opened to exercise the one thing that cannot be validated locally — the `ASPECT_OSS_API_TOKEN` → JWT exchange.

Follows aspect-build/rules_lint#1003, adapted to this repo's two differences: a self-hosted runner pool, and ~20 sibling Bazel workspaces that CI drives through the Aspect CLI rather than vanilla `bazel`.

### What it takes

`setup-aspect` already pipes its `aspect-api-token` input into `aspect auth login --with-api-token`. That records a credential but **no endpoints**, and Aspect Cloud's built-in entry ships with none, so a token alone leaves `--remote` with nothing to wire. The endpoints come from a committed `.aspect/config.json`, found through `$ASPECT_WORKSPACE` — pinned to the checkout root, since the test matrix builds from subdirectories. That variable selects only the deployment list, never which `.aspect/config.axl` a task loads, nor which workspace is built.

`REMOTE_FLAG` and `BAZEL_ASPECT_FLAGS` are empty when the secret is absent, which is every fork PR: forks get no org secrets, the login never runs, and either flag would reach Aspect Cloud with no credential.

### Why `test` computes its own `REMOTE_FLAG`

`--remote` is a **hard failure** on a self-hosted Aspect Workflows runner — `REMOTE_REFUSED_BY_ENV`: the `ASPECT_WORKFLOWS_*` environment already names that machine's deployment, and the CLI refuses to route the build off it rather than wire a second one. The job's `REMOTE_FLAG` therefore reads its own `runs-on` condition back, so the four workspaces that ask for the 32-core pool keep the runner's wiring. Without the gate they would fail outright.

### Why the version bump

The pin was 2026.28.2, which predates the deployment feature — no `--remote`, no `aspect auth configure`. `--task-key` is gone at 2026.38.14, and an unrecognized flag is **forwarded to Bazel** rather than rejected, so the three call sites move to `--task:name` or the build fails. Only the root pins a version; the launcher walks up, so the sibling workspaces follow it.

### One rc, two workspaces

`e2e/crossbuild` is the only sibling workspace CI builds with vanilla `bazel`. It imports the root's rc through `%workspace%/../..` instead of carrying a copy — `%workspace%` expands to an absolute path before Bazel opens the file, so that reaches the root from any working directory, unlike a bare relative path. Inside the imported file `%workspace%` re-expands to the *importing* workspace, which is why `.aspect/generated/` is committed there too: that is where its execution log lands, and Bazel fails the build rather than create the directory.

### Verified locally

- `aspect build --remote` from the repo root and from `e2e/crossbuild` with `ASPECT_WORKSPACE` at the root — the CI shape — both print `Using deployment 'aspect-cloud'`, `Streaming build events to grpcs://bes.aspect.build`, and an `app.aspect.build/i/…` link.
- With `~/.aspect/config.json` moved aside, `--remote` fails `no default is set` **without** `ASPECT_WORKSPACE` and resolves **with** it, confirming the committed file is what CI will read.
- `bazel build --config=aspect-cloud` streams to Aspect Cloud from the root, and `--config=aspect-cloud` resolves in `e2e/crossbuild` and from a subdirectory of it.
- The execution log from a `e2e/crossbuild` build lands in `e2e/crossbuild/.aspect/generated/`, leaving the root's untouched.
- Removing `.aspect/generated/` fails the build, confirming the committed directory is load-bearing.

### What this PR is meant to prove

Local runs used a browser login, not the token, so CI is the first exercise of:

- [ ] `ASPECT_OSS_API_TOKEN` exchanges for a JWT. **Note:** setup-aspect `core.warning`s rather than fails on a bad login, so check that step's output explicitly — a green setup step does not mean it worked.
- [ ] That account has cache/BES write access for the org.
- [ ] The CLI jobs print `Using deployment 'aspect-cloud'`; the vanilla-`bazel` jobs print `Streaming build results to`.
- [ ] The CLI bump from 2026.28.2 holds across all 21 workspaces on both Bazel 8 and 9.
- [ ] The 32-core matrix entries still run green on the self-hosted pool with `REMOTE_FLAG` empty.

### Deliberately out of scope

The `test.sh` scripts in `e2e/*` invoke plain `bazel` in ~20 nested workspaces. They stay off Aspect Cloud: an rc import alone would not connect them, since those scripts pass no `--config=aspect-cloud` and `$BAZEL_ASPECT_FLAGS` is not threaded into them.


.bazelrc
```diff
@@ -1,3 +1,6 @@
+# Aspect's rc for vanilla `bazel`, maintained by `aspect setup bazelrc`.
+try-import %workspace%/.aspect/bazelrc
```
jbedard, line 2: Doesn't this always exist and should be `import`?

e2e/crossbuild/.bazelrc
```diff
@@ -1,3 +1,11 @@
+# Aspect's rc for vanilla `bazel`. Generated at the repository root by
+# `aspect setup bazelrc`; regenerate there, not here. %workspace% is this
+# workspace, so `../..` reaches the root it sits under — and re-expands to this
+# workspace inside the imported file, which is why .aspect/generated/ is
+# committed here too: --config=aspect-cloud writes the execution log there and
+# Bazel will not create the directory.
+try-import %workspace%/../../.aspect/bazelrc
```
jbedard, line 7: Doesn't this always exist and should be `import`?

.github/workflows/workspace_status.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+
+# Stopgap for `aspect setup workspace-data` (aspect-build/aspect-cli#1465).
+#
+# Emits the build metadata `aspect <task>` sends as `--build_metadata`, in the
+# `KEY value` form Bazel's --workspace_status_command expects, so a vanilla
+# `bazel` invocation is attributed in the Aspect Web UI instead of arriving
+# anonymous. Once #1465 ships, delete this file and pass the CLI directly:
```
jbedard: Should we wait for 1465? This file is ugly

jbedard (approved):

e2e/crossbuild/.bazelrc
joshcbarnes: This is what aspect-cli adds. Its justifiction is ` "try-import rather than import so a generated file that later disappears degrades to no Aspect configuration instead of failing every bazel call."`

jbedard: If this is generated then leave it as-is 👍

PR #1548, opened 2026-09-18
fix(uv): match CPython wheel ABIs to the interpreter minor

dzbarsky:
CPython ABI tags currently use lower-bound Python version flags, so Python 3.13 also matches `cp311` and `cp312` wheels. Exclude the next minor's version flag when matching a version-specific ABI. The `abi3` constraint keeps its forward-compatible behavior.

The test project runs Python 3.13 but its lockfile only included Python 3.10–3.11 wheels. Regenerate the lockfile for the configured interpreter and update `rpds-py` to 0.19.1, its first release with `cp313` wheels; all other package versions stay unchanged. Regenerate the affected virtualenv snapshots.

Validation: the new Python 3.12 and 3.13 analysis tests fail before the fix. All 22 tests in the ABI and virtualenv test packages pass afterward on Bazel 8.6.0.


xangcastle:
Pushed two commits on top of yours: `//uv/private/uv_hub:snapshots` was red after the widened lock (new cp312 configurations, markers and SCC arms), regenerated; and the ABI tests now also cover 3.13 free-threaded, 3.13 debug and the last minor in MINORS, with flag labels through `Label()`. Full root suite green locally on Bazel 8.

Follow-up, not for this PR: the python-tag axis has the same lower-bound shape for `cpXY` (`constraints/python` aliases `cp312` to `py312`). Right for `cpXY-abi3`, wrong for `cpXY-none-<plat>` wheels, which pip treats as exact.
cc: @jbedard

jbedard:
@codex

jbedard:
@codex review

jbedard:
@xangcastle I added one more that codex flagged

xangcastle (approved):

PR #1551, opened 2026-09-22
ci: connect builds to Aspect Cloud

joshcbarnes:
Wires this repo's CI to Aspect Cloud so every invocation — task-driven and vanilla `bazel` alike — lands in the Web UI with full attribution.

## What's here

**`aspect <task>` half.** `aspect buildifier`/`gazelle`/`test` authenticate via `setup-aspect` with `ASPECT_OSS_API_TOKEN` and send commit, branch, PR and actor as `--build_metadata`. `.aspect/config.json` is committed so CI runners know `cache.aspect.build` and `bes.aspect.build` are Aspect Cloud's; it is read only when `$ASPECT_WORKSPACE` names its workspace, hence the workflow-level env var.

**Vanilla `bazel` half.** `--config=aspect-cloud` from the committed `.aspect/bazelrc` carries the endpoints and credential helpers. Attribution needs per-invocation data the rc constants cannot hold, which is what `--workspace_status_command` is for. This started as a stopgap script (`.github/workflows/workspace_status.sh`) while aspect-build/aspect-cli#1465 was open; that has since shipped, so the final commit deletes the script and points the flag at `aspect setup workspace-data`.

Keeping the flag in `.aspect/bazelrc` rather than the workflow env also attributes local developer builds, which the workflow-only wiring could not reach.

## Verification

Checked against the production `ivy` database rather than by inspection — rules_py invocations arrive with branch, commit, author, PR number and run type populated, for both the `aspect test` tasks and the vanilla `bazel run`/`bazel build` steps.

Also confirmed for CLI 2026.38.24 (the version this bumps to):

- `aspect setup workspace-data` emits the same key set the deleted script did, under a simulated GitHub Actions `pull_request` environment.
- Its task banner goes to stderr, so stdout stays clean `KEY value` lines.
- It exits 0 outside a git repository, so a non-git checkout degrades rather than failing the build.
- Bazel's rc tokenizer handles the quoted multi-word value; a real `--config=aspect-cloud` build wrote the keys to `volatile-status.txt`.
- The status command resolves `aspect` from the client's PATH, so `setup-aspect` after `setup-bazel` in the darwin job is fine.

## Notes for the reviewer

- **The rc line is hand-added.** `aspect setup bazelrc` does not emit `--workspace_status_command`, so regenerating `.aspect/bazelrc` silently drops CI attribution. The comment above the line says so.
- **A missing `aspect` now fails the build outright** (`status 127`), where previously a credential-helper miss only failed on remote traffic. Every `bazel` step therefore needs `setup-aspect` ahead of it.
- **Fork PRs get no org secrets**, so both `REMOTE_FLAG` and `BAZEL_ASPECT_FLAGS` are empty there and nothing reaches Aspect Cloud uncredentialed.
- **New per-build log noise**: `--config=aspect-cloud` invocations now print the CLI task banner through `BazelWorkspaceStatusAction`. The CLI has no quiet flag for it.
- **`crossbuild-wheels-darwin` has not exercised this yet** — it is gated on `main` or a branch name containing `macos`, so its new `setup-aspect` step and `$BAZEL_ASPECT_FLAGS` have never run. Worth a throwaway `…-macos` branch before merge.


joshcbarnes:
Closing: this branch was based on a stale main and duplicates the wiring already merged in #1546.

